PR #59, opened 2022-05-29
Implement fleetshard runtime

SimonBaeumer:
### Specification

 - Centrals should be polled every X seconds
 - A Central should be reconciled in parallel
 - No more than one Central reconciliation can run for a Central tenant
 - The status is reported once per reconcile to reduce cross-thread communication
 - Errors don't stop the reconciler, they are just logged (and should be reported to the fleet-manager)

Docs PR: https://github.com/stackrox/acs-fleet-manager/pull/75

## Follow-Up PR

 - Contains tests
 - Contains docs

## Notes

 - Centrals should not be indexed by name because of collisions
 - 

```
# To run tests locally run:
```


fleetshard/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,97 @@
...
+		k8sClient:         k8sClient,
+		client:            client,
+		reconcilerErrorCh: make(chan centralreconciler.ReconcilerError),
+	}
+}
+
+func (r *Runtime) Stop() {
+}
+
+func (r *Runtime) Start() error {
```
kovayur, line 63: To me the reconciler now has too many responsibilities. It handles not only the reconciliation with the k8s cluster, now it also handles the locks and updates the status.

How about keeping it synchronous with the same signature and handle the parallelisation and sending the statues by the caller (`Runtime`)?

How about this (not tested, just sharing the idea):
```
func (r *Runtime) Start() error {
	ticker := concurrency.NewRetryTicker(func(ctx context.Context) (timeToNextTick time.Duration, err error) {
		startedAt := time.Now()

		list, err := r.client.GetManagedCentralList()
		if err != nil {
			glog.Errorf("Failed to list central", err)
			return 0, err
		}

		results := make(chan reconcileResult)
		reconciler := centralreconciler.NewCentralReconciler(r.k8sClient, r.client)
		statusesMap := make(map[string]private.DataPlaneDinosaurStatus)

		for _, remoteCentral := range list.Items {
			remoteCentral := remoteCentral
			go func() {
				status, err := reconciler.Reconcile(context.Background(), remoteCentral)
				if err != nil {
					glog.Errorf("failed to reconcile central %s: %s", remoteCentral.Metadata.Name, err)
				}
				results <- reconcileResult{
					id:     remoteCentral.Id,
					status: status,
				}
			}()
		}

		for i := 0; i < len(list.Items); i++ {
			result := <-results
			if result.status != nil {
				statusesMap[result.id] = *result.status
			}
		}

		_, err = r.client.UpdateStatus(statusesMap)
		if err != nil {
			glog.Errorf("failed to update statuses %s: %s", err)
		}

		return r.nextTick(startedAt), nil
	}, 10 * time.Minute, backoff)

	return ticker.Start()
}

func (r *Runtime) nextTick(startedAt time.Time) time.Duration {
	duration := time.Now().Sub(startedAt)
	return max(0, 30 * time.Second - duration)
}

func max(a, b time.Duration) time.Duration {
	if a > b {
		return a
	}
	return b
}

type reconcileResult struct {
	id     string
	status *private.DataPlaneDinosaurStatus
}
```

kovayur (approved):

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -10,53 +10,66 @@ import (
...
-	client ctrlClient.Client
+func (r CentralReconciler) ReceiveCh() chan *private.ManagedDinosaur {
+	return r.receiveCh
 }
 
-func (r ClusterReconciler) Reconcile(ctx context.Context, remoteCentral private.ManagedDinosaur) (*private.DataPlaneDinosaurStatus, error) {
+func (r CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.ManagedDinosaur) error {
+	// Only allow to start reconcile function once
+	if !atomic.CompareAndSwapInt32(r.blocked, FreeStatus, BlockedStatus) {
+		return errors.New("Reconciler still busy, wait until current reconciliation has stopped.")
```
kovayur, line 66: `wait until ...` sounds like a blocking operation.

```suggestion
		return errors.New("Reconciler still busy, skipping reconciliation attempt.")
```

fleetshard/pkg/runtime/runtime.go
SimonBaeumer: > handles the locks

Yes, it is by intention to be safe to call. The reconcile can now called as often as it want without running into unsafe operations (i.e. parallel reconciles on the same Central instance).
This makes it more convenient in our use-case due to the polling mechanism.
Handing the locking in the reconcile itself also allows us triggering the reconcile from other sources, like kube events, too.

> and updates the status

Why do you consider sending the status updates to the fleet-manager should not be done in the reconciler?
The reconciler takes care of the end-to-end deployment/upgrade flow of a central. 
The only reason not sending the updates from the reconcile would be if status should not be reported to a fleet-manager.

> handle the parallelisation and sending the statues by the caller (Runtime)?

Updates see above.
Handling parallelisation by the runtime would mean the reconcile function allows parallel execution by default only prevented by the runtime.
With this approach the caller is responsible to implement guarding logic to not run into race-conditions on the reconcile.

SimonBaeumer: Starting the reconcile from the runtime in a goroutine imho can make sense. 
The idea for the reconciler was to have only one instance which can receive from multiple sources - to prevent parallel reconciles it should only ever exists one reconciler per Central.

For our use-cases now likely to be too complex.

fleetshard/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,97 @@
...
+	Duration: 5 * time.Second,
+	Factor:   3.0,
+	Jitter:   0.1,
+	Steps:    5,
+	Cap:      10 * time.Minute,
+}
+
+type Runtime struct {
+	client            *fleetmanager.Client
+	reconcilers       reconcilerCache
```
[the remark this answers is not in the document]
SimonBaeumer: Good catch.
If a central instance is deleted it should be cleaned up from this map. Wouldn't call it cache anymore though, `Registry` would be a better term.

fleetshard/pkg/centralreconciler/reconciler.go
SimonBaeumer: done

fleetshard/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,94 @@
...
+			continue
+		}
+
+		resp, err := r.client.UpdateStatus(map[string]private.DataPlaneCentralStatus{
+			result.Central.Id: result.Status,
+		})
+		if err != nil {
+			glog.Errorf("error occurred %s: %s", result.Central.Metadata.Name, err.Error())
+		}
+		glog.Infof(string(resp))
```
mtesseract, line 92: Is it useful to output the response with Info severity in case `err != nil`?

mtesseract (commented):
Partial review.

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -10,50 +10,64 @@ import (
...
-	}
+// CentralReconciler reconciles the central cluster
+type CentralReconciler struct {
+	client             ctrlClient.Client
+	central            private.ManagedCentral
+	fleetManagerClient *fleetmanager.Client
+	receiveCh          chan *private.ManagedCentral
+	resultCh           chan ReconcilerResult
+	stopCh             chan struct{}
+	blocked            *int32
```
mtesseract, line 31: The name `blocked` seems to correspond to a single variant of the above enum, namely `BlockedStatus`. I'd suggest to rename this field to something like `status` or `blockingStatus`.

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -10,50 +10,64 @@ import (
...
-	if err != nil {
-		glog.Fatal(err)
-	}
-	glog.Infof(string(resp))
 }
 
-// ClusterReconciler reconciles the central cluster
-type ClusterReconciler struct {
-	client ctrlClient.Client
+func (r CentralReconciler) ReceiveCh() chan *private.ManagedCentral {
```
mtesseract, line 60: nit: I find it a bit misleading that the exported getter here carries "Receive" in its name, while the actual user of this function (the runtime) uses it for *sending* something.

mtesseract (changes requested):

dev/scripts/ffm_curl.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
SimonBaeumer: Moved the `dev/scripts` to the `scripts/` dir and remove `ffm_curl.sh`

fleetshard/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/stackrox/acs-fleet-manager/fleetshard/pkg/centralreconciler"
+	"github.com/stackrox/acs-fleet-manager/fleetshard/pkg/fleetmanager"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/private"
+	"github.com/stackrox/rox/pkg/concurrency"
+	"k8s.io/apimachinery/pkg/util/wait"
+	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
+)
+
+type reconcilerRegistry map[string]*centralreconciler.CentralReconciler
```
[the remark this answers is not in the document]
SimonBaeumer: Added a small comment. With that being said most of the things need more docs and polishing

fleetshard/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,94 @@
...
+	reconcilers        reconcilerRegistry //TODO(yury): remove central instance after deletion
+	k8sClient          ctrlClient.Client
+	reconcilerResultCh chan centralreconciler.ReconcilerResult
+	statusResponseCh   chan private.DataPlaneCentralStatus
+}
+
+func NewRuntime(devEndpoint string, clusterID string, k8sClient ctrlClient.Client) *Runtime {
+	client, err := fleetmanager.NewClient(devEndpoint, clusterID)
+	if err != nil {
+		glog.Fatal("failed to create fleetmanager client", err)
```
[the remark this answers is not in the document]
SimonBaeumer: Return the error now and exit on the caller with `glog.Fatal`

fleetshard/pkg/runtime/runtime.go
SimonBaeumer: This needs more polishing, edit a To Do comment

fleetshard/pkg/centralreconciler/reconciler.go
SimonBaeumer: Good catch, renamed to `InputChannel`

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -4,56 +4,68 @@ import (
...
+	stopCh     chan struct{}
+	status     *int32
+	responseCh chan private.DataPlaneCentralStatus
+}
 
-	statuses := make(map[string]private.DataPlaneCentralStatus)
-	for _, remoteCentral := range list.Items {
-		glog.Infof("received cluster: %s", remoteCentral.Metadata.Name)
+// TODO: Correct error creation?
+type ReconcilerResult struct {
```
kovayur, line 34: `ReconcileResult`?

mtesseract (approved):

fleetshard/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,99 @@
...
+
+		// Start for each Central its own reconciler which can be triggered by sending a central to the receive channel.
+		for _, central := range list.Items {
+			if _, ok := r.reconcilers[central.Metadata.Name]; !ok {
+				r.reconcilers[central.Metadata.Name] = centralreconciler.NewCentralReconciler(r.k8sClient, central, r.reconcilerResultCh)
+				go r.reconcilers[central.Metadata.Name].Start()
+			}
+
+			reconciler := r.reconcilers[central.Metadata.Name]
+			reconciler.InputChannel() <- &central
```
kovayur, line 74: Instead of using `resultCh` and `inputCh` why don't just call the `Reconcile` function in a goroutine here:

```
for _, central := range list.Items {
	if _, ok := r.reconcilers[central.Metadata.Name]; !ok {
		r.reconcilers[central.Metadata.Name] = centralreconciler.NewCentralReconciler(r.k8sClient, r.client, central, r.reconcilerResultCh)
	}

	reconciler := r.reconcilers[central.Metadata.Name]
	go func() {
		status, err := reconciler.Reconcile(context.Background(), central)
		r.handleResult(central, status, err)
	}()
}
```

Initaly I thought about an asynchronous [pipeline](https://go.dev/blog/pipelines) which consists of 2 stages:
1) Perform reconciliation and return the status to the `runtime`
2) Report the status to the fleet manager

But after reading about how it can be implemented in go I ended up with this. 
I think that's even simpler. What do you think?

fleetshard/pkg/centralreconciler/reconciler.go
SimonBaeumer: done

PR #60, opened 2022-05-30
Drop quay creds overrides in build_deploy.sh

kurlov:
## Description

`build_deploy.sh` should rely on `QUAY_USER` and `QUAY_TOKEN`.
Also, image tag should be 7 chars long.

Not relevant

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -29,7 +29,7 @@
 # the built images can be pushed to quay.io.
 
 # The version should be the short hash from git. This is what the deployment process expects.
-VERSION="$(git log --pretty=format:'%h' -n 1)"
+VERSION=`git rev-parse --short=7 HEAD`
 
 # Set the directory for docker configuration:
 DOCKER_CONFIG="${PWD}/.docker"
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -59,7 +59,7 @@ if [ -z "${QUAY_TOKEN}" ]; then
 fi
 if [ -z "${QUAY_IMAGE_REPOSITORY}" ]; then
   echo "The quay.io image repository hasn't been provided."
-  echo "Make sure to set the QUAY_IMAGE_REPOSITORY environment variable."
+  echo "Make sure to set the QUAY_IMAGE_REPOSITORY environment variable to '<org>/<repo>'."
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


build_deploy.sh
```diff
@@ -78,15 +73,15 @@ else
 fi
 
 # Push the image:
-echo "Quay.io user and token is set, will push images to $QUAY_ORG org"
+echo "Quay.io user and token is set, will push images to $QUAY_IMAGE_REPOSITORY"
```
porridge, line 76: `$QUAY_IMAGE_REPOSITORY` has no default but is required.
Please add an `if [ -z` block like for the other `QUAY_...` vars.

Also, how do we set it in jenkins? 🤔

porridge: I think we can follow this example:
```
data/services/cincinnati/cicd/ci-int/jobs.yaml:        build_deploy_script_path: env GRAPH_URL=https://api.stage.openshift.com:443/api/upgrades_info/v1/gra
ph ./dist/run_containerized_loadtesting.sh
```

kurlov: We will set env var for it in Jenkins

build_deploy.sh
```diff
@@ -29,12 +29,7 @@
 # the built images can be pushed to quay.io.
 
 # The version should be the short hash from git. This is what the deployment process expects.
```
porridge, line 31:
```suggestion
# The version should be a 7-char hash from git. This is what the deployment process in app-interface expects.
```

build_deploy.sh
```diff
@@ -54,12 +49,17 @@ cd "${LINK}"
...
 fi
 if [ -z "${QUAY_TOKEN}" ]; then
   echo "The quay.io push token hasn't been provided."
-  echo "Make sure to set the QUAY_USER_PASSWORD environment variable."
+  echo "Make sure to set the QUAY_TOKEN environment variable."
+  exit 1
+fi
+if [ -z "${QUAY_IMAGE_REPOSITORY}" ]; then
+  echo "The quay.io image repository hasn't been provided."
+  echo "Make sure to set the QUAY_IMAGE_REPOSITORY environment variable."
```
porridge, line 62:
```suggestion
  echo "Make sure to set the QUAY_IMAGE_REPOSITORY environment variable to '<org>/<repo>'."
```

porridge (approved):
Just a couple of nitpicks.

PR #73, opened 2022-05-30
enable setting operator versions in dataplane config file

johannes94:
## Description
In order to get a central resource in fleet manager to transition from `accepted` to `provisioning` you have to set available operator version for a cluster. Until now it was expected that this would be done by fleet shard components with a PUT request to `/api/rhacs/v1/<cluster-id>/status` when they are deployed and ready on the data plane cluster. Since we're bringing the clusters in manually, it makes sense to also allow setting this configuration with the `dataplane-cluster-configuration.yaml` file. When using the new introduced configuration:
```
...
   available_central_operator_versions:
    - version: "0.1.0"
      ready: true
      central_versions:
       - version: "0.1.0"
```
The cluster will be in `ready` state and have an operator version available, thus there is no further preparation necessary to start with proper central reconciliation.

## Test manual

Manual testing:

```bash

# do this for the config file for the cluster you want to test it with
cp dev/config/dataplane-cluster-configuration-dockerdesktop.yaml ./config/dataplane-cluster-configuration.yaml
make run

# after leader election cluster should be in database looking like this:
make db/login
select * from clusters;

-[ RECORD 1 ]------------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------
id                                   | caaspvq87d5jrvreqb60
created_at                           | 2022-05-31 08:08:31.23905+00
updated_at                           | 2022-05-31 08:08:31.23905+00
deleted_at                           | 
cloud_provider                       | standalone
cluster_id                           | 1234567890abcdef1234567890abcdef
external_id                          | 
multi_az                             | f
region                               | standalone
status                               | ready
identity_provider_id                 | 
cluster_dns                          | kubernetes.docker.internal
provider_type                        | standalone
provider_spec                        | 
cluster_spec                         | 
available_dinosaur_operator_versions | \x5b7b2276657273696f6e223a22302e312e30222c227265616479223a747275652c2263656e7472616c56657273696f6e73223a5b7b2276657273696f6e223a22302e312e30227d5d7d5d
supported_instance_type              | eval,standard
```

```
# To run tests locally run:
```

### DIFF
--- a/dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
+++ b/dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
@@ -17,27 +17,17 @@
 #    provider_type: "ocm" #Valid values are `ocm` and `standalone`. `ocm` will be used if not specified.
 #    cluster_dns: apps.example.com #Valid cluster DNS. This will be used to build dinosaur host url and to communicate with standalone clusters. Required when "provider_type" is "standalone" 
 #    supported_instance_type: "eval" # could be "eval", "standard" or both i.e "standard,eval" or "eval,standard". Defaults to "standard,eval" if not set 
-clusters:
-#  - name: docker-desktop  # Uncomment if using docker-desktop
-#    cluster_id: 1234567890abcdef1234567890abcdef
-#    cloud_provider: standalone
-#    region: standalone
-#    schedulable: true
-#    status: ready
-#    provider_type: standalone
-#    supported_instance_type: "eval,standard"
-#    cluster_dns: kubernetes.docker.internal
-#    dinosaur_instance_limit: 5
- - name: minikube  # Uncomment if using minikube
+clusters:   # For a list of development clusters see dev/config/dataplane-cluster-configuration.yaml
+ - name: docker-desktop  # Uncomment if using docker-desktop
    cluster_id: 1234567890abcdef1234567890abcdef
    cloud_provider: standalone
    region: standalone
    schedulable: true
    status: ready
-   dinosaur_instance_limit: 5
    provider_type: standalone
    supported_instance_type: "eval,standard"
-   cluster_dns: cluster.local
+   cluster_dns: kubernetes.docker.internal
+   dinosaur_instance_limit: 5
    available_dinosaur_operator_versions:
     - version: "0.1.0"
       ready: true
--- a/dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
+++ b/dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
@@ -1,24 +1,11 @@
 ---
 #- A list of clusters for fleet manager
 #- The `cluster_id` field can not be empty
-#- All clusters in fleet manager DB already but are missing in the list will be marked as
+#- All clusters which are already stored in fleet-manager's DB but are missing in the list will be marked as
 #  'deprovisioning' and will later be deleted.
 #- This list is ordered, any new cluster should be appended at the end.
-#e.g.:
-#clusters:
-#  - name: anyname # This field is required for a standalone cluster i.e when the provider_type is "standalone". The value has to match the cluster / context name in the given kubeconfig file via the `--kubeconfig` flag.
-#    cluster_id: 1jp6kdr7k0sjbe5adck2prjur8f39378  #This field is required
-#    cloud_provider: aws
-#    region: us-east-1
-#    multi_az: true
-#    schedulable: true
-#    dinosaur_instance_limit: 2
-#    status: "cluster_provisioning" #Valid values are `cluster_provisioning`, `cluster_provisioned` and `ready`. `cluster_provisioning` will be used if not specified.
-#    provider_type: "ocm" #Valid values are `ocm` and `standalone`. `ocm` will be used if not specified.
-#    cluster_dns: apps.example.com #Valid cluster DNS. This will be used to build dinosaur host url and to communicate with standalone clusters. Required when "provider_type" is "standalone" 
-#    supported_instance_type: "eval" # could be "eval", "standard" or both i.e "standard,eval" or "eval,standard". Defaults to "standard,eval" if not set 
-clusters:   # For a list of development clusters see dev/config/dataplane-cluster-configuration.yaml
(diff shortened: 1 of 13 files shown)


dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
```diff
@@ -0,0 +1,35 @@
...
+   schedulable: true
+   status: ready
+   provider_type: standalone
+   supported_instance_type: "eval,standard"
+   cluster_dns: kubernetes.docker.internal
+   dinosaur_instance_limit: 5
+   available_dinosaur_operator_versions:
+    - version: "0.1.0"
+      ready: true
+      dinosaur_versions:
```
mtesseract, line 34: Personally I would prefer getting rid of these occurrences of "dinosaurs" incrementally.

internal/dinosaur/pkg/config/dataplane_cluster_config.go
```diff
@@ -97,17 +97,18 @@ func NewDataplaneClusterConfig() *DataplaneClusterConfig {
...
+	CloudProvider                     string                        `yaml:"cloud_provider"`
+	Region                            string                        `yaml:"region"`
+	MultiAZ                           bool                          `yaml:"multi_az"`
+	Schedulable                       bool                          `yaml:"schedulable"`
+	DinosaurInstanceLimit             int                           `yaml:"dinosaur_instance_limit"`
+	Status                            api.ClusterStatus             `yaml:"status"`
+	ProviderType                      api.ClusterProviderType       `yaml:"provider_type"`
+	ClusterDNS                        string                        `yaml:"cluster_dns"`
+	SupportedInstanceType             string                        `yaml:"supported_instance_type"`
+	AvailableDinosaurOperatorVersions []api.DinosaurOperatorVersion `yaml:"available_dinosaur_operator_versions"`
```
mtesseract, line 111: There might have been discussions around this which I have missed, but especially for things like public struct fields I would prefer avoiding the "dinosaur" terminology.

dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
```diff
@@ -0,0 +1,35 @@
+---
+#- A list of clusters for fleet manager
+#- The `cluster_id` field can not be empty
+#- All clusters in fleet manager DB already but are missing in the list will be marked as
+#  'deprovisioning' and will later be deleted.
+#- This list is ordered, any new cluster should be appended at the end.
```
mtesseract, line 6: What exactly does it mean for list being ordered? What would happen if a new cluster is inserted at the beginning or in the middle?

mtesseract (commented):
Looking good to me, just have minor stylistic remarks and two questions:

First: Do we have an idea already on how a dinosaur operator would look like? I was under the impression that we don't know yet what kind of operator components we need on the dataplane clusters.

Second: I understand that this change allows to register a new cluster in `ready` state with a certain dinosaur operator version being set. Somewhere in the comments for this PR I would have expected some remark along the lines of saying that when registering clusters in `ready` state it is your own responsibility to also make sure you have set up the claimed operator version on the dataplane cluster, because this is nothing the fleet-manager would do for you (right?).

Does this make sense?

dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
johannes94: Tbh. I don't know. The content of this file was copied from the dataplane-cluster-configuration-minikube.yaml. I did not check all the comments when I was working on this.

internal/dinosaur/pkg/config/dataplane_cluster_config.go
johannes94: You're right that makes sense. I'm going to change that.

dev/config/dataplane-cluster-configuration-dockerdesktop.yaml
ebensh: I believe this *may* be a way to detect removal of a cluster to trigger deprovisioning and deletion - addition should happen at the end, but **I haven't confirmed this in code**.

internal/dinosaur/pkg/config/dataplane_cluster_config_test.go
```diff
@@ -19,6 +20,11 @@ dinosaur_instance_limit: 5
...
 supported_instance_type: "eval,standard"
 cluster_dns: cluster.local
+available_dinosaur_operator_versions:
+  - version: "0.1.0"
+    ready: true
+    dinosaur_versions:
+      - version: "0.1.0"
 `)
 
 	c := ManualCluster{}
```
ebensh, line 30: I can't add on line 34 but I'd expect the test to verify that the available operator version is indeed 0.1.0

ebensh (changes requested):

internal/dinosaur/pkg/config/dataplane_cluster_config_test.go
johannes94: Added some verification here in last commit.

internal/dinosaur/pkg/config/dataplane_cluster_config.go
johannes94: Changed dinosaur occurrences as far as it made sense in context of this PR.

ebensh (approved):

mtesseract (approved):

internal/dinosaur/pkg/workers/clusters_mgr.go
```diff
@@ -637,6 +637,13 @@ func (c *ClusterManager) reconcileClusterWithManualConfig() []error {
 			ClusterDNS:            p.ClusterDNS,
 			SupportedInstanceType: p.SupportedInstanceType,
 		}
+
+		if len(p.AvailableCentralOperatorVersions) > 0 {
+			if err := clusterRequest.SetAvailableDinosaurOperatorVersions(p.AvailableCentralOperatorVersions); err != nil {
+				return []error{errors.Wrapf(err, "Failed to set operator versions for manual cluster %s with config file", p.ClusterId)}
+			}
+		}
```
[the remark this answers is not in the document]
mtesseract: Thanks for the hint!

PR #79, opened 2022-06-01
Provide docker compose environment for local testing for fleet-manager & fleetshard-sync

mtesseract:

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/dev/env/docker-compose/README.md
+++ b/dev/env/docker-compose/README.md
@@ -1,36 +1,59 @@
-# How to use
+# About
 
 This directory defines a docker compose based test environment which is easy to spin up.
 
-Make sure that no other DB is running:
-```
-$ make -C $(git rev-parse --show-toplevel) db/teardown
-```
+## Preprequisites
 
-Either source `dev/env/.envrc` or use `direnv` for loading this environment automatically.
-This will make sure that some settings are available in the current environment, which are
-then picked up by `docker-compose.yaml`. In particular, this includes OCM tokens and secrets
-from `secrets/`. If the OCM token needs to be refreshed, re-source `dev/env/.envrc` or use
-`direnv reload`.
+1. A Docker daemon needs to be running and the `docker` CLI needs to be installed.
+1. Setup [direnv](https://direnv.net/). Note: This is the recommended approach since `direnv` provides automatic environment loading and unloading for your shell, but this is not a strict requirement; if desired it is possible to load the required shell environments manually with `source`.
 
-A Docker daemon needs to be running and the `docker` CLI needs to be installed.
+## Spinning up Test Environment
 
-Spin up the test environment using:
-```
-$ docker compose up
-```
+1. Make sure that no other DB is running:
+    ```
+    $ make -C $(git rev-parse --show-toplevel) db/teardown
+    ```
 
-The `fleet-manager` service will be reachable at `localhost:8000`.
+1. Enter this directory:
+    ```
+    $ cd $(git rev-parse --show-toplevel)/dev/env/docker-compose
+    ```
+    Initial setup of `direnv` requires whitelisting of the provided test environment:
+    ```
+    $ direnv allow
+    ```
+    This will make sure that certain settings are available in the current environment, which are
+    then picked up by `docker-compose.yaml`. In particular, this includes OCM tokens and secrets
+    from `secrets/`. If the OCM token needs to be refreshed, use `direnv reload`.
 
-If new Docker images for `fleet-manager` and `fleetshard-sync` shall be built, use
-`docker compose up --build`.
+1. Spin up the test environment using:
+    ```
+    $ docker compose up
+    ```
+    The `fleet-manager` service will be reachable at `localhost:8000`. Single data-plane cluster is configured
+    as (see `config/dataplane-cluster-configuration.yaml` in this directory, not in the top-level directory of this repository).
 
-To shut down the test environment:
-```
-$ docker compose down
-```
+1. If new Docker images for `fleet-manager` and `fleetshard-sync` shall be built, use:
+    ```
+    $ docker compose up --build
+    ```
+    instead.
 
-If the Postgres volume shall also be deleted:
+1. To shut down the test environment:
+    ```
+    $ docker compose down
+    ```
+
+1. If the Postgres volume shall also be deleted:
+    ```
+    $ docker compose down -v
+    ```
+
+## Executing API calls
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,16 +1,35 @@
...
+WORKDIR /workspace
+
+COPY go.mod ./
+COPY go.sum ./
+RUN go mod download
+
+COPY . ./
+
+RUN make clean binary
+
```
SimonBaeumer, line 19: Why was this change necessary?

dev/env/docker-compose/config/dataplane-cluster-configuration.yaml
```diff
@@ -0,0 +1,17 @@
+---
+clusters:
+  - name: docker-desktop # Uncomment if using docker-desktop
```
SimonBaeumer, line 3:
```suggestion
  - name: docker-compose # Uncomment if using docker-desktop
```

SimonBaeumer (changes requested):
Can you add a README in the `dev/docker-compose` dir with simple commands how to use this setup?

Dockerfile
mtesseract: With this change in place docker compose can also take care of building images. In the `docker-compose.yaml` the image to use for `fleet-manager` and `fleetshard-sync` is provided in terms of a `Dockerfile`, hence `docker compose`  tries to build the image using `docker build`. But this requires the `Dockerfile` to be self-contained.

Previously `docker build` using our `Dockerfile` simply picked up the two application binaries in the top-level directory (asuming that they are built for the correct target platform!) and included them in the image. This process required the two relevant binaries being built ahead of time using something like `make binary`, this is -- as far as I can tell - incompatible with what `docker compose` requires.

Besides satisfying this requirement for `docker compose` I also think that this approach is generally more reliable since it guarantees that the correct binaries are copied into the image -- good for build reproducibility.

dev/env/docker-compose/config/dataplane-cluster-configuration.yaml
mtesseract: changed

Dockerfile
```diff
@@ -1,16 +1,35 @@
...
+# install go 1.17.8
+RUN curl -O -J https://dl.google.com/go/go1.17.8.linux-amd64.tar.gz
+RUN tar -C /usr/local -xzf go1.17.8.linux-amd64.tar.gz
+RUN ln -s /usr/local/go/bin/go /usr/local/bin/go
+
+WORKDIR /workspace
+
+COPY go.mod ./
+COPY go.sum ./
+RUN go mod download
```
SimonBaeumer, line 14: It is recommended to use `go mod tidy`. This downloads dependencies and removes deps which are not necessary.

```suggestion
RUN go mod tidy
```

Dockerfile
```diff
@@ -1,16 +1,35 @@
...
+
+COPY . ./
+
+RUN make clean binary
```
SimonBaeumer, line 18: This might run the targets in parallel.

```
RUN make clean && make binary
```

dev/env/docker-compose/README.md
```diff
@@ -0,0 +1,36 @@
+# How to use
+
+This directory defines a docker compose based test environment which is easy to spin up.
+
+Make sure that no other DB is running:
+```
+$ make -C $(git rev-parse --show-toplevel) db/teardown
+```
+
+Either source `dev/env/.envrc` or use `direnv` for loading this environment automatically.
```
SimonBaeumer, line 10: Can you directly document the recommended behaviour as a list of commands?
It is more convenient to follow the doc with commands to execute and not taking my own decisions whether I should use `source` or `direnv`.

A small comment `# ...` which explains the reason of a command above it is enough.


Example: 

```
# Docker is required to be running

# Stop running postgres instances:
$ make -C $(git rev-parse --show-toplevel) db/teardown

# Set environment
$ cd acs-fleet-manager/env && direnv

# Spin up environment and build
$ docker compose up --build

# To tear down environment run:
$ docker compose down

# To additionally delete the postgres volume run:
$ docker compose down -v
```

SimonBaeumer (changes requested):

Dockerfile
mtesseract: Yes, but I doubt that this is the desired behaviour in this case. (Potentially) deleting previous build artifacts needs to come strictly before creating the same build artifacts.

Dockerfile
mtesseract: Not an expert on Golang build tools and procedures here, but can you quote a source for this claim?
My sources are:
* (the official Docker manual)[https://docs.docker.com/language/golang/build-images/]
* (Medium article)[https://petomalina.medium.com/using-go-mod-download-to-speed-up-golang-docker-builds-707591336888]
* (Codefresh article)[https://codefresh.io/docs/docs/learn-by-example/golang/golang-hello-world/]
among others.

The worry I am having regarding "go mod tidy" is that it can, as a side-effect, modify `go.sum`, which would render a given repo as dirty. If, as we do in StackRox, image tags are derived from Git revision information, this would produce undesired behaviour, I guess.

dev/env/docker-compose/README.md
mtesseract: Done.

dev/env/docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,71 @@
+version: "3"
```
kovayur, line 1: Wouldn't it be more convenient to have this file in the root directory?

dev/env/docker-compose/config/dataplane-cluster-configuration.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
kovayur, line 1: Can't we reuse the existing config files in the `$GITROOT/config` directory?

kovayur (commented):
Good job! Left just a few clarifying questions. Looking forward to trying docker-compose in action!

Dockerfile
SimonBaeumer: Yeah, a lot of docs point to `go mod download` 

`go mod tidy` ensures that the dependencies match the source. If `go.sum` changed after running `go mod tidy` the build should ideally fail because dependencies are out of date.
In stackrox builds often failed because of `go mod download` because deps were out of sync.

Compare docs between [tidy](https://go.dev/ref/mod#go-mod-tidy) and [download](https://go.dev/ref/mod#go-mod-download).

I can't find the github discussion anymore though on this topic...

SimonBaeumer (approved):

PR #84, opened 2022-06-02
Retrieve fleet-manager endpoint and cluster ID from environment

mtesseract:
## Description

Currently these values are hard-coded. Make them configurable using environment variables.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/config/config.go
+++ b/fleetshard/config/config.go
@@ -1,0 +1,31 @@
+package config
+
+import (
+	"time"
+
+	"github.com/caarlos0/env/v6"
+	"github.com/pkg/errors"
+)
+
+// Config contains this application's runtime configuration.
+type Config struct {
+	FleetManagerEndpoint string        `env:"FLEET_MANAGER_ENDPOINT" envDefault:"http://127.0.0.1:8000"`
+	ClusterID            string        `env:"CLUSTER_ID"`
+	RuntimePollPeriod    time.Duration `env:"RUNTIME_POLL_PERIOD" envDefault:"1s"`
+}
+
+// Load retrieves the current runtime configuration from the environment and returns it.
+func Load() (*Config, error) {
+	c := Config{}
+	if err := env.Parse(&c); err != nil {
+		return nil, errors.Wrapf(err, "Unable to parse runtime configuration from environment")
+	}
+	if c.ClusterID == "" {
+		return nil, errors.New("CLUSTER_ID unset in the environment")
+	}
+	if c.FleetManagerEndpoint == "" {
+		return nil, errors.New("FLEET_MANAGER_ENDPOINT unset in the environment")
+	}
+
+	return &c, nil
+}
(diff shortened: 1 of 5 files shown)


SimonBaeumer (approved):
Nice!

fleetshard/config/config.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/caarlos0/env/v6"
+	"github.com/pkg/errors"
+)
+
+// Config contains this application's runtime configuration.
+type Config struct {
+	FleetManagerEndpoint string        `env:"FLEET_MANAGER_ENDPOINT" envDefault:"http://127.0.0.1:8000"`
+	ClusterID            string        `env:"CLUSTER_ID"`
+	RuntimePollPeriod    time.Duration `env:"RUNTIME_POLL_PERIOD" envDefault:"1s"`
```
kovayur, line 14: I think 1s period is too small. 60s ?

fleetshard/main.go
```diff
@@ -28,12 +25,19 @@ func main() {
...
 		glog.Info("Unable to set logtostderr to true")
 	}
 
-	runtime, err := runtime.NewRuntime(devEndpoint, clusterID, k8s.CreateClientOrDie())
+	config, err := config.Load()
+	if err != nil {
+		glog.Fatalf("Failed to load configuration: %v", err)
+	}
+
+	glog.Infof("Starting application with configuration: %+v\n", *config)
```
kovayur, line 38: If in a future we'll add sensitive parameters (secrets), is it ok to print them? 
I mean, such logging may be easily forgotten which may lead to an audit issue.
Is this valid in our case after all?

fleetshard/config/config.go
kovayur, line 14: WDYT about `FleetManagerPollPeriod`? I think the intention was to stick to the place where it is used, but out of this context the environment variable name sounds a bit confusing to me.

kovayur (commented):
Looks good. Just a few clarifying questions from me.

fleetshard/config/config.go
mtesseract: I don't know, you tell me. :) I just copied what the runtime used to have as a delay until now.

fleetshard/main.go
mtesseract: You have a point there. I like application printing out the configuration during startup, but, you are right, this should be done in a less ad-hoc manner. Will change and make it a bit more verbose.

fleetshard/config/config.go
mtesseract: This was @SimonBaeumer's idea. I kind of wanted to keep myself out of this naming decision. I can also use your idea.

kovayur: Thanks, let's keep it for now

fleetshard/main.go
kovayur: Thanks!

fleetshard/config/config.go
kovayur: No objections, let's keep as is

kovayur (approved):
LGTM

PR #102, opened 2022-06-07
added cache to fleetshard-sync reconciler

johannes94:
## Description
Implemented a cache in fleetshard-sync to make sure ManagedCentrals, that have not been changed since the last reconciliation loop, are not reconciled again.

## Test manual

Manual testing:
- Run db, fleet-manager and stackrox operator
- Run fleetshard-sync in debug mode using your IDE
- Set Breakpoint at `fleetshard/pkg/runtime/runtime.go:86` (1) and `fleetshard/pkg/runtime/runtime.go:89` (2)
- Create central resource with POST requests to fleetmanager
- Wait for first reconcilation, it should run into breakpoint 2 
- Next reconcilations should always run into breakpoint 1
- change host field in db table `update dinosaur_requests set host='cachtest1.cluster.local';`
- Next reconcilation should run into breakpoint 2
- Further reconcilations should again run into breakpoint 1

```
# To run tests locally run:
```


fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -28,23 +32,28 @@ const (
...
-func (r CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.ManagedCentral) (*private.DataPlaneCentralStatus, error) {
+func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.ManagedCentral) (*private.DataPlaneCentralStatus, error) {
 	// Only allow to start reconcile function once
 	if !atomic.CompareAndSwapInt32(r.status, FreeStatus, BlockedStatus) {
 		return nil, errors.New("Reconciler still busy, skipping reconciliation attempt.")
 	}
 	defer atomic.StoreInt32(r.status, FreeStatus)
 
+	if err := r.setLastCentralHash(remoteCentral); err != nil {
+		return nil, errors.Wrapf(err, "filling central reconcilation cache")
```
SimonBaeumer, line 54:
```suggestion
		return nil, errors.Wrapf(err, "setting central reconcilation cache")
```

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -28,23 +32,28 @@ const (
...
 // TODO(create-ticket): Add cache to only reconcile if a change to the ManagedCentral was made
-func (r CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.ManagedCentral) (*private.DataPlaneCentralStatus, error) {
+func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.ManagedCentral) (*private.DataPlaneCentralStatus, error) {
 	// Only allow to start reconcile function once
 	if !atomic.CompareAndSwapInt32(r.status, FreeStatus, BlockedStatus) {
 		return nil, errors.New("Reconciler still busy, skipping reconciliation attempt.")
 	}
 	defer atomic.StoreInt32(r.status, FreeStatus)
 
+	if err := r.setLastCentralHash(remoteCentral); err != nil {
```
SimonBaeumer, line 53: **Check Central changed:**

If the Central has not changed the reconcile must not be executed and can safely skipped in all cases. 
Implementing it in the caller (`runtime.go`) delegates the implementation of a changed-check to the caller.
Ideally this would be determined in fleet-manager but is more complex in its current implementation due to the state-machine and potential consistency problems.
Instead of setting the Central hash here implement the `centralChanged` check and skip the reconcile.

**Setting the hash:**

If the reconcile fails the `lastCentralHash` represents a Central state which was never applied to the server.
It should be set after the update/create was successful was successful.

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -100,6 +109,27 @@ func (r CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.
...
+	return !bytes.Equal(r.lastCentralHash[:], currentHash[:]), nil
+}
+
+func (r *CentralReconciler) setLastCentralHash(central private.ManagedCentral) error {
+	centralBytes, err := json.Marshal(&central)
+	if err != nil {
+		return err
+	}
+
+	r.lastCentralHash = md5.Sum(centralBytes)
```
SimonBaeumer, line 129: Duplicate code here. Can you extract it into a `fleetshard/pkg/util/hash` package with a func something like `md5.SumFromJSONStruct`?

fleetshard/pkg/runtime/runtime.go
```diff
@@ -76,6 +76,16 @@ func (r *Runtime) Start() error {
 
 			reconciler := r.reconcilers[central.Metadata.Name]
 			go func(reconciler *centralreconciler.CentralReconciler, central private.ManagedCentral) {
+				changed, err := reconciler.CentralChanged(central)
```
SimonBaeumer, line 79: Move implementation to `Reconcile` in runtime.go

fleetshard/pkg/centralreconciler/reconciler_test.go
```diff
@@ -71,3 +72,61 @@ func TestReconcileUpdateSucceeds(t *testing.T) {
 	assert.Equal(t, centralName, central.GetName())
 	assert.Equal(t, "4", central.GetAnnotations()[revisionAnnotationKey])
 }
+
+func TestCentralChanged(t *testing.T) {
+
+	tests := []struct {
+		name           string
```
SimonBaeumer, line 79: You can use a map indexed by a string which represents the test name.

fleetshard/pkg/centralreconciler/reconciler_test.go
```diff
@@ -71,3 +72,61 @@ func TestReconcileUpdateSucceeds(t *testing.T) {
 	assert.Equal(t, centralName, central.GetName())
 	assert.Equal(t, "4", central.GetAnnotations()[revisionAnnotationKey])
 }
+
+func TestCentralChanged(t *testing.T) {
```
SimonBaeumer, line 76: Can you add this as a test over the complete `Reconcile` func? 
Please include a test which checks that the hash is not updated when the reconcile failed.

SimonBaeumer (changes requested):

fleetshard/pkg/centralreconciler/reconciler.go
johannes94: Moved it to runtime, because I did not know how to handle cache hit in Reconcile (what to return). What do you think? Should it return an error saying that it didn't change? Should it return the last return value? After all the code in runtime will send a request to fleet-manager with the return value as a status when I don't change the code there.

fleetshard/pkg/centralreconciler/reconciler.go
johannes94: Done.

fleetshard/pkg/centralreconciler/reconciler_test.go
johannes94: Did it this way, because this pattern is used by other table driven tests in this repo. But I don't mind to change it here If you think it's better.

fleetshard/pkg/runtime/runtime.go
johannes94: See comment above.

fleetshard/pkg/centralreconciler/reconciler_test.go
johannes94: Done.

fleetshard/pkg/centralreconciler/reconciler_test.go
```diff
@@ -71,3 +73,108 @@ func TestReconcileUpdateSucceeds(t *testing.T) {
...
+		central: private.ManagedCentral{},
+	}
+
+	expectedHash, err := util.MD5SumFromJSONStruct(&simpleManagedCentral)
+	require.NoError(t, err)
+
+	_, err = r.Reconcile(context.TODO(), simpleManagedCentral)
+	require.NoError(t, err)
+
+	assert.Equal(t, expectedHash, r.lastCentralHash)
```
SimonBaeumer, line 121: After the last assertion please add another call to `Reconcile` to check that the function execution was skipped.
Only possible if the check was moved from `runtime.Runtime` to the `Reconcile` func.

fleetshard/pkg/centralreconciler/reconciler.go
johannes94: Done

fleetshard/pkg/centralreconciler/reconciler_test.go
```diff
@@ -71,3 +73,111 @@ func TestReconcileUpdateSucceeds(t *testing.T) {
...
+
+	_, err = r.Reconcile(context.TODO(), simpleManagedCentral)
+	require.Error(t, err)
```
SimonBaeumer, line 124: Please also assert the error message

fleetshard/pkg/util/hash.go
```diff
@@ -0,0 +1,18 @@
...
+func MD5SumFromJSONStruct(in interface{}) ([16]byte, error) {
+	var sum [16]byte
+
+	b, err := json.Marshal(in)
+	if err != nil {
+		return sum, err
+	}
+
+	sum = md5.Sum(b)
+	return sum, nil
```
SimonBaeumer, line 17: nit: you can omit the temporary variable.

```suggestion
	b, err := json.Marshal(in)
	if err != nil {
		return [16]byte{}, err
	}
	return md5.Sum(b), nil
```

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -28,24 +31,34 @@ const (
...
 
 	remoteNamespace := remoteCentral.Metadata.Name
+
+	changed, err := r.centralChanged(remoteCentral)
+	if err != nil {
+		return nil, errors.Wrapf(err, "checking if central changed")
+	}
+
+	if !changed {
+		return nil, errors.New("Central not changed, skipping reconcilation")
```
SimonBaeumer, line 59: Did you wanted to return `nil` instead of an error here?
Otherwise create a separate error type and check it with `error.Is` on the caller side to omit the action handler.

fleetshard/pkg/centralreconciler/reconciler_test.go
johannes94: Done.

fleetshard/pkg/util/hash.go
johannes94: Done

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -25,27 +28,39 @@ const (
 	k8sManagedByLabelKey  = "app.kubernetes.io/managed-by"
 )
 
+var ErrTypeCentralNotChanged = errors.New("central not changed, skipping reconcilation")
```
SimonBaeumer, line 31: Comment on exposed variable.
Thought the linter may be angry about it, but the code base does not follow it though initially.

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -100,6 +119,26 @@ func (r CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.
 	}, nil
 }
 
+// centralChanged compares the given central to the last central Reconciled using a hash
```
SimonBaeumer, line 122:
```suggestion
// centralChanged compares the given central to the last central reconciled using a hash
```

fleetshard/pkg/util/hash.go
```diff
@@ -0,0 +1,15 @@
+package util
+
+import (
+	"crypto/md5"
+	"encoding/json"
+)
+
+func MD5SumFromJSONStruct(in interface{}) ([16]byte, error) {
```
SimonBaeumer, line 8: Comment on exposed function pls

fleetshard/pkg/centralreconciler/reconciler_test.go
```diff
@@ -71,3 +73,111 @@ func TestReconcileUpdateSucceeds(t *testing.T) {
...
 }
+
+func TestReconcileLastHashEmptyOnError(t *testing.T) {
+	fakeClient := testutils.NewFakeClientBuilder(t, &v1alpha1.Central{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      centralName,
+			Namespace: centralName,
+			// most straighforward way to simulat Reconcile returning an error
+			// is setting this value to something that is not a number
+			Annotations: map[string]string{revisionAnnotationKey: "NaN"},
```
SimonBaeumer, line 84: You can extract "NaN" to a constant named "invalidRevision" or add an invalid text directly to it to avoid the comment.
For example:

```suggestion
			Annotations: map[string]string{revisionAnnotationKey: "invalid annotation"},
```

fleetshard/pkg/centralreconciler/reconciler_test.go
```diff
@@ -71,3 +73,111 @@ func TestReconcileUpdateSucceeds(t *testing.T) {
 	assert.Equal(t, centralName, central.GetName())
 	assert.Equal(t, "4", central.GetAnnotations()[revisionAnnotationKey])
 }
+
+func TestReconcileLastHashEmptyOnError(t *testing.T) {
```
SimonBaeumer, line 77: Would it be more correct to say "hash not updated on error" instead of "empty"?

SimonBaeumer (changes requested):

fleetshard/pkg/centralreconciler/reconciler.go
johannes94: Thanks for pointing that out, default linter in VS code (since it was changed to use gopls) was no longer complaining about such things. Changed it to `golint` now and added the comment.

fleetshard/pkg/centralreconciler/reconciler.go
johannes94: Done.

fleetshard/pkg/util/hash.go
johannes94: Done.

fleetshard/pkg/centralreconciler/reconciler_test.go
johannes94: Done.

fleetshard/pkg/centralreconciler/reconciler_test.go
johannes94: Yes, fits better. I changed it.

SimonBaeumer (approved):
Nice, 🚢 it 👊

PR #112, opened 2022-06-08
ROX-9843: Add Helm subchart for dataplane terraforming of observability

juanrh:
## Description

This is a port of the current status of the observability resources repo as a Helm sub-chart. The purpose of this PR is to just create that chart as an integration point between data plane terraform and observability, so the observability team can work on it autonomously.

## Test manual

Manual testing installing both the observability sub-chart and the parent terraforming chart. After the installation the 3 routes for grafana, alertmanager and prometheus are eventually visible after a while.

Installing and uninstalling the observability sub-chart - directory `dp-terraform/helm/rhacs-terraform/charts/observability`:
```bash
# render the chat to stdout for debugging
helm template rhacs-terraform-obs --debug --namespace rhacs --values ~/.rh/obs-values.yaml .

# install
helm upgrade --install rhacs-terraform-obs --namespace rhacs --values ~/.rh/obs-values.yaml .

# uninstall
helm uninstall rhacs-terraform-obs --namespace rhacs .
```

Installing and uninstalling the whole dataplane terraform chart - directory `dp-terraform/helm/rhacs-terraform`:
```bash
# render the chat to stdout for debugging
helm template rhacs-terraform --debug --namespace rhacs --values ~/.rh/terraform-values.yaml \
   --set fleetshardSync.ocmToken=$(ocm token) \
   --set fleetshardSync.fleetManagerEndpoint=${FM_ENDPOINT} \
   --set fleetshardSync.clusterId=${cluster_id} \
   --set acsOperator.enabled=true .

helm upgrade --install rhacs-terraform --namespace rhacs --values  ~/.rh/terraform-values.yaml \
   --set fleetshardSync.ocmToken=$(ocm token) \
   --set fleetshardSync.fleetManagerEndpoint=${FM_ENDPOINT} \
   --set fleetshardSync.clusterId=${cluster_id} \
   --set acsOperator.enabled=true .

helm uninstall rhacs-terraform --namespace rhacs .
```


dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-05-subscription.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    operators.coreos.com/observability-operator.rhacs-observability: ''
+  name: observability-operator
+  namespace: {{ include "observability.namespace" . }}
+spec:
+  channel: alpha
+  installPlanApproval: Automatic
+  name: observability-operator
+  source: observability-operator-manifests
+  sourceNamespace: rhacs-observability
```
mtodor, line 13: I think we should use a namespace where `CatalogSource` is defined, and in our case, it's a variable.
```suggestion
  sourceNamespace: {{ include "observability.namespace" . }}
```

dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-02-catalog-source.yaml
```diff
@@ -0,0 +1,9 @@
+
```
mtodor, line 1: nitpick: Empty line can be removed.

mtodor (commented):
@juanrh Awesome work! It looks great!

There is one additional thing. We have discussed that the current workaround to set custom names for sub-operators is not important at this point. This removes the need for the hack that we have created. This would require a lot of changes in PR and maybe we can do it in a follow-up PR once when the observability operator is fixed. I'm not sure what are your opinions? @stehessel @dhaus67

juanrh:
> We have discussed that the current workaround to set custom names for sub-operators is not important at this point. This removes the need for the hack that we have created. This would require a lot of changes in PR and maybe we can do it in a follow-up PR once when the observability operator is fixed. I'm not sure what are your opinions?

The goal of this PR is just setting an integration point for your team, I don't have context to evaluate what is required and what it's not. However, I agree it would be better to just use pure Helm and not rely on Makefile hacks, because currently the make targets `helm/post-install` and `helm/post-uninstal` are effectively an integration point between both charts. So I would be happy to remove the hack so the integration point is simpler. Do you have some code branch that I can use as reference for that? Thanks

mtodor:
I'll continue working on this PR. Thank you for a head start!

dp-terraform/helm/rhacs-terraform/charts/observability/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  repository: ""
+  tag: "master"
+
+# Credentials for Observatorium https://observatorium.io/ instance
+observatorium:
+  tenant: ""
+  gateway: ""
+  authType: ""
+  redHatSsoAuthServerUrl: ""
+  redHatSsoRealm: ""
```
dhaus67, line 23: How about we keep it consistent and specific the default values within the `values.yaml` instead of within the resources via `.Values.<some-value> | default "<some-default-value>"`? Seeing as we already do that for `tag, observabilityOperatorVersion` I personally find it a bit confusing when other default values are embedded within the template.

mtodor: @dhaus67 I agree that it's bad to have multiple solutions for the same functionality. My intention was to keep only relevant changes in the values file per different stages. But it seems that helm is already handling that with different specificity - https://helm.sh/docs/chart_template_guide/values_files/

With that in mind, I'll remove defaults from the parent chart values file.

dp-terraform/helm/rhacs-terraform/README.md
```diff
@@ -1,17 +1,47 @@
...
+Chart to terraform data plane OSD clusters.
 
 ## Usage
 
-The env var `FM_ENDPOINT` should point to an endpoint for the fleet manager. An option to use a fleet manager instance running in your laptop is to [setup ngrok](https://ngrok.com/docs/getting-started), launch the fleet manager, and run `ngrok http 8000` to expose it to the internet. That commands outputs an endpoint that you can use for `FM_ENDPOINT`.
+**Prepare environment variables**
 
-Install the chart as follows:
+The env var `FM_ENDPOINT` should point to an endpoint for the fleet manager. An option to use a fleet manager instance running in your laptop is to [setup ngrok](https://ngrok.com/docs/getting-started), launch the fleet manager, and run `ngrok http 8000` to expose it to the internet. That commands outputs an endpoint that you can use for `FM_ENDPOINT`.  
+To get the cluster id for staging use `export CLUSTER_ID=$(grep cluster_id ../../../dev/config/dataplane-cluster-configuration-staging.yaml | tail -n1 | tr -s ' '  | cut -d ' ' -f 3)`
```
stehessel, line 10: From what directory should this be executed? Maybe this command could be simplified, it looks like a prime candidate for breakage. Perhaps we should write it as "look up the cluster id in file `dev/config/dataplane-cluster-configuration-staging.yaml` and export as `CLUSTER_ID` env variable" or something like that.

dp-terraform/helm/rhacs-terraform/README.md
```diff
@@ -1,17 +1,47 @@
...
+
+Create a file `terraform-values.yaml` with the values for the parameters in [values.yaml](./values.yaml) that are missing or that you want to override. That file will contain credentials, so make sure you put it in a safe location, and with suitable permissions. 
+
+**Render the chart to see the generated templates during development**
+
+```bash
+helm template rhacs-terraform \
+  --debug \
+  --namespace rhacs \
+  --values ~/.rh/terraform-values.yaml \
```
stehessel, line 22: Why `~/.rh`? You didn't specify where the file should be created and it seems irrelevant. I would just remove that part.

dp-terraform/helm/rhacs-terraform/README.md
```diff
@@ -1,17 +1,47 @@
...
+  --set fleetshardSync.ocmToken=$(ocm token) \
```
stehessel, line 23: With what account should we `ocm login`?

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -9,3 +9,15 @@ fleetshardSync:
 acsOperator:
   enabled: false
   startingCSV: rhacs-operator.v3.70.0
+
+# See available parameters in charts/observability/values.yaml
+# - enabled flag is used to completely enable/disable observability sub-chart
+observability:
+  enabled: true
+  github:
+    accessToken: ""
```
stehessel, line 18: Why are these variables defined here and in `observability/values.yaml`? Who would win if the values are different?

mtodor: Based on documentation from helm: https://helm.sh/docs/chart_template_guide/values_files/. With specificity defined, it would be the value from the parent chart. So this file has priority over the sub-chart.

stehessel (approved):
Feel free to consider my remaining comments optional.

dp-terraform/helm/rhacs-terraform/README.md
juanrh: I wrote all commands in README files assuming they are run using the parent directory of the README as the working directory. But I'm also ok with your suggestion

dp-terraform/helm/rhacs-terraform/README.md
juanrh: Yes, that's irrelevant.

dp-terraform/helm/rhacs-terraform/README.md
juanrh: Currently you can use any RH SSO account, I use my RH login. When we complete the private API auth work (see https://app.slack.com/client/T030RBGDB/C0313JYKH8W/thread/C0313JYKH8W-1654083120.446879 for context) then we'll have to use a specific account specified specified in other Helm values introduced in https://github.com/stackrox/acs-fleet-manager/pull/124, and we'll remove `fleetshardSync.ocmToken` altogether. `fleetshardSync.ocmToken` it's not a long term solution because OCM tokens expire, as part of that auth work we'll need to modify the fleet shard sync to renew the token. That specific account is stored in a Helm values files that is available in the "Infrastructure & Accounts" wiki in the row "Helm values for dataplane terraforming chart".

dp-terraform/helm/rhacs-terraform/README.md
mtodor: I have changed this a bit.

PR #127, opened 2022-06-17
ROX-11309: Make liveness probe based on /api

kurlov:
## Description

Return liveness probe back to /api based


porridge (approved):

mtesseract:
Hi @kurlov, do you know what the state of this PR is? Should we rebase and re-review?

mtesseract:
Can this be closed @kurlov ?

porridge:
I don't think anyone ever did an in-depth analysis of what is the right approach here.
Although apparent lack of related problems so far hints that this indeed might be unnecessary.

kurlov:
I think it might make sense to have an /api based liveness probe. App might not be ready to serve traffic due to long migration or not yet connected to DB and in this case /api probe should fail

PR #128, opened 2022-06-20
ROX-11407: Configurable resource limits for Central and Scanner

mtesseract:

## Test manual

```
# To run tests locally run:
```


mtesseract:
Regarding the test failures:

I thought using `MustParse` from `"k8s.io/apimachinery/pkg/api/resource"` would be the canonical way for validating resource quantities. But apparently it is not quite providing the desired semantics: It doesn't capture the fact that plain numbers like "0.5" could also be provided for CPU resources instead of something like "500m", which would be validated correctly.

Hints regarding the correct validation logic welcome!

mtesseract:
Regarding your offline remark, @SimonBaeumer: As a quick way to proceed here we could add a runtime switch (e.g. an environment variable `ALLOW_CUSTOM_CENTRAL_RESOURCES`) for enabling support for custom resource settings. Defaulting to `false`.

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -63,12 +65,47 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 	remoteCentralName := remoteCentral.Metadata.Name
 	remoteNamespace := remoteCentral.Metadata.Namespace
 
+	centralResources, err := convertPrivateResourceRequirementsToCoreV1(&remoteCentral.Spec.Central.Resources)
```
johannes94, line 68: nit: add additional function to create complete central spec, would make it easier for me to focus on reconcile logic instead of conversion logic

internal/dinosaur/pkg/migrations/20220114114500_add_centrals.go
```diff
@@ -43,6 +43,8 @@ func addCentralRequest() *gormigrate.Migration {
 		Namespace                     string     `json:"namespace"`
 		RoutesCreationId              string     `json:"routes_creation_id"`
 		DeletionTimestamp             *time.Time `json:"deletionTimestamp"`
+		Central                       api.JSON   `json:"central"`
```
johannes94, line 46: We have to be careful with schema changes like this. This change would not be applied to the DB for the deployment in staging SRE cluster. If we want it to be applied we either have to again rollback and reapply migrations in the DB (what I did when I was merging the dinosaur change of this Model) or we would have to create another migration file, that adds the additional columns to the already existing table.

openapi/fleet-manager-private.yaml
```diff
@@ -22,7 +22,7 @@ tags:
 
 paths:
   # Endpoints for data plane communications
-  '/api/rhacs/v1/agent-clusters/{id}/status':
+  "/api/rhacs/v1/agent-clusters/{id}/status":
```
johannes94, line 25: Looks like a lot of quote changes here, do we default to single or double quotes in yaml files?

openapi/fleet-manager.yaml
```diff
@@ -22,7 +22,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/VersionMetadata'
+                $ref: "#/components/schemas/VersionMetadata"
```
johannes94, line 25: again a lot of quote changes.

internal/dinosaur/pkg/presenters/dinosaur.go
```diff
@@ -22,6 +24,21 @@ func ConvertDinosaurRequest(dinosaurRequestPayload public.CentralRequestPayload,
 
 // PresentDinosaurRequest - create CentralRequest in an appropriate format ready to be returned by the API
 func PresentDinosaurRequest(request *dbapi.CentralRequest) public.CentralRequest {
+	var central public.CentralSpec
+	var scanner public.ScannerSpec
+
+	if err := json.Unmarshal(request.Central, &central); err != nil {
+		// Assuming here that what is in the DB is guaranteed to conform to the expected schema.
+		// TODO: Add error propagation.
```
johannes94, line 32: Just want to make sure, you are still aware of the TODO's added.

johannes94 (changes requested):

openapi/fleet-manager-private.yaml
mtesseract: Yeah, that was annoying... It is what VSCode did without asking me. :-(

internal/dinosaur/pkg/presenters/dinosaur.go
mtesseract: Yes, thank you. Before fleshing this out, I'd suggest to first discuss the approach to take from a timeline perspective.

internal/dinosaur/pkg/migrations/20220114114500_add_centrals.go
mtesseract: Ah, yes, adding another migration file seems like the way to go for changes like this, I assume.

fleetshard/pkg/centralreconciler/reconciler.go
mtesseract: good point, thanks.

PR #147, opened 2022-06-27
Integrate pre-commit plugin detect-secrets

mtesseract:
## Description

See https://pre-commit.com/index.html.
This is an extensible mechanism for integrating pre-commit hooks.
I suggest to use the `detect-secrets` plugin. I have tried out `ripsecrets` but for some reason it did not catch some example secrets (high entropy secrets) I have added into the codebase for testing.

Installation instructions are behind the above link.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/Yelp/detect-secrets
+    rev: v1.2.0
+    hooks:
+      - id: detect-secrets
+        args: ["--baseline", ".secrets.baseline"]
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,712 @@
+{
+  "version": "1.2.0",
+  "plugins_used": [
+    {
+      "name": "ArtifactoryDetector"
+    },
+    {
+      "name": "AWSKeyDetector"
+    },
+    {
+      "name": "AzureStorageKeyDetector"
+    },
+    {
+      "name": "Base64HighEntropyString",
+      "limit": 4.5
+    },
+    {
+      "name": "BasicAuthDetector"
+    },
+    {
+      "name": "CloudantDetector"
+    },
+    {
+      "name": "GitHubTokenDetector"
+    },
+    {
+      "name": "HexHighEntropyString",
+      "limit": 3.0
+    },
+    {
+      "name": "IbmCloudIamDetector"
+    },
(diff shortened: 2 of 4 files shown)


SimonBaeumer:
Can you provide a script/make traget to automatically install the pre-commit hooks?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
```
SimonBaeumer, line 1: I prefer moving these configs into `/dev/githook` or similar. 
This keeps the root directory more clean.

mtesseract: As I understand it the path is fixed:
image

SimonBaeumer: Ah then nvm

mtesseract:
> Can you provide a script/make traget to automatically install the pre-commit hooks?

Thanks for the hint. I have noticed that there is in fact already a target `setup/git/hooks`.
Starting from that I attempted to also include the pre-commit hook under `.githooks` from the repo into the pre-commit.com tool.

But I have tried for 30 minutes now with no luck. Looking at the documentation I would expect something like this to work:
```
  - repo: local
    hooks:
      - id: custom
        name: custom
        entry: ./.githooks/pre-commit
        language: script
        pass_filenames: false
        always_run: true

```

But it doesn't get executed. I don't know why.
I would like to merge not be blocked by this now so that I will come back to this issue at a later point, ok?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
...
+  - repo: https://github.com/Yelp/detect-secrets
+    rev: v1.2.0
+    hooks:
+      - id: detect-secrets
+        args: ["--baseline", ".secrets.baseline"]
```
dhaus67, line 6: If you want to circumvent adding the existing githooks, which currently run `gofmt | go vet` against staged files, you can add the following to the config to achieve the same with pre-commit:
```yaml
  - repo: https://github.com/dnephin/pre-commit-golang
    rev: 96221dc
    hooks:
      - id: go-fmt
      - id: go-vet
```

mtesseract: I will look into this in a seperate PR, ok?
I have also noticed that there is a golangci-lint plugin, which might be a good fit for us.

SimonBaeumer (approved):

PR #148, opened 2022-06-27
Test OpenShift CI

mtesseract:

## Test manual

```
# To run tests locally run:
```


dev/env/scripts/up.sh
```diff
@@ -0,0 +1,107 @@
...
+
+Image: ${FLEET_MANAGER_IMAGE}
+Namespace: ${ACSMS_NAMESPACE}
+Inheriting ImagePullSecrets for Quay.io: ${INHERIT_IMAGEPULLSECRETS}
+
+EOF
+
+if [[ "$FLEET_MANAGER_IMAGE" =~ ^fleet-manager-.*/fleet-manager:.* ]]; then
+    # Local image reference, which cannot be pulled.
+    if ! $DOCKER image inspect "${FLEET_MANAGER_IMAGE}" >/dev/null 2>&1; then
```
johannes94, line 18: Is it on purpose that this is `$DOCKER` and not `docker`, script fails for me because no $DOCKER variable set, if I replace this with `docker` everything is working

mtesseract: Thanks, fixed.

SimonBaeumer (commented):
Can you add a description to the PR @mtesseract?

mtesseract:
> Can you add a description to the PR @mtesseract?

Done, @SimonBaeumer .

dev/env/scripts/enable-port-forwarding
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
johannes94, line 1: I'd rename the script e.g to `kubernets-port-forwarding` so it indicates which kind of forwarding is created. Also enable-port-forwarding indicates that you just can start port forwarding but not stop it.

johannes94 (commented):
Good job! Added some comments. When whole PR is ready for review pls ping me again. I also think it would make sense to restructure filenames/folder before merging. So it is more clear what scripts are general "util" scripts and what are specific to e2e tests execution.

dev/env/manifests/crds/01-platform.stackrox.io_centrals.yaml
```diff
@@ -0,0 +1,705 @@
+apiVersion: apiextensions.k8s.io/v1
```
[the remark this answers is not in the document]
mtesseract: I'm also not a fan of the duplication, but I haven't thought about clean/feasible ways to circumvent this, given that we need to be running in CI.

Are you having something in mind like CURLing the CRDs from the stackrox repo and injecting them into, for example, the src image?

dev/env/scripts/enable-port-forwarding
mtesseract: Thanks. Fine if I drop the "kubernetes-" prefix alltogether? Basically all these scripts are built for K8s.

dev/env/manifests/crds/01-platform.stackrox.io_centrals.yaml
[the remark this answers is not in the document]
johannes94: Yes, that's what I was thinking. Accessing the repo in CI shouldn't be a problem as the repo is public. We could download the files from a specific tag/release so that we're sure builds are reproducible and working with the correct files.

dev/env/scripts/enable-port-forwarding
johannes94: Yes 👍

dev/env/scripts/lib.sh
```diff
@@ -0,0 +1,107 @@
...
+
+log() {
+    echo "$*"
+}
+
+verify_environment() {
+    return
+}
+
+get_current_cluster_name() {
```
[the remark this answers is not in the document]
mtesseract: done, thanks

dev/env/manifests/cluster-type-minikube/04-rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: acsms-namespaced
+  namespace: "$ACSMS_NAMESPACE"
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: default
```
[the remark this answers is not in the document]
mtesseract: Thanks, I have removed the role and the rolebinding.

dev/env/manifests/crds/01-platform.stackrox.io_centrals.yaml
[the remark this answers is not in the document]
mtesseract: See https://github.com/stackrox/acs-fleet-manager/pull/173

.openshift-ci/tests/e2e-test.sh
```diff
@@ -0,0 +1,41 @@
...
+up.sh
+
+log "Environment up and running"
+log "Waiting for fleet-manager to complete leader election..."
+# Don't have a better way yet to wait until fleet-manager has completed the leader election.
+$KUBECTL -n "$ACSMS_NAMESPACE" logs -l application=fleet-manager -c fleet-manager -f |
+    grep --line-buffered 'Running as the leader and starting' |
+    head -1 >/dev/null || true
+sleep 1
+
```
johannes94, line 16: Probably it is due to my limited bash skills but I don't understand how this stops following logs on the first match, can you explain? I expected something like this:
```suggestion
$KUBECTL -n "$ACSMS_NAMESPACE" logs -l application=fleet-manager -c fleet-manager -f |
    grep --line-buffered -m1 'Running as the leader and starting' > /dev/null
sleep 1

```
Also why do we need the sleep afterwards?

dev/env/scripts/bootstrap.sh
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+
+export GITROOT="$(git rev-parse --show-toplevel)"
+source "${GITROOT}/dev/env/scripts/lib.sh"
+init
+
+if [[ "$CLUSTER_TYPE" == "minikube" ]]; then
+    if ! minikube status >/dev/null; then
+        unset DEBUG
```
johannes94, line 9: Should we set DEBUG again when minikube start is done?

dev/env/scripts/bootstrap.sh
```diff
@@ -0,0 +1,140 @@
...
+            --delete-on-failure=true
+    else
+        if ! kc_output=$($KUBECTL api-versions >/dev/null 2>&1); then
+            die "Sanity check for contacting Kubernetes cluster failed: ${kc_output}"
+        fi
+    fi
+fi
+
+# Create Namespaces.
+apply "${MANIFESTS_DIR}/shared"
```
johannes94, line 24: We could use a function with the namespace as an argument for this loops. Also maybe we could use [kubectl wait ](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#wait) do wait for the service accounts but I've not tested it.

.openshift-ci/tests/e2e-test.sh
mtesseract: I didn't know about `grep -m`, thanks! That's a simplification.
Well, with `head` it also works because head will stop reading after it has retrieved the first line (which is the first match), this will then terminate the whole pipe, all the way back to `kubectl logs`.

Thoughout this PR you might here and there see a `sleep 1`. Sometimes even in the presence of a technically more correct wait condition. These sleeps are there because they seem to prevent certain races/failures while bringing up the environment. It is totally not obvious why you need them, but I have sometimes seen failures without them and they seem to fix it. Of course it would be better to get rid of them and use proper wait conditions wherever possible.
For now I wouldn't care too much about them if it works.

mtesseract: I guess in this particular case the situation was that there was small time window after fleet-manager has finished its leader election during which API calls triggered by the tests would still fail.

dev/env/manifests/fleet-manager/02-fleet-manager-deployment.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
mtesseract: Let's remove the annotations.
I have heavily modified the files after generation with kompose.

For the future: At some would it would be nice to use some Helm chart under the hood which is also used for production deployments (but has some internal toggles for deploying the test environments). For now I decided to not depend on any coordinated Helm chart efforts in order to reduce the dependencies while everything is in flux.
But we are now at a point where requirements become clearer and at some point we should consolidate on a unified deployment method, I think.

dev/env/scripts/bootstrap.sh
mtesseract: I have simplify removed the DEBUG. It was an experiment. Remember the verbose debug messages during running the tests? I wanted to figure out where they were coming from. I  had the hypothesis that some code was simply testing for `DEBUG != ""`  in order to check if debugging mode should be turned on and so even our `DEBUG=false` would incorrectly enable debugging mode.

The hypothesis was correct but it wasn't minikube. It was the port-forwarding started with `kubectl port-forwarding`. :-D 

How I found out: searched on whole GitHub for the specific debug  messages and found some streaming library emitting these messages. From that I was able to identify the port-forwarding as the culprit.

dev/env/scripts/bootstrap.sh
mtesseract: I am not aware of how to implement something like "wait for resource to *appear*" with `kubectl wait`. Most (all?) of the wait conditions are designed for pod resources.

I have put the waiting code into a function.

dev/env/scripts/lib.sh
```diff
@@ -0,0 +1,191 @@
...
+
+    log "Waiting for pod ${pod_selector} within namespace ${namespace} to become ready..."
+
+    for i in $(seq 10); do
+        if $KUBECTL -n "$ACSMS_NAMESPACE" wait --timeout=5s --for=condition=ready pod -l "$pod_selector" 2>/dev/null >&2; then
+            break
+        fi
+        sleep 1
+    done
+    $KUBECTL -n "$ACSMS_NAMESPACE" wait --timeout=20s --for=condition=ready pod -l "$pod_selector"
```
johannes94, line 160: I'm confused about the different waiting mechanisms here. Why the loop and why running the same kubectl wait command again after the loop? Wouldn't a single wait with higher timeout be enough?
```suggestion
    $KUBECTL -n "$ACSMS_NAMESPACE" wait --timeout=60s --for=condition=ready pod -l "$pod_selector"
```

dev/env/scripts/lib.sh
```diff
@@ -0,0 +1,191 @@
...
+    return 1
+}
+
+wait_for_default_service_account() {
+    local namespace="$1"
+    if wait_for_resource_to_appear "$namespace" "serviceaccount" "default"; then
+        return 0
+    else
+        return 1
+    fi
```
johannes94, line 190: I think that should also work instead of the if (not tested)
```suggestion
    wait_for_resource_to_appear "$namespace" "serviceaccount" "default"
    return $?
```

dev/env/scripts/lib.sh
johannes94: I see this has changed, but what I meant was that we could remove the for loop completely, now we have both. Is there a reason for this, that I'm not aware of?

dev/env/scripts/lib.sh
mtesseract: The "if" is for protecting the script from terminating due to the use of "set -e".
I agree this sounds weird, but the command within the if-condition is protected from early termination here and I don'g know of a better way.

dev/env/scripts/lib.sh
mtesseract: I can explain. Again, I agree this seems weird, but I found this solution to be the most reliable from the solutions I have tried.

The problem is that the pods do not exist initially, therefore a plain `kubectl wait` can in fact fail immediately (hence terminating the script). This is of course not what we want.

I *could* just add something like `sleep 10` (and guessing a reasonable delay value), but what I am trying here is a little bit smarter: I try multiple times (10) to wait for a short amount of time. When the pod is not there, the wait will fail, I will sleep and try again to wait. Also, if the pod is already there, I don't want to wait for too long, because the *actual* kubectl wait comes after to loop. 

Probably we can modify this so that we do it in two *different* steps. First we wait for the pod to come up (without using `kubectl wait`) and then we do the `kubectl wait`.

I would suggest that I look into that in my folllow-up polishing PR, ok?

dev/env/scripts/bootstrap.sh
```diff
@@ -0,0 +1,121 @@
...
+inject_ips() {
+    local namespace="$1"
+    local service_account="$2"
+    local secret_name="$3"
+
+    log "Patching ServiceAccount ${namespace}/default to use Quay.io imagePullSecrets"
+    $KUBECTL -n "$namespace" patch sa "$service_account" -p "\"imagePullSecrets\": [{\"name\": \"${secret_name}\" }]"
+}
+
+# TODO: use a function.
```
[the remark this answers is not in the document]
mtesseract: done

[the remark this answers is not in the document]
mtesseract: done

dev/env/scripts/lib.sh
johannes94: Ok, got it now, thanks for explaining.

johannes94 (approved):
LGTM

PR #151, opened 2022-06-28
Add OCM refresh token authentication within fleetshard sync.

dhaus67:
## Description

This is the first PR in a series of PRs to add additional authentication types to the fleetshard synchronizer.
This will include the following authentication options:
- OCM authentication using the refresh token instead of short-lived access tokens.
- RH SSO authentication using the client_credentials grant with client ID / secret to retrieve access tokens.
- Static token authentication for development using a non-expiring static token valid for the fleet manager's API.

This PR will do a couple of things:
- Pre-factor the fleet manager client within fleetshard sync to add an abstraction for adding authentication information to the request.
- Implement the OCM authentication using the refresh token instead of the expiring access token.
- Add a new config option for the fleetshard sync called `AUTH_TYPE` which can be used to specify the different authentication types. Right now defaults to `OCM`.

There's a change in behavior:
Previously, the `OCM_TOKEN` variable was always using the value of `ocm token`. Now, the value should be yielded from `ocm token --refresh`.

## Test manual

```
# Start fleet manager locally with the dev config:
# Start fleetshard with the following:
OCM_TOKEN=$(ocm token --refresh) CLUSTER_ID=1234567890abcdef1234567890abcdef ./fleetshard-sync
# Observe within the logs that requests to central are done correctly without any errors
```


dhaus67:
* main
  * **PR #151** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/151" target="_blank">Graphite</a>  👈
    * **PR #152** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/152" target="_blank">Graphite</a> 
      * **PR #153** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/153" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/151?utm_source=stack-comment).

fleetshard/pkg/fleetmanager/static-token
```diff
@@ -0,0 +1 @@
...Nsb3VkLXNlcnZpY2VzIiwibm9uY2UiOiJlYjE0ODVkNi04MzUxLTQ0OTQtOTc4ZC04YjYyYmViZTEyNmIiLCJzZXNzaW9uX3N0YXRlIjoiMjA4NzNkMTItYWFlOC00ZDNiLTljMTQtNDRhNWEyNTNjMzY3IiwiYWNyIjoiMCIsInJlYWxtX2FjY2VzcyI6eyJyb2xlcyI6WyJhdXRoZW50aWNhdGVkIiwiZmxlZXQtbWFuYWdlci1hZG1pbi1mdWxsIl19LCJzY29wZSI6Im9wZW5pZCBpYW0uY2xpZW50cy5zZXJ2aWNlX2FjY291bnRzIG9mZmxpbmVfYWNjZXNzIiwic2lkIjoiMjA4NzNkMTItYWFlOC00ZDNiLTljMTQtNDRhNWEyNTNjMzY3IiwiYWNjb3VudF9udW1iZXIiOiIxMjM0NTY3IiwiaXNfaW50ZXJuYWwiOnRydWUsImlzX2FjdGl2ZSI6dHJ1ZSwibGFzdF9uYW1lIjoiVGVzdCIsInByZWZlcnJlZF91c2VybmFtZSI6InJoYWNzLW1zLXRlc3RAcmVkaGF0LmNvbSIsInR5cGUiOiJVc2VyIiwibG9jYWxlIjoiZW5fdXMiLCJpc19vcmdfYWRtaW4iOnRydWUsImFjY291bnRfaWQiOiIxMjM0NTY3OCIsInVzZXJfaWQiOiIxMjM0NTY3OCIsIm9yZ19pZCI6IjEyMzQ1Njc4IiwiZmlyc3RfbmFtZSI6IlRlc3QiLCJlbWFpbCI6InJoYWNzLW1zLXRlc3RAcmVkaGF0LmNvbSIsInVzZXJuYW1lIjoicmhhY3MtbXMtdGVzdEByZWRoYXQuY29tIn0.OlX7HrtPhlm0AZF_qj7IGemAUVbJX4hv8voNGZDIwK0vs2CoMzKiGEPqy6ds3cujS4zw8-GN35sKE2FzYuqy6fJpzG_v65P7l1b2jxri3a01CXbwUQS3c-_B5qyUMgfsA7Jcr57kwyc2JOHeITUibJ3mNa9J8UEVnlD5qLcoLC8gIH465qgaIzo4r_jry47lzSAU8ZI_AcQCIX8qkf4ZuHSR6gXYd-cHGMxfsHCdg3JGU4HtKVJjJq8lHdsI6KQuyC8AuVew02awoovJ5jHl2IvoDZw3vrsLBa4YwTrxI5o1G7WmaDAxV1rw7ZzGYA91XCJm6VZ2h36gRqvNPRbGuQ
```
kovayur, line 1: Is it ok to commit this token?

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -0,0 +1,124 @@
...
+type noAuth struct{}
+
+func (n noAuth) AddAuth(_ *http.Request) error {
+	return nil
+}
+
+// setBearer is a helper to set a bearer token as authorization header on the http.Request.
+func setBearer(req *http.Request, token string) {
+	// Do not attempt to modify any existing authorization headers.
+	if req.Header.Get("Authorization") == "" {
```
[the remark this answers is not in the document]
dhaus67: So, right now theoretically only the caller could do that (for whatever reason). Added it to be more defensive about it, but I also don't mind simply removing it to avoid potential confusion.

fleetshard/pkg/fleetmanager/static-token
dhaus67: Removed it.

mtodor (approved):
Fantastic work!

fleetshard/pkg/fleetmanager/client.go
```diff
@@ -24,21 +23,15 @@ const (
 // Client represents the REST client for connecting to fleet-manager
 type Client struct {
 	client                http.Client
+	auth                  Auth
 	ocmToken              string
```
ivan-degtiarenko, line 27: Why do we need `ocmToken` field now?

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -0,0 +1,71 @@
...
+
+// NewAuth will return Auth that can be used to add authentication of a specific AuthType to be added to HTTP requests.
+func NewAuth(t AuthType) (Auth, error) {
+	switch t {
+	default:
+		return newOcmAuth()
+	}
+}
+
+type noAuth struct{}
```
ivan-degtiarenko, line 55: would it make sense to move it to the test section since that's the only place it's used?

fleetshard/pkg/fleetmanager/auth_ocm.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, errors.New("empty ocm token")
+	}
+
+	l, err := sdk.NewGlogLoggerBuilder().Build()
+	if err != nil {
+		return nil, err
+	}
+
+	builder := sdk.NewConnectionBuilder().
+		Client("cloud-services", "").
```
ivan-degtiarenko, line 31: maybe extract it to `const`?

fleetshard/pkg/fleetmanager/client.go
dhaus67: Removed it.

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -0,0 +1,65 @@
+package fleetmanager
```
SimonBaeumer, line 1: Could this file be seen as a registry or factory of auth providers?
(If you agree, it is okay for me to rename in a following PR to avoid rebases)

SimonBaeumer (changes requested):

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -0,0 +1,65 @@
...
+	"strings"
+)
+
+// AuthType represents the supported authentication types for the client.
+type AuthType int
+
+const (
+	// OCMTokenAuthType supports authentication via the refresh_token grant using an offline token provided by
+	// console.redhat.com. The access token will be refreshed 1 minute before expiring via the refresh_token grant.
+	OCMTokenAuthType AuthType = iota
```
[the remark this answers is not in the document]
dhaus67: The constants need to be definied within a common place due to the `iota` usage, otherwise we would have to manually track and set the integers accordingly to keep the enumeration correct. Seems non-idiomatic to define them separately and manually set the integer instead of using `iota` at a common place.

fleetshard/pkg/fleetmanager/auth.go
dhaus67: Theoretically yes, although I've decided against defining the whole abstraction of factory (although it very well could be done). What's your suggestion, renaming it to `auth_factory`, even though there's no such concept of a factory?

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func (a AuthType) String() string {
+	return [...]string{
```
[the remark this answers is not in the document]
dhaus67: The `[...]string` is called an `Array literal`, where the length of the array will be equal to the number of elements given (in our case `OCM`).  We are not initializing a string but rather a string array with the length of the given elements.
Afterwards, its just a simple access to the array, not sure what's so special about that 😅

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -0,0 +1,65 @@
...
+		"OCM",
+	}[a]
+}
+
+// AuthTypeFromString will return the AuthType based on the string representation given. If no matching AuthType is
+// found, the issue will be logged and StaticTokenAuthType will be used as default.
```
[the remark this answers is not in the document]
dhaus67: Think there's two things: by default should be `OCMTokenAuthType`, but we can also change it to return an error and only allow valid authentication types, I don't have a strong opinion on it.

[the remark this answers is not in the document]
SimonBaeumer: Not defaulting to OCM is more clean. The API has unexpected behaviour because I provide an AuthType, but instead of getting an error if I provide an invalid just one random will be returned.

fleetshard/pkg/fleetmanager/auth.go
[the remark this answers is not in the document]
SimonBaeumer: The challenging part here is to understand that `AuthType` is a `iota` representing the index of the string array and the auth types string representation.
In this case the string array's index is depended on the value of the `AuthType`.
I would expect `AuthType` a struct and not an int, like discussed with `GetName()` func.

fleetshard/pkg/fleetmanager/auth.go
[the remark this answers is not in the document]
SimonBaeumer: This breaks the abstraction, no?
On one side every AuthProvider has its own file, on the other name and ID are defined in another file. 
The AuthProvider can know its name themselves, thus leaking less implementation details to the registry which only knows the interface types.

SimonBaeumer (approved):
/approve

dev/config/dataplane-cluster-configuration-minikube.yaml
```diff
@@ -18,7 +18,7 @@
 #    cluster_dns: apps.example.com #Valid cluster DNS. This will be used to build central host url and to communicate with standalone clusters. Required when "provider_type" is "standalone" 
 #    supported_instance_type: "eval" # could be "eval", "standard" or both i.e "standard,eval" or "eval,standard". Defaults to "standard,eval" if not set 
 clusters:
- - name: minikube  # Uncomment if using minikube
+ - name: rancher-desktop  # Uncomment if using minikube
```
ebensh, line 21: Please don't change this file, it's intended specifically for minikube. If you are using rancher-desktop either modify the copy or create a new one for your environment

PR #152, opened 2022-06-28
Add RH SSO authentication within fleetshard sync.

dhaus67:
## Description

This is the second PR in a series of PRs to add additional authentication types to the fleetshard synchronizer.
This will include the following authentication options:
- OCM authentication using the refresh token instead of short-lived access tokens.
- RH SSO authentication using the client_credentials grant with client ID / secret to retrieve access tokens.
- Static token authentication for development using a non-expiring static token valid for the fleet manager's API.

This PR will do a couple of things:
- Conditionally adding the [token-refresher](https://gitlab.cee.redhat.com/mk-ci-cd/mk-token-refresher) (which is also used for the observatorium access token) to the Helm deployment. It will write a valid access token to a file within a shared volume of fleetshard sync and the token refresher, and the token refresher will ensure the token is not expired.
- Implement the authentication using RH SSO by reading the token written to the file.
- Add a new option to the `AUTH_TYPE` called `RHSSO`, which will enabled RH SSO authentication and the deployment of the token-refresher within the Helm templated.

## Test manual

```
# Start fleet manager locally with the dev config:
# Deploy the fleetshard sync with helm
helm install -f values.yaml --set fleetshardSync.authType=RHSSO --set fleetshardSync.redhatSSO.clientId=<client-id> --set fleetshardSync.redhatSSO.clientSecret=<client-secret> --set fleetshardSync.clusterId=1234567890abcdef1234567890abcdef fleetshard dp-terraform/helm/rhacs-terraform
# Observe within the logs that requests to central are done correctly without any errors
```

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
@@ -20,8 +20,8 @@ spec:
       serviceAccountName: fleetshard-sync
       containers:
       - name: fleetshard-sync
-        image: fleetsync:latest
-        imagePullPolicy: Never
+        image: quay.io/app-sre/acs-fleet-manager:main
+        imagePullPolicy: Always
         command: 
         - /usr/local/bin/fleetshard-sync
         env:
@@ -35,11 +35,11 @@ spec:
           value: {{ .Values.fleetshardSync.authType }}
       {{- if and .Values.fleetshardSync.redHatSSO.clientId .Values.fleetshardSync.redHatSSO.clientSecret (eq .Values.fleetshardSync.authType "RHSSO") }}
         volumeMounts:
-          - mountPath: /run/secrets/rhsso
+          - mountPath: /run/secrets/rhsso-token
             name: rhsso-token
             readOnly: true
       - name: token-refresher
-        image: {{ .Values.fleetshardSync.tokenRefresher.image.name }}:{{ .Values.fleetshardSync.tokenRefresher.image.tag }}
+        image: {{ .Values.fleetshardSync.tokenRefresher.image }}
         imagePullPolicy: Always
         env:
           - name: CLIENT_ID
@@ -53,16 +53,16 @@ spec:
                 name: fleetshard-sync-rhsso-creds
                 key: clientSecret
           - name: ISSUER_URL
-            value: https://sso.redhat.com/auth/realms/redhat-external
+            value: {{ .Values.fleetshardSync.tokenRefresher.issuerUrl }}
         command:
           - /bin/token-refresher
           - --oidc.client-id=$(CLIENT_ID)
           - --oidc.client-secret=$(CLIENT_SECRET)
           - --oidc.issuer-url=$(ISSUER_URL)
           - --margin=1m
-          - --file=/rhsso/token
+          - --file=/rhsso-token/token
         volumeMounts:
-          - mountPath: /rhsso
+          - mountPath: /rhsso-token
             name: rhsso-token
       volumes:
         - name: rhsso-token
@@ -78,8 +78,8 @@ metadata:
   labels:
     app: fleetshard-sync
 type: Opaque
-stringData:
-  clientId: {{ .Values.fleetshardSync.redHatSSO.clientId }}
-  clientSecret: {{ .Values.fleetshardSync.redHatSSO.clientSecret }}
+data:
+  clientId: {{ .Values.fleetshardSync.redHatSSO.clientId | b64enc | quote }}
+  clientSecret: {{ .Values.fleetshardSync.redHatSSO.clientSecret | b64enc | quote }}
 ---
 {{- end }}
--- a/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
@@ -20,8 +20,8 @@ spec:
       serviceAccountName: fleetshard-sync
       containers:
       - name: fleetshard-sync
-        image: quay.io/app-sre/acs-fleet-manager:main
-        imagePullPolicy: Always
+        image: fleetsync:latest
+        imagePullPolicy: Never
         command: 
         - /usr/local/bin/fleetshard-sync
         env:
@@ -31,16 +31,43 @@ spec:
           value: {{ .Values.fleetshardSync.fleetManagerEndpoint }}
         - name: CLUSTER_ID
           value: {{ .Values.fleetshardSync.clusterId }}
-        {{- if and .Values.fleetshardSync.redHatSSO.clientId .Values.fleetshardSync.redHatSSO.clientSecret }}
+        - name: AUTH_TYPE
+          value: {{ .Values.fleetshardSync.authType }}
+      {{- if and .Values.fleetshardSync.redHatSSO.clientId .Values.fleetshardSync.redHatSSO.clientSecret (eq .Values.fleetshardSync.authType "RHSSO") }}
         volumeMounts:
-        - name: rhsso-creds
-          mountPath: /run/secrets/rhsso-creds
-          readOnly: true
+          - mountPath: /run/secrets/rhsso
+            name: rhsso-token
+            readOnly: true
+      - name: token-refresher
+        image: {{ .Values.fleetshardSync.tokenRefresher.image.name }}:{{ .Values.fleetshardSync.tokenRefresher.image.tag }}
+        imagePullPolicy: Always
+        env:
+          - name: CLIENT_ID
+            valueFrom:
+              secretKeyRef:
+                name: fleetshard-sync-rhsso-creds
+                key: clientId
+          - name: CLIENT_SECRET
+            valueFrom:
+              secretKeyRef:
+                name: fleetshard-sync-rhsso-creds
+                key: clientSecret
+          - name: ISSUER_URL
+            value: https://sso.redhat.com/auth/realms/redhat-external
+        command:
+          - /bin/token-refresher
+          - --oidc.client-id=$(CLIENT_ID)
+          - --oidc.client-secret=$(CLIENT_SECRET)
+          - --oidc.issuer-url=$(ISSUER_URL)
+          - --margin=1m
+          - --file=/rhsso/token
+        volumeMounts:
+          - mountPath: /rhsso
+            name: rhsso-token
       volumes:
-        - name: rhsso-creds
-          secret:
-            secretName: fleetshard-sync-rhsso-creds
-        {{- end }}
+        - name: rhsso-token
+          emptyDir: {}
+      {{- end }}
 ---
 {{- if and .Values.fleetshardSync.redHatSSO.clientId .Values.fleetshardSync.redHatSSO.clientSecret }}
 apiVersion: v1
@@ -52,8 +79,7 @@ metadata:
     app: fleetshard-sync
 type: Opaque
 stringData:
-  clientId: |
-    {{ .Values.fleetshardSync.redHatSSO.clientId | nindent 4 }}
-  clientSecret: |
-    {{ .Values.fleetshardSync.redHatSSO.clientSecret | nindent 4 }}
+  clientId: {{ .Values.fleetshardSync.redHatSSO.clientId }}
+  clientSecret: {{ .Values.fleetshardSync.redHatSSO.clientSecret }}
+---
 {{- end }}
--- a/fleetshard/README.md
+++ b/fleetshard/README.md
@@ -32,3 +32,12 @@ Execute all commands from git root directory.
     ```
     $ ./scripts/create-central.sh
     ```
+
+## Authentication types
+
+Fleetshard sync provides different authentication types that can be used when calling the fleet manager's API:
+- OCM refresh token
+  - This will use the OCM refresh token obtained via `ocm token --refresh` and will be refreshed before expiring.
+- RH SSO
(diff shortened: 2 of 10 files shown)


dhaus67:
* main
  * **PR #151** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/151" target="_blank">Graphite</a> 
    * **PR #152** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/152" target="_blank">Graphite</a>  👈
      * **PR #153** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/153" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/152?utm_source=stack-comment).

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -31,16 +31,43 @@ spec:
...
           value: {{ .Values.fleetshardSync.clusterId }}
-        {{- if and .Values.fleetshardSync.redHatSSO.clientId .Values.fleetshardSync.redHatSSO.clientSecret }}
+        - name: AUTH_TYPE
+          value: {{ .Values.fleetshardSync.authType }}
+      {{- if and .Values.fleetshardSync.redHatSSO.clientId .Values.fleetshardSync.redHatSSO.clientSecret (eq .Values.fleetshardSync.authType "RHSSO") }}
         volumeMounts:
-        - name: rhsso-creds
-          mountPath: /run/secrets/rhsso-creds
-          readOnly: true
+          - mountPath: /run/secrets/rhsso
```
mtodor, line 38: nitpick: I would suggest using the same name in `mountPath` as we have in `name` - i.e. `/run/secrets/rhsso-token`. I know that in some cases it will redundant - `/rhsso-token/token`. But it's still easier to track what volume is related to that.

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -31,16 +31,43 @@ spec:
...
+            name: rhsso-token
+            readOnly: true
+      - name: token-refresher
+        image: {{ .Values.fleetshardSync.tokenRefresher.image.name }}:{{ .Values.fleetshardSync.tokenRefresher.image.tag }}
```
mtodor, line 42: nitpick: I would suggest just using a string property for the image. i.e. `.Values.fleetshardSync.tokenRefresher.image`

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -31,16 +31,43 @@ spec:
...
+              secretKeyRef:
+                name: fleetshard-sync-rhsso-creds
+                key: clientId
+          - name: CLIENT_SECRET
+            valueFrom:
+              secretKeyRef:
+                name: fleetshard-sync-rhsso-creds
+                key: clientSecret
+          - name: ISSUER_URL
+            value: https://sso.redhat.com/auth/realms/redhat-external
```
mtodor, line 56: I can't remember this, but is the issuer the same for production and staging RedHat SSO? If they are different, then I would suggest moving this value to Helm values. If they are the same, we can keep them as it is.

EDIT: Looks like you have already planned that. 😄 As I see, there is already `.Values.fleetshardSync.tokenRefresher.issuerUrl`.

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -51,11 +63,37 @@ type Auth interface {
 // NewAuth will return Auth that can be used to add authentication of a specific AuthType to be added to HTTP requests.
 func NewAuth(t AuthType) (Auth, error) {
 	switch t {
+	case RHSSOAuthType:
+		return newRHSSOAuth(tokenPath)
```
mtodor, line 67: I would suggest using a similar interface for factory functions. Maybe we can call them without params.

For example:
```
newRHSSOAuth()
newOcmAuth()
```

And then inside the function, we are using constant or `os.Getenv("OCM_TOKEN")` in the case of `newOcmAuth`.

If we apply a suggestion to split implementations of Auth types into separate files from the first PR, then this change will also get nicely together.

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -12,25 +12,38 @@ import (
 	"time"
 )
 
+const (
+	tokenPath = "/run/secrets/rhsso/token"
```
mtodor, line 16: Should we consider providing this value via some command option or environment variable?

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -51,11 +63,37 @@ type Auth interface {
...
+}
+
+func (r *rhSSOAuth) AddAuth(req *http.Request) error {
+	// The file is populated by the token-refresher, which will ensure the token is not expired.
+	contents, err := os.ReadFile(r.tokenFilePath)
+	if err != nil {
+		return err
+	}
+
+	setBearer(req, string(contents))
```
mtodor, line 93: nit-nitpick: Let's call it a token.
```suggestion
	token, err := os.ReadFile(r.tokenFilePath)
	if err != nil {
		return err
	}

	setBearer(req, string(token))
```

fleetshard/pkg/fleetmanager/auth.go
dhaus67: Yeah, was thinking about this as well.
I've changed the config to be a singleton and re-use it within here, as we have the ability to provide sane defaults (like the default mount path we have in the deployment). I don't particularly like how it's currently handled with i.e. the `OCM_TOKEN` environment variable, would rather have it in a single place, so changed it to the singleton.

fleetshard/pkg/fleetmanager/auth.go
dhaus67: Fine with aligning the function signatures, albeit I won't go for the stretch and define an interface for auth type factories, i.e.
```go
type authFactory interface {
    Create() (Auth, error)
}
```
If we grow the numbers of authentication types, I think it's worthwhile to have it, but right now don't see the need to overcomplicate things.

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -20,8 +20,8 @@ spec:
       serviceAccountName: fleetshard-sync
       containers:
       - name: fleetshard-sync
-        image: quay.io/app-sre/acs-fleet-manager:main
-        imagePullPolicy: Always
+        image: fleetsync:latest
```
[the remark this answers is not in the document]
dhaus67: Yeah, that wasn't planned on being pushed, my bad.

I'd vote to do the changes in a separate PR regarding defining this in Helm values, since it's impact will be a bit more broad.

dhaus67:
@mtodor I've thought about using the RH SSO client, but ultimately decided against it due to the additionally required configuration options (i.e. realm, token endpoint URI). Instead, I followed the pattern that's already existing for the observatorium tokens. I know it'll increase complexity for the deployment / helm template, but at the very least keeps the config options to a minimum within the fleetshard sync (functionality wise, there's no difference between the token-refresher and the `GetToken()`, both will yield the required access token and handle creating a new one before expiration.

fleetshard/pkg/fleetmanager/auth_rhsso.go
```diff
@@ -0,0 +1,36 @@
...
+		return nil, err
+	}
+	return &rhSSOAuth{
+		tokenFilePath: tokenFilePath,
+	}, nil
+}
+
+func (r *rhSSOAuth) AddAuth(req *http.Request) error {
+	// The file is populated by the token-refresher, which will ensure the token is not expired.
+	contents, err := os.ReadFile(r.tokenFilePath)
```
mtodor, line 29: nit-nitpick. I see that you have marked a suggestion to change this into `token` with 👍 . Probably because of refactoring, that change has been lost.

mtodor (approved):
It looks great. I have added just two small tweaks, nothing crucial.

fleetshard/config/config.go
```diff
@@ -1,34 +1,52 @@
...
+	cfgErr error
+)
+
 // Config contains this application's runtime configuration.
 type Config struct {
 	FleetManagerEndpoint string        `env:"FLEET_MANAGER_ENDPOINT" envDefault:"http://127.0.0.1:8000"`
 	ClusterID            string        `env:"CLUSTER_ID"`
 	RuntimePollPeriod    time.Duration `env:"RUNTIME_POLL_PERIOD" envDefault:"5s"`
 	AuthType             string        `env:"AUTH_TYPE" envDefault:"OCM"`
+	RHSSOTokenFilePath   string        `env:"RHSSO_TOKEN_FILE" envDefault:"/run/secrets/rhsso-token/token"`
```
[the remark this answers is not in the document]
dhaus67: It can be done, I'll add some more documentation within #153 , as I've added additional docs within there.

fleetshard/config/config.go
```diff
@@ -1,34 +1,52 @@
...
+	OCMRefreshToken      string        `env:"OCM_TOKEN"`
 }
 
-// Load retrieves the current runtime configuration from the environment and returns it.
-func Load() (*Config, error) {
+func loadConfig() {
 	c := Config{}
 	if err := env.Parse(&c); err != nil {
-		return nil, errors.Wrapf(err, "Unable to parse runtime configuration from environment")
+		cfgErr = errors.Wrapf(err, "Unable to parse runtime configuration from environment")
```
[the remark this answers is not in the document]
dhaus67: Sure thing, I'll use `errorhelpers.ErrorList` since we already have the dependency within the rpeo. I also decided to short-circuit on the parsing error, since it doesn't really make sense to go through the different environment variables.

fleetshard/README.md
```diff
@@ -32,3 +32,12 @@ Execute all commands from git root directory.
     ```
     $ ./scripts/create-central.sh
     ```
+
+## Authentication types
+
+Fleetshard sync provides different authentication types that can be used when calling the fleet manager's API:
+- OCM refresh token
```
SimonBaeumer, line 39: I think I missed some of the conversations around SSO.
Can you refresh my memory why we support two different auth types in fleet-shard synchronizer?
(I thought OCM is for inter-service auth and RHSSO for customer auth into Central)

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -12,21 +12,30 @@ import (
...
+	// sso.redhat.com. The access token will be refreshed 1 minute before expiring via the client_credentials grant.
+	// NOTE: It can only be used in conjunction with the token-refresher.
+	// See dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml for more information.
+	RHSSOAuthType
 )
 
 func (a AuthType) String() string {
 	return [...]string{
 		"OCM",
+		"RHSSO",
```
SimonBaeumer, line 29: Ahhhh now I understood the code 🎉 
https://github.com/stackrox/acs-fleet-manager/pull/151#discussion_r911025015

fleetshard/pkg/fleetmanager/auth.go
```diff
@@ -12,21 +12,30 @@ import (
...
 const (
+
 	// OCMTokenAuthType supports authentication via the refresh_token grant using an offline token provided by
 	// console.redhat.com. The access token will be refreshed 1 minute before expiring via the refresh_token grant.
 	OCMTokenAuthType AuthType = iota
+	// RHSSOAuthType supports authentication via the client_credentials grant using client_id / client_secret via
+	// sso.redhat.com. The access token will be refreshed 1 minute before expiring via the client_credentials grant.
+	// NOTE: It can only be used in conjunction with the token-refresher.
+	// See dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml for more information.
+	RHSSOAuthType
```
SimonBaeumer, line 23: Instead of having all auth types as a constant I would prefer using an `init()` func which automatically registers all auth types here by here name.
Each `AuthType` has an interface which fulfills the function `GetName() string` and returns just the static name.
The registry has a map indexed by AuthType's name `map[string]AuthType` and can access the auth type by the index.

SimonBaeumer (changes requested):

fleetshard/README.md
dhaus67: The reason for OCM auth type is to have the possibility to use personal tokens when using the fleetshard synchronizer locally instead of requiring everyone to use the service account's client_id / client_secret. It's merely there for enhancing the development experience.
In production, we shall always use RH SSO auth type, using the client_credentials grant.

I don't understand the comment about `inter-service auth and RHSSO for customer auth into central`.
When you obtain an OCM token, you obtain a token issued by `sso.redhat.com`. The difference is _the way of obtaining the token_. The token obtained via `ocm token` will use the `refresh_token` grant, and cannot be used by service accounts. To use service accounts and non-personalized tokens, you have to use a service account with the `client_credentials` grant.

We still are using RH SSO for log in to central as an auth provider, but this is something unrelated, as there an OAuth client is used, which is something completely different.

fleetshard/pkg/fleetmanager/auth.go
dhaus67: I guess we can go that route if you want to, I just wanted to keep it simple initially since we don't plan to add anything additional, but 🤷 .

I'll do it in a follow-up PR to change all three implementations at once, to not make this one bigger.

fleetshard/README.md
SimonBaeumer: Thanks for clarifying @dhaus67.
Do you have a good doc, article or video at hand to educate myself more on these differences?

PR #154, opened 2022-06-28
Pass auth config to fleetshard

ivan-degtiarenko:
## Description
This PR passes auth provider config for `sso.redhat.com` from `acs-fleet-manager` to `fleetshard-sync`.

`sub` needed to be extracted from the token passed to `acs-fleet-manager`

This should close the work on [ROX-11063](https://issues.redhat.com/browse/ROX-11063) as well as help make progress on [ROX-11428](https://issues.redhat.com/browse/ROX-11428)

## Test manual

Manual test: 
1. Added logging to `fleetshard-sync`
2. Created central and looked into the central request sent to fleetshard-sync

### DIFF
--- a/internal/dinosaur/pkg/api/dbapi/central_request_types.go
+++ b/internal/dinosaur/pkg/api/dbapi/central_request_types.go
@@ -19,7 +19,7 @@ type CentralRequest struct {
 	SubscriptionId string `json:"subscription_id"`
 	Owner          string `json:"owner" gorm:"index"` // TODO: ocm owner?
 	OwnerAccountId string `json:"owner_account_id"`
-	OwnerUserid    string `json:"owner_userid"`
+	OwnerUserId    string `json:"owner_user_id"`
 	// The DNS host (domain) of the Central service
 	Host           string `json:"host"`
 	OrganisationId string `json:"organisation_id" gorm:"index"`
--- a/internal/dinosaur/pkg/api/dbapi/central_request_types.go
+++ b/internal/dinosaur/pkg/api/dbapi/central_request_types.go
@@ -19,7 +19,7 @@ type CentralRequest struct {
 	SubscriptionId string `json:"subscription_id"`
 	Owner          string `json:"owner" gorm:"index"` // TODO: ocm owner?
 	OwnerAccountId string `json:"owner_account_id"`
-	OwnerUserid    string `json:"userid"`
+	OwnerUserid    string `json:"owner_userid"`
 	// The DNS host (domain) of the Central service
 	Host           string `json:"host"`
 	OrganisationId string `json:"organisation_id" gorm:"index"`
--- a/internal/dinosaur/pkg/api/dbapi/central_request_types.go
+++ b/internal/dinosaur/pkg/api/dbapi/central_request_types.go
@@ -19,6 +19,7 @@ type CentralRequest struct {
 	SubscriptionId string `json:"subscription_id"`
 	Owner          string `json:"owner" gorm:"index"` // TODO: ocm owner?
 	OwnerAccountId string `json:"owner_account_id"`
+	OwnerUserid    string `json:"userid"`
 	// The DNS host (domain) of the Central service
 	Host           string `json:"host"`
 	OrganisationId string `json:"organisation_id" gorm:"index"`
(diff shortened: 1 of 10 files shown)


internal/dinosaur/pkg/api/dbapi/central_request_types.go
```diff
@@ -19,6 +19,7 @@ type CentralRequest struct {
 	SubscriptionId string `json:"subscription_id"`
 	Owner          string `json:"owner" gorm:"index"` // TODO: ocm owner?
 	OwnerAccountId string `json:"owner_account_id"`
+	OwnerUserid    string `json:"owner_userid"`
```
dhaus67, line 22: Would propose to use `owner_user_id` to keep the same naming scheme. Same applies to the migration.

internal/dinosaur/pkg/handlers/validation.go
```diff
@@ -103,6 +103,7 @@ func ValidateDinosaurClaims(ctx context.Context, dinosaurRequestPayload *public.
 		dinosaurRequest.Owner, _ = claims.GetUsername()
 		dinosaurRequest.OrganisationId, _ = claims.GetOrgId()
 		dinosaurRequest.OwnerAccountId, _ = claims.GetAccountId()
+		dinosaurRequest.OwnerUserid, _ = claims.GetSub()
```
dhaus67, line 106: Got the naming a bit confused, with the whole `sub == owner_user_id`. Would it make sense to just use `claims.GetUserId()` instead, to make it more clear for callers?

pkg/auth/context_config.go
```diff
@@ -13,6 +13,7 @@ var (
 	// sso.redhat.com token claim keys
 	alternateTenantUsernameClaim string = "preferred_username"
 	tenantUserIdClaim            string = "account_id"
+	subClaim                     string = "sub"
```
dhaus67, line 16: nit, to keep it in sync with the current naming of claims:
```suggestion
	tenantSubClaim                     string = "sub"
```

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -814,6 +814,13 @@ func BuildManagedDinosaurCR(dinosaurRequest *dbapi.CentralRequest, dinosaurConfi
 			},
 		},
 		Spec: manageddinosaur.ManagedDinosaurSpec{
+			Auth: manageddinosaur.AuthSpec{
+				ClientSecret: dinosaurConfig.RhSsoClientSecret,
+				// TODO: make part of dinosaurConfig
```
dhaus67, line 819: Do you have a ticket for that, or will this be done within this PR?

internal/dinosaur/pkg/migrations/20220114114500_add_centrals.go
```diff
@@ -26,6 +26,7 @@ func addCentralRequest() *gormigrate.Migration {
 		SubscriptionId                string     `json:"subscription_id"`
 		Owner                         string     `json:"owner" gorm:"index"`
 		OwnerAccountId                string     `json:"owner_account_id"`
+		OwnerUserid                   string     `json:"owner_userid"`
```
johannes94, line 29: We can not change existing migration this easy anymore, because a migration file is applied to DB only once and we have a long running instance of fleet manager within staging AppSRE environment. It uses a Database provisioned by AppSRE, which we can't access from our Dev Machines to rollback migrations manually. We have to options:

1. Use proper migration i.e add a new migration that changes the DB model
2. Rollback migration in AppSRE DB by temporary changing deployments init container to run `fleet-manager migrate rollback-all` like we did in #114 to prepare for #82

johannes94 (changes requested):

internal/dinosaur/pkg/api/dbapi/central_request_types.go
ivan-degtiarenko: Migration automatically generates `owner_userid` name for the column. We could probably use `RenameColumn` to change to `owner_user_id`, but is it really worth it? 

Alternatively, we could rename the field to "OwnerUserId", then it will be `owner_user_id` column name.

internal/dinosaur/pkg/migrations/20220114114500_add_centrals.go
ivan-degtiarenko: Let's do it a proper way and create a new migration 👍

internal/dinosaur/pkg/api/dbapi/central_request_types.go
dhaus67: Well, then let's rename it to `UwnerUserId`, this also fits better to `OwnerAccountId, SubscriptionId etc...`.

internal/dinosaur/pkg/services/dinosaur.go
ivan-degtiarenko: Created a ticket

dhaus67 (approved):
Looks good to me 🚀 
Probably would be best to await a review from Draconians before merging tho.

johannes94 (approved):
LGTM

PR #155, opened 2022-06-29
Allow to configure arbitrary issuer/jwks URIs for token verification

dhaus67:
## Description

Currently, the UI development is blocked when working with the staging version of [console.stage.redhat.com](https://console.stage.redhat.com/). This is due to the fact that the staging environment is also using 
[sso.stage.redhat.com](https://sso.stage.redhat.com), which is not a valid issuer with the current default settings of fleet manager. 

Since our service accounts and personal tokens (obtained via `ocm token`) are within the production instance of RH SSO ([sso.redhat.com](https://sso.redhat.com)) and there's currently no way to add arbiratry additional SSO URIs,
we have to create a new configuration setting to set those.

I've chosen to use the file mechanism for those, allowing us to have more flexibility in the future w.r.t the amount of endpoints we want to add. The configuration option is similar to the [access control list 
config](https://github.com/stackrox/acs-fleet-manager/blob/main/pkg/acl/access_control_list.go#L31-L34), requiring explicit opt-in. For the development environment, additional URIs are enabled and `sso.stage.redhat.com` is already 
configured.

I've currently decided against explicitly configuring the realm, but we can easily do this by requiring the endpoint within the files to be of the format `https://<uri>/auth/<realm>` instead. However I'm not sure we require this 
use-case for now.

## Test manual
```
# Create an OCM token on [console.stage.redhat.com/openshift/token](https://console.stage.redhat.com/openshift/token/) using the credentials within Bitwarden of `ACS RH SSO Stage Corporate Account Test Login Info`.
OCM_TOKEN=$(ocm token)
# Run the fleet manager locally without any additional configuration besides the usual cluster config:
# Call the fleet manager's API with the OCM token issued by `https://sso.stage.redhat.com/auth/realms/redhat-external`:
curl --header 'Authorization: Bearer $OCM_TOKEN’ 'http://localhost:8000/api/rhacs/v1/centrals' 
# You should observe a successful status response.
```


dhaus67:
* main
  * **PR #155** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/155" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/155?utm_source=stack-comment).

pkg/client/keycloak/config.go
```diff
@@ -133,5 +144,43 @@ func (kc *KeycloakConfig) ReadFiles() error {
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+const (
+	issuerPath = "/auth/realms/redhat-external"
```
mtodor, line 165: It would be good if we could make this also part of the configuration. It's RedHat external SSO specific. Maybe we can generalize a bit.

Maybe instead of endpoints `config/additional-sso-endpoints.yaml` - we could define `issuers`. In this case: `https://sso.stage.redhat.com/auth/realms/redhat-external` -> from there we can fetch `https://sso.stage.redhat.com/auth/realms/redhat-external/.well-known/openid-configuration` and get `jwks_uri` from JSON. It should be also a safer option if something changes within the SSO configuration.

pkg/client/keycloak/config.go
```diff
@@ -133,5 +144,43 @@ func (kc *KeycloakConfig) ReadFiles() error {
 	kc.OSDClusterIDPRealm.setDefaultURIs(kc.BaseURL)
 	kc.RedhatSSORealm.setDefaultURIs(kc.SsoBaseUrl)
 
+	// Read the additional endpoints file. This will add additional SSO endpoints which shall be used as valid issuers
+	// for tokens, i.e. sso.stage.redhat.com.
+	if kc.AdditionalSSOEndpoints.EnableAdditionalSSOEndpoints {
```
mtodor, line 149: Should we avoid an additional file and simply accept additional issuers via a command option. (comma separated) i.e.
```
--additional-oidc-issuers https://sso.stage.redhat.com/auth/realms/redhat-external,https://sso.redhat.com/auth/realms/redhat-external
```
It would be fewer things to manage and a bit simpler solution.

sachaudh:
@dhaus67 I tried running the commands you displayed in the Description section but was getting some errors:

```
➜  acs-fleet-manager git:(dh/Allow_to_configure_arbitrary_issuer/jwks_URIs_for_token_verification) ✗ ./fleet-manager serve --dataplane-cluster-config-file dev/config/dataplane-cluster-configuration-minikube.yaml
I0629 14:24:25.672600   50867 environment.go:59] Environment variable "OCM_ENV" not specified, using default "development"
I0629 14:24:25.674092   50867 environment.go:108] Initializing development environment
I0629 14:24:25.674521   50867 config.go:128] Specified OSD SSO TLS certificate file "secrets/keycloak-service.crt" does not exist. Proceeding as if OSD SSO TLS certificate was not provided
I0629 14:24:25.674533   50867 config.go:138] Specified service account limits skip org IDs  file "config/service-account-limits-check-skip-org-id-list.yaml" does not exist. Proceeding as if no service account org ID skip list was provided
I0629 14:24:25.675137   50867 dinosaur.go:68] Central Red Hat OIDC client secret from secret file "" is missing.
E0629 14:24:25.691873   50867 environment.go:119] unable to read configuration files: standalone cluster with ID: 1234567890abcdef1234567890abcdef, and Name minikube not in kubeconfig context
F0629 14:24:25.691921   50867 cmd.go:21] Unable to initialize environment: unable to read configuration files: standalone cluster with ID: 1234567890abcdef1234567890abcdef, and Name minikube not in kubeconfig context
```

sachaudh:
@dhaus67 I tried this out and things seem to be working for me. The UI can make authenticated API calls now

Screen Shot 2022-06-29 at 2 49 54 PM
Screen Shot 2022-06-29 at 2 50 01 PM

I tried doing the API call for creating a Central request but was seeing an error

Screen Shot 2022-06-29 at 2 54 20 PM
Screen Shot 2022-06-29 at 2 54 25 PM
Screen Shot 2022-06-29 at 2 54 29 PM

I think I've seen something similar before so I thought changing `line 11` in `dataplane-cluster-configuration-dockerdesktop.yaml` to `region: us-east-1` would fix it. I tried again, but seems like it didn't really fix the issue. Now I see the following:

Screen Shot 2022-06-29 at 2 58 35 PM

Any idea what to do to allow the POST to go through successfully?

sachaudh (commented):
In terms of what I needed for the UI, this PR looks good to me. I was able to make authenticated API calls using sso.stage.redhat.com. I'll leave code review to the rest of the reviewers

pkg/client/keycloak/config.go
dhaus67: Think that's easier for configuration, I've added your solution to the config.

pkg/client/keycloak/config.go
dhaus67: I thought about this as well. It would work well for local deployments, passing the flag there, but when deploying it on staging / prod we would have to ensure the flags are set there. For this, we currently use the [SimpleEnvLoader](https://github.com/stackrox/acs-fleet-manager/blob/a578cf070ab4903036b633954b3b4b04d6f409b7/pkg/environments/env_loader.go#L8). If we were to support the args as multiple strings, we would have to ensure that the env loader can handle multiple strings as values, which it currently can't. I don't want to make too many changes related to this, so I opted for the file approach as it also is currently used in other places.

We could also agree with the CLI flags and say that staging won't have the possibility for arbitrary SSO endpoints and you have to run locally to make use of it, but that's also _meh_.

dhaus67:
@sachaudh nice to hear that authN now works correctly!

Regarding your issue with the central request: I haven't done this yet (provisioning request) and can't really help you with the issue 😞
Best would be to loop in @ebensh / @juanrh , they are more knowledgable when it comes to that part of the API and potentially the specific error message.

sachaudh:
> @sachaudh nice to hear that authN now works correctly!
> 
> Regarding your issue with the central request: I haven't done this yet (provisioning request) and can't really help you with the issue 😞 Best would be to loop in @ebensh / @juanrh , they are more knowledgable when it comes to that part of the API and potentially the specific error message.

Sounds good. Yea it's not really related to this PR so I'll ask them separately 👍🏼

pkg/client/keycloak/config.go
```diff
@@ -133,5 +149,82 @@ func (kc *KeycloakConfig) ReadFiles() error {
...
+			return errors.Errorf("no jwks URI found within open-id configuration for %q", issuerURI)
+		}
+		jwksURIs = append(jwksURIs, cfg.JwksURI)
+	}
+	a.JWKSURIs = jwksURIs
+	return nil
+}
+
+func getOpenIDConfiguration(c http.Client, baseURL string) (*openIdConfiguration, error) {
+	url := baseURL + openidConfigurationPath
```
mtodor, line 200: nitpick: To be on the safe side we should trim right `/`. Sometimes copy paste slips and we copy URLs with slash at the end.
```suggestion
	url := strings.TrimRight(baseURL, "/") + openidConfigurationPath
```

mtodor (approved):
It looks good! I have added a few nitpicks, nothing special. 🍻

dhaus67:
@kurlov I also had to change some things here regarding the `service-template.yaml`, since we are adding a new file to allow for additional issuers for tokens (specifically, sso.stage.redhat.com). I've added it to the staging environment + in a config map, would be great if you could take a look as well.

internal/dinosaur/pkg/handlers/authentication.go
```diff
@@ -24,7 +24,14 @@ func NewAuthenticationBuilder(ServerConfig *server.ServerConfig, KeycloakConfig
 		return nil, pkgErrors.Wrap(err, "unable to create authentication logger")
 	}
 
-	return authentication.NewHandler().
+	authenticationBuilder := authentication.NewHandler()
+
+	// Add additional JWKS endpoints to the builder if there are any.
+	for _, jwksEndpointURI := range KeycloakConfig.AdditionalSSOIssuers.JWKSURIs {
```
mtodor, line 30: flavor-nitpick: I have noticed that we are using issuer and endpoint interchangeably in comments, error logs, command descriptions, etc. Maybe we should align them in some way.

I understand that wording like endpoint is a valuable description for users of commands. Still, when I read error and command descriptions, they are a bit off. I'm really not sure what is the best solution here. 😕

dhaus67: Yeah, I'll clean it up to be more precise and use `issuer` within comments and for flags possibly `issuer URI` instead of `endpoint`.

templates/service-template.yml
```diff
@@ -410,6 +410,15 @@ objects:
     data:
       deny-list-configuration.yaml: |-
         ${DENIED_USERS}
+  - kind: ConfigMap
+    apiVersion: v1
+    metadata:
+      name: fleet-manager-additional-sso-issuers-config
+      annotations:
+        qontract.recycle: "true"
```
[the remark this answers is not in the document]
dhaus67: Seems to be coming from [app-sre/qontract-reconciler](https://github.com/app-sre/qontract-reconcile)

internal/dinosaur/pkg/environments/development.go
```diff
@@ -32,5 +32,7 @@ func NewDevelopmentEnvLoader() environments.EnvLoader {
 		"observability-red-hat-sso-token-refresher-url":   "http://localhost:8085",
 		"observability-red-hat-sso-observatorium-gateway": "https://observatorium-mst.api.stage.openshift.com",
 		"observability-red-hat-sso-tenant":                "manageddinosaur",
+		"enable-additional-sso-endpoints":                 "true",
```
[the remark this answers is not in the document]
dhaus67: Thanks, good catch!

PR #158, opened 2022-06-29
ROX-9843: Enable monitoring endpoint for ACS central

mtodor:
## Description

The observability stack is configured to use metrics provided by central. These metrics include billing metrics which are essential for managed services.

Recently we have added an option for the ACS operator to enable monitoring endpoint for ACS central.

With this PR, we are using a new option and enabling monitoring during provisioning of ACS operator by Fleetshard Synchronizer.

**Changes**

- added flag to enable exposing of monitoring port for central
- updated stackrox "lib" to have a new Central struct
- fixed unit tests because the gorm library is updated too, and there have been some changes in the query builder

One clarification. If we have an older version of the ACS operator on the cluster (version without new flags). If a Central resource is created with a new version of fleetshard synchronizer where the flag is set. The ACS operator will simply ignore that flag. It will not exist in YAML of the central custom resource.

- ~[ ] Documentation added if necessary~

## Test manual

1. Follow the guide to create a staging OSD cluster (be sure to follow the steps where you install ACS operator from master branch)
2. Build and deploy fleetshard synchronizer from this branch
3. Create central via fleetshard manager
4. This central should have monitoring port enabled. To check that, you can run following command:
```
kubectl get deployment --namespace rhacs-cavei8287d5rsd2of9k0 central --output yaml | yq '.spec.template.spec.containers[0].ports'
```
and also check that network policy exists
```
kubectl get networkpolicies.networking.k8s.io --namespace rhacs-cavei8287d5rsd2of9k0 monitoring
```


stehessel (approved):
We will need to do the same thing for Scanner.

PR #163, opened 2022-07-01
Mount stage dataplane configuration file

kurlov:
## Description

Mount OSD dataplane cofig file for stage using on appSRE control plane

### DIFF
--- a/templates/service-template.yml
+++ b/templates/service-template.yml
@@ -896,7 +896,7 @@ objects:
             - name: fleet-manager-dataplane-cluster-scaling-config
               mountPath: /config/dataplane-cluster-configuration.yaml
               subPath: dataplane-cluster-configuration.yaml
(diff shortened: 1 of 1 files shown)


templates/service-template.yml
```diff
@@ -896,6 +896,9 @@ objects:
             - name: fleet-manager-dataplane-cluster-scaling-config
               mountPath: /config/dataplane-cluster-configuration.yaml
               subPath: dataplane-cluster-configuration.yaml
+            - name: feet-manager-dataplane-cluster-configuration-staging.yaml
```
mtesseract, line 899: feet? :-D

mtesseract (changes requested):

templates/service-template.yml
kurlov: 🦵 😅

PR #176, opened 2022-07-05
Move image to values and improve README

mtodor:
## Description

This PR provides two small tweaks:
1. Image from terraforming helm template is moved to `values.yaml` - this will help us define different images on staging/production/dev environments
2. Improved README to use helm upgrade command with `--create-namespace` - this helps when the OSD cluster is new and we do not have the required namespace created. This option will simply create it.

- ~[ ] Unit and integration tests added~ - this is not important for this PR

## Test manual

1. You can execute the following commands to test it:
```
# in root of the project
cd dp-terraform/helm/rhacs-terraform

helm template rhacs-terraform \
  --debug \
  --namespace rhacs \
  --values ./values.yaml \
  --set fleetshardSync.image="test-image" .
```
After that check output and search for Deployment with the name `fleetshard-sync`. Check that the specified `fleetshardSync.image` is used there for the `image` property of the first container.

2. It's possible to test by selecting not existing namespace. Also, documentation is available here: https://helm.sh/docs/helm/helm_upgrade/

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/README.md
+++ b/dp-terraform/helm/rhacs-terraform/README.md
@@ -31,6 +31,7 @@ helm template rhacs-terraform \
 ```bash
 helm upgrade --install rhacs-terraform \
   --namespace rhacs \
+  --create-namespace \
   --values ~/acs-terraform-values.yaml \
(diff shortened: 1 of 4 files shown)


stehessel (approved):
👍

mtesseract (approved):

PR #178, opened 2022-07-06
ROX-11428: Initialise auth provider from fleetshard-sync

ivan-degtiarenko:
## Description
This PR initialises sso.redhat.com auth provider in freshly deployed Centrals. 

### Algo
1. Wait until central is deployed
2. Obtain admin password
3. Obtain Central's address
4. POST auth provider
While central is not deployed, we don't treat it as "Ready". "Ready" central = deployed + auth provider initialised.

### Limitations
Current implementation supports only fleetshard-sync deployed in the same k8s cluster as Central. To avoid breaking `fleetshard-sync` development deployments, I added `CREATE_AUTH_PROVIDER` flag which set to false by default. However, in helm chart `CREATE_AUTH_PROVIDER` is set by default to true.

`createAuthProvider` flag in reconciler helps to prevent `AlreadyExists` errors, which can arise between auth provider POST and `fleet-manager` switching central's state to `Ready`.

## Test manual

1. Deployed fleet-manager  and operator locally
2. Added `fleetshard-sync` image to `minikube` cluster
3. Deployed `fleetshard-sync` using modified Helm chart(modified image tag)
4. POST Central to `fleet-manager`
5. After Central is initialised, port-forward -> see `Red Hat SSO(Stage)` auth provider


fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+		Enabled:    true,
+		Config: map[string]string{
+			// TODO: ROX-11619: make configurable
+			"issuer":        "https://sso.stage.redhat.com/auth/realms/redhat-external",
+			"client_id":     central.Spec.Auth.ClientId,
+			"client_secret": central.Spec.Auth.ClientSecret,
+			"mode":          "post",
+		},
+		// TODO: for testing purposes only; remove once host is correctly specified in fleet-manager
+		ExtraUiEndpoints: []string{"localhost:8443"},
```
SimonBaeumer, line 90: Is this because of the lack of correct exposure of Central?
Which implications does this have if we release it by accident?

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+		Namespace: central.Metadata.Namespace,
+	}
+	secret := &core.Secret{}
+	err := client.Get(ctx, secretRef, secret)
+	if err != nil {
+		return "", errors.Wrap(err, "obtaining admin password secret")
+	}
+	password := string(secret.Data[adminPasswordSecretKey])
+	if password == "" {
+		return "", errors.Errorf("No password present in %s secret. This should not be the case in Central instances installed via operator.", centralHtpasswdSecretName)
```
SimonBaeumer, line 123: You can disable htpasswd, the short version of the error message is okay too.

```suggestion
		return "", errors.Errorf("no password present in %s secret.", centralHtpasswdSecretName)
```

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -31,6 +31,8 @@ spec:
           value: {{ .Values.fleetshardSync.fleetManagerEndpoint }}
         - name: CLUSTER_ID
           value: {{ .Values.fleetshardSync.clusterId }}
+        - name: CREATE_AUTH_PROVIDER
```
SimonBaeumer, line 34: Does this configuration not break all current builds which were executed locally and change the development workflow by needing to set an additional environment variable?

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+
+// InitRHSSOAuthProvider initialises sso.redhat.com auth provider in a deployed Central instance.
+func InitRHSSOAuthProvider(ctx context.Context, central private.ManagedCentral, client ctrlClient.Client) error {
+	// Acquire central admin password.
+	pass, err := acquireAdminPassword(ctx, central, client)
+	if err != nil {
+		return err
+	}
+
+	// Acquire central address.
```
SimonBaeumer, line 42: Comment duplicate function name

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+
+func init() {
+	insecureTransport = http.DefaultTransport.(*http.Transport).Clone()
+	// TODO: once certificates will be added, we probably will be able to replace with secure transport
+	insecureTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+}
+
+// InitRHSSOAuthProvider initialises sso.redhat.com auth provider in a deployed Central instance.
+func InitRHSSOAuthProvider(ctx context.Context, central private.ManagedCentral, client ctrlClient.Client) error {
+	// Acquire central admin password.
```
SimonBaeumer, line 36: Comment duplicates function name

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+		Transport: insecureTransport,
+	}
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "sending new auth provider request to central")
+	}
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	} else {
+		return acsErrors.NewErrorFromHTTPStatusCode(resp.StatusCode, "Failed to create auth provider for central %s", central.Metadata.Name)
```
SimonBaeumer, line 71: Imho this code (line 48 - 71) can be extracted into its own package `central_client` to be re-usable. 
It is likely we need more API calls to Central in the future, having a package which allows us to quickly add an additional call with password auth is great.

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+	req = req.WithContext(ctx)
+
+	httpClient := http.Client{
+		Transport: insecureTransport,
+	}
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "sending new auth provider request to central")
+	}
+	if resp.StatusCode == http.StatusOK {
```
SimonBaeumer, line 68: Is it really an http 200? ie. http 201 is `Created` which would be more correct.
Imho a check for `2xx` would be better here. StackRox has a helper function which can do exactly that.

SimonBaeumer (changes requested):
Great work @ivan-degtiarenko!

fleetshard/pkg/centralreconciler/init_auth.go
ivan-degtiarenko: Correct, current `host` specified in spec is unreachable. 

The implication: if you have port-forward into Central/can access it on `localhost:8443`, you can log in if you have correct credentials.

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
ivan-degtiarenko: It's not required and set to `false` by default, leaving development workflow unchanged

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -132,10 +134,46 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
+		return nil, err
+	}
+	if !centralReady {
+		return installingStatus(), nil
+	}
+
+	// Skip auth provider initialisation if:
+	// 1. Auth provider is created by this specific reconciler
+	// 2. OR reconciler creator specified auth provider not to be created
+	// 3. OR Central request is in status "Ready" - meaning auth provider should've been initialised earlier
```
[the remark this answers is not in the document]
ivan-degtiarenko: For now it's not, auth provider is created once and is saved to database.

If we ever decide to implement client secret rotation, we can change auth provider via `fleetshard-sync`, but then we would probably need to implement comparison of existing auth provider with the one coming to `fleetshard-sync`.

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+	centralHtpasswdSecretName = "central-htpasswd"
+	adminPasswordSecretKey    = "password"
+	centralServiceName        = "central"
+)
+
+var (
+	insecureTransport *http.Transport
+)
+
+func init() {
```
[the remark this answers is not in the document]
SimonBaeumer: Yeah that makes sense. But why creating a package-wide variable if it can be in the scope of the function as it is only used once in the `http.Client` creation?
It is a variable initialization needing now a `var ()` block + an `init` function which could be just created before the `http.Client{}` creation in line 61.

fleetshard/pkg/centralreconciler/reconciler.go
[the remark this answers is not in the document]
SimonBaeumer: Should the auth provider be non-editable in Central? Or the auth provider is always overwritten an idempotent.

Not for this PR but still a thing we need to take care of. In case a customer accidentally (or intentionally) changes the auth provider it is hard to recover if the auth provider is never re-created/fixed by fleetshard.
Also we don't have access to a production cluster.

This can be very frustrating that a customer might lock themselves out of the system.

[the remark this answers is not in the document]
SimonBaeumer: I'll prefer the idempotent change, thus no changes to Central are necessary for now but still allows us to trigger updates which will fix the auth provider automatically.

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,126 @@
...
+func createAuthProviderRequest(central private.ManagedCentral) *storage.AuthProvider {
+	request := &storage.AuthProvider{
+		// TODO: ROX-11619: change depending on whether environment is stage or not
+		Name:       "Red Hat SSO(Stage)",
+		Type:       "oidc",
+		UiEndpoint: central.Spec.Endpoint.Host,
+		Enabled:    true,
+		Config: map[string]string{
+			// TODO: ROX-11619: make configurable
+			"issuer":        "https://sso.stage.redhat.com/auth/realms/redhat-external",
```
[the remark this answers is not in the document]
ivan-degtiarenko: I'm not sure I get you.

The only two possible values are `https://sso.stage.redhat.com/auth/realms/redhat-external` and `https://sso.redhat.com/auth/realms/redhat-external`. Depending on whether we are in stage vs prod, fleet-manager will send either one or another issuer value to fleetshard-sync. 
This value would be unmodifiable for deployed Central instances.

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,168 @@
...
+	err := client.Get(ctx,
+		ctrlClient.ObjectKey{Name: centralServiceName, Namespace: central.Metadata.Namespace},
+		service)
+	if err != nil {
+		return "", errors.Wrapf(err, "getting k8s service for central")
+	}
+	if len(service.Spec.Ports) == 0 {
+		return "", errors.Errorf("No ports present in %s service", centralServiceName)
+	}
+	address := fmt.Sprintf("https://%s.%s.svc.cluster.local:%d", centralServiceName, central.Metadata.Namespace, service.Spec.Ports[0].Port)
```
dhaus67, line 149: Hm, maybe we should use the name specifically here as well, as defined within [here](https://github.com/stackrox/stackrox/blob/master/image/templates/helm/stackrox-central/templates/_central_endpoints.tpl#L5)? In case of i.e. adding monitoring ports or other things to the service, we can't really hope that ordering stays the same.
To make things safer, we can check for name as well as targetPort.

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,168 @@
...
+		Namespace: central.Metadata.Namespace,
+	}
+	secret := &core.Secret{}
+	err := client.Get(ctx, secretRef, secret)
+	if err != nil {
+		return "", errors.Wrap(err, "getting admin password secret")
+	}
+	password := string(secret.Data[adminPasswordSecretKey])
+	if password == "" {
+		return "", errors.Errorf("no password present in %s secret.", centralHtpasswdSecretName)
```
dhaus67, line 165:
```suggestion
		return "", errors.Errorf("no password present in %s secret", centralHtpasswdSecretName)
```

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+func sendGroupRequest(ctx context.Context, central private.ManagedCentral, groupRequest *storage.Group, address string, pass string) error {
+	resp, err := centralClient.SendRequestToCentral(ctx, groupRequest, address+"/v1/groups", pass)
+	if err != nil {
+		return errors.Wrap(err, "sending new group to central")
+	}
+	if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		return acsErrors.NewErrorFromHTTPStatusCode(resp.StatusCode, "Failed to create group for central %s", central.Metadata.Name)
```
dhaus67, line 112:
```suggestion
		return acsErrors.NewErrorFromHTTPStatusCode(resp.StatusCode, "failed to create group for central %s", central.Metadata.Name)
```

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+
+	// Send auth provider POST request to Central.
+	authProviderRequest := createAuthProviderRequest(central)
+	resp, err := centralClient.SendRequestToCentral(ctx, authProviderRequest, address+"/v1/authProviders", pass)
+	if err != nil {
+		return errors.Wrap(err, "sending new auth provider to central")
+	} else if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		return acsErrors.NewErrorFromHTTPStatusCode(resp.StatusCode, "Failed to create auth provider for central %s", central.Metadata.Name)
```
dhaus67, line 79:
```suggestion
		return acsErrors.NewErrorFromHTTPStatusCode(resp.StatusCode, "failed to create auth provider for central %s", central.Metadata.Name)
```

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+	}
+)
+
+type AuthProviderResponse struct {
+	Id string `json:"id"`
+}
+
+// CreateRHSSOAuthProvider initialises sso.redhat.com auth provider in a deployed Central instance.
+func CreateRHSSOAuthProvider(ctx context.Context, central private.ManagedCentral, client ctrlClient.Client) error {
```
dhaus67, line 62: Does it need to be made public? AFAICT its only used within the reconciler, so I don't see why we have to make it public.

fleetshard/pkg/centralreconciler/reconciler.go
```diff
@@ -143,10 +145,46 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
+	if r.createAuthProvider && remoteCentral.RequestStatus != centralConstants.DinosaurRequestStatusReady.String() {
```
dhaus67, line 161: Just out of curiosity: we don't really plan to support anything like "changing owner of an existing instance", right? Cause this would imply that `3. OR Central request is in status "Ready"` to not be true anymore, as we would need to reconcile the auth provider and change the groups, specifically the one using the OwnerID, right?

fleetshard/pkg/centralreconciler/reconciler_test.go
```diff
@@ -40,12 +42,13 @@ func conditionForType(conditions []private.DataPlaneClusterUpdateStatusRequestCo
...
 	r := CentralReconciler{
-		status:    pointer.Int32(0),
-		client:    fakeClient,
-		central:   private.ManagedCentral{},
-		useRoutes: true,
+		status:             pointer.Int32(0),
+		client:             fakeClient,
+		central:            private.ManagedCentral{},
+		useRoutes:          true,
+		createAuthProvider: false,
```
dhaus67, line 51: nit: since `useRoutes` is skipped being explicitly set when false, maybe we should do the same with `createAuthProvider` or at the very least be consistent there.

fleetshard/pkg/centralreconciler/reconciler.go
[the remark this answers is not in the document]
SimonBaeumer: > We still need to figure out how secret rotation will work then 

Is my understanding correct that it the use-cases we have are:
 - Update secret from API to rotate secrets
 - Disable editing from end users

Do you think the secret rotation will get to a special case for managed services in the UI?

fleetshard/pkg/centralreconciler/init_auth.go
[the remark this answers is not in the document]
SimonBaeumer: Ah sorry, I confused some words.
All settings for the auth provider are passed from the fleet-manager to the fleetshard, why is the issuer URL not passed from fleet-manager?
In fleet-manager then make it configurable that a new image is not necessary on URL changes via an env var `RED_HAT_SSO_PROVIDER_ISSUER` or `RED_HAT_SSO_PROVIDER_ENDPOINT`.

fleetshard/pkg/central_client/client.go
```diff
@@ -0,0 +1,35 @@
...
+		return nil, errors.Wrap(err, "marshalling new request to central")
+	}
+	req, err := http.NewRequest(http.MethodPost, address, bytes.NewBuffer(jsonBytes))
+	if err != nil {
+		return nil, errors.Wrap(err, "creating HTTP request to central")
+	}
+	req.SetBasicAuth("admin", pass)
+	req = req.WithContext(ctx)
+
+	insecureTransport := http.DefaultTransport.(*http.Transport).Clone()
```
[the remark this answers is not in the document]
SimonBaeumer: I've suggested @ivan-degtiarenko to do it inline.
Why would you move such a simple operation to an `init` function introducing package-wide variables?

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,168 @@
...
+		// TODO: for testing purposes only; remove once host is correctly specified in fleet-manager
+		ExtraUiEndpoints: []string{"localhost:8443"},
+	}
+	return request
+}
+
+// TODO: ROX-11644: doesn't work when fleetshard-sync deployed outside of Central's cluster
+func getServiceAddress(ctx context.Context, central private.ManagedCentral, client ctrlClient.Client) (string, error) {
+	service := &core.Service{}
+	err := client.Get(ctx,
```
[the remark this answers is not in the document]
SimonBaeumer: The API call would work but the returned endpoint would not be accessible from outside the cluster I guess.

fleetshard/pkg/centralreconciler/reconciler.go
SimonBaeumer: No, Central instances can not be transitioned between different customers currently.

fleetshard/pkg/centralreconciler/init_auth.go
```diff
@@ -0,0 +1,168 @@
...
+		group := groupCreator(authProviderResp.Id, central.Spec.Auth)
+		err = sendGroupRequest(ctx, central, group, address, pass)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func sendGroupRequest(ctx context.Context, central private.ManagedCentral, groupRequest *storage.Group, address string, pass string) error {
```
SimonBaeumer, line 106: This `sendGroupRequest` function can be part of the centralclient.

fleetshard/pkg/central_client/client.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"github.com/pkg/errors"
+	"net/http"
+)
+
+func SendRequestToCentral(ctx context.Context, requestBody interface{}, address string, pass string) (*http.Response, error) {
```
SimonBaeumer, line 12: Add a small struct to the central client called `Client` which is initliazed with the endpoint URL.

```
type Client struct {
  address string
}

func NewCentralClient(endpoint string) *Client {
  return ...
}
```

After that all functions which send requests to some Central endpoint can be moved to the client to have a canonical location for Central requests.

fleetshard/pkg/centralreconciler/init_auth.go
SimonBaeumer, line 79: This can also be added as a receiver function on the Central client.

SimonBaeumer (changes requested):
@ivan-degtiarenko Changes look good to me. Most comments are around the central client, you can fix these in an additional PR if you prefer that to merge this PR.

fleetshard/pkg/central_client/client.go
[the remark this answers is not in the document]
dhaus67: We've had this discussion in the past ([here](https://github.com/stackrox/stackrox/pull/706#discussion_r820515606)).
By sharing the transport, we will make use of connection pooling, as we use the same transport across multiple requests. If we were to re-create the transport for each request, we wouldn't benefit from this.

I don't really agree with the "package-wide variables" sentiment as an argument against it, since the package is so small and its anyways unexported (also, could be used for stubbing within tests if we really wanted to).

fleetshard/pkg/centralreconciler/init_auth.go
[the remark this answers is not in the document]
ivan-degtiarenko: That's what ROX-11619 is for :)

fleetshard/pkg/central_client/client.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	req, err := http.NewRequest(http.MethodPost, address, bytes.NewBuffer(jsonBytes))
+	if err != nil {
+		return nil, errors.Wrap(err, "creating HTTP request to central")
+	}
+	req.SetBasicAuth("admin", pass)
+	req = req.WithContext(ctx)
+
+	insecureTransport := http.DefaultTransport.(*http.Transport).Clone()
+	// TODO: once certificates will be added, we probably will be able to replace with secure transport
```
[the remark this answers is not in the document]
ivan-degtiarenko: Created https://issues.redhat.com/browse/ROX-11795

fleetshard/pkg/central_client/client.go
ivan-degtiarenko: Sounds good. I also extended `Client` with some other repeating fields like `pass`, `central` and `httpClient`

fleetshard/pkg/central/client/client.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/private"
+	acsErrors "github.com/stackrox/acs-fleet-manager/pkg/errors"
+	"github.com/stackrox/rox/generated/storage"
+	"github.com/stackrox/rox/pkg/httputil"
+	"net/http"
+)
+
+// reusing transport allows us to benefit from connection pooling
```
SimonBaeumer, line 17: Thanks!

SimonBaeumer (approved):

PR #183, opened 2022-07-08
Drop check for fleetshard-sync rbac existence

kurlov:
## Description

Drop fleetshard-sync RBAC role lookup in the helm chart. With such check only every second deployment will create correct RBAC role for fleet.
Thanks @mtodor for spotting it


mtodor (approved):
Thank you for the fix @kurlov!

PR #184, opened 2022-07-08
Fix detect-secrets pre-commit hook.

dhaus67:
## Description

When running the pre-commit against all files via:
```shell
$ pre-commit run --all-files
```

The `detect-secrets` had issues regarding undetected secrets.

Updated the `.secrets.baseline` file + audited it once more to make the pre-commit hook pass:
```shell
$ detect-secrets scan > .secrets.baseline
$ detect-secrets audit .secrets.baseline
```

## Test manual

```shell
# Run pre-commit against all files and expect no errors.
$ pre-commit run --all-files
```

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -135,6 +135,48 @@
         "is_secret": false
       }
     ],
+    "config/jwks-file-static.json": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "config/jwks-file-static.json",
+        "hashed_secret": "551c2aa179bc3c0e2e8176d4b458d077ed358e25",
+        "is_verified": false,
+        "line_number": 8,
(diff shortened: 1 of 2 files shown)


mtesseract:
Do you understand what the underlying problem is/was?
I would have expected that with an outdated baseline file you would potentially only get more false positives at commit time.

mtesseract (approved):

SimonBaeumer (approved):

dhaus67:
@mtesseract that's essentially the case, right now you would get the false-positives for each non-existent secret within `secrets.baseline` that are not already set to `is_secret: false`. To get rid of the false-positives and make the pre-commit hook pass, we need to update the `secrets.baseline` file + audit it once more, to get rid of the false-positives.

PR #196, opened 2022-07-13
added cluster config for local CRC

johannes94:
## Description
Added cluster config for Code Ready Containers (CRC)

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/dev/config/dataplane-cluster-configuration-crc.yaml
+++ b/dev/config/dataplane-cluster-configuration-crc.yaml
@@ -1,0 +1,22 @@
+---
+#- A list of clusters for fleet manager
+#- The `cluster_id` field can not be empty
+#- All clusters in fleet manager DB already but are missing in the list will be marked as
+#  'deprovisioning' and will later be deleted.
+#- This list is ordered, any new cluster should be appended at the end.
+clusters:
+ - name: crc-admin  # Uncomment if using minikube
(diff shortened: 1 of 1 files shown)


kovayur:
WDYT about merging all kinds of local clusters into a single file?

SimonBaeumer (approved):

johannes94:
> WDYT about merging all kinds of local clusters into a single file?

Do you mean putting them all into one file and commenting the configs you are not using? I like the way it is currently working, cause you can specify which file to use when running fleet-manager without changing something in the files.

kovayur:
> Do you mean putting them all into one file and commenting the configs you are not using? I like the way it is currently working, cause you can specify which file to use when running fleet-manager without changing something in the files.
I mean that we can leave multiple clusters with different IDs and names for each cluster type (without commenting). Then we don't need to change something in the files either. Moreover we can switch, say, from `colima` to `crc` without restarting the fleet manager

johannes94:
Good idea @kovayur. I tried to test and run into the problem that fleet-manager crashes if a given cluster in the config is not found in the current kubeconfig you're running fleet-manager with. Looks to me like that would only work if one would have all supported cluster types in his kubeconfig. Is it OK if I merge, and we revisit another time?

kovayur:
Sure, no objections @johannes94. Thanks for checking.

PR #199, opened 2022-07-13
Polishing of Test Environment

mtesseract:
PR Notes

This PR contains some restructuring and streamlining of the test env setup scripts.

1. Previously there were two major configuration settings: the `CLUSTER_TYPE` (which could be "openshift" or "minikube") on the one hand and additionally there was the `OPENSHIFT_CI` setting injected from, well, OpenShift CI. I found it unnecessary confusing to have these two settings separate even though they share a huge functional overlap. Furthermore, having these settings separately allows to easily configure unsupported modes of operation (e.g., what would `CLUSTER_TYPE=minikube` and `OPENSHIFT_CI=true` mean?). I have streamlined this in the following way: For the actual setup logic only `CLUSTER_TYPE` is relevant; a new cluster type "openshift-ci" is introduced, which models exactly the OpenShift CI flavour of the "openshift" cluster type. Right at initialization we convert the `OPENSHIFT_CI` setting into the `CLUSTER_TYPE=openshift-ci` setting and afterwards we can forget that `OPENSHIFT_CI` exists.

   This approach also allows simplifying the mechanism for loading the different configuration profiles. Previously we had the cluster-type specific configurations plus the OpenShift CI specific configuration profile (openshift-ci.env) which had to be loaded last.

   Now we simply load cluster-type specific configuration files (multiple, ordered) from cluster-type specific directories. No `openshift-ci.env` needed anymore.

2.  The Makefile from the fleet-manager template already references the environment variables `QUAY_USER` and `QUAY_TOKEN`. I have therefore replaced the usage of `QUAY_IO_USERNAME` & `QUAY_IO_PASSWORD` with `QUAY_USER` & `QUAY_TOKEN`.

3. I have restructured the building and tagging of images for local deployment purposes. This change:

   ```
    image/build: GOOS=linux
    image/build: GOARCH=amd64
    image/build: fleet-manager fleetshard-sync
   -	docker --config="${DOCKER_CONFIG}" build -t "$(image_repository):$(image_tag)" .
   +	docker --config="${DOCKER_CONFIG}" build -t "$(external_image_registry)/$(image_repository):$(image_tag)" .
   ```
   
   reverts the definition of this Makefile target to the previous version from the template. I did that change originally because I considered the referenced external image registry to be an inappropriate setting for our situation. (If this turns out to actually be the case, we should simply change the value of external_image_registry instead ). At the same time for
   local building of images for testing there is no canonical hostname for the image references; a plain "fleet-manager:tag" style would be sufficient and clearer to process
   within the scripts. Therefore I have added the following target:
   
   ```
   image/build/local: GOOS=linux
   image/build/local: GOARCH=amd64
   image/build/local: IMAGE_TAG ?= $(image_tag)
   image/build/local: image/build
   	docker image tag "$(external_image_registry)/$(image_repository)/fleet-manager:$(IMAGE_TAG)" "fleet-manager:$(IMAGE_TAG)"
   .PHONY: image/build/local
   ```
   
   which basically retags the built image as `fleet-manager:tag`. This change allows us to get rid of distinguishing between `FLEET_MANAGER_IMAGE` and `FULL_FLEET_MANAGER_IMAGE`, which reduces complexity.
   
4. `CLUSTER_ID` is now initialized, makes it more convenient to work with the e2e test suite in the environment context of our scripts.

5. Remove unnecessary environment variables (there used to be e.g. `POSTGRES_USER` and `DATABASE_USER`).

6. The files `dev/env/defaults/*.env` implement specific defaulting logic. This allows to reduce the noise in `lib.sh` somewhat by keeping defaulting logic in well-defined and separated units. For example, `dev/env/defaults/01-operator-source.env` takes care of setting the default value for `OPERATOR_SOURCE` to `quay` if the operator shall be installed on a non-openshift cluster. Or `02-imagepullsecrets.env` takes care of setting `INHERIT_IMAGEPULLSECRETS_DEFAULT="true"` in case some images have to be downloaded from Quay.

7. Moved some Kubernetes manifests (clusterrole, rolebinding, configmap) from the cluster-type specific folders to `shared/`.

8. Simplify the transferring of images to Minikube significantly: Previously one had to use two different local docker registries: the one from your host system (e.g. from Docker Desktop) and then the one from Minikube's built-in Docker. An image was built in the former and then transferred (hopefully) to the latter. As you know, this was unreliable sometimes.

   Now this works differently: When you intend to use Minikube, image building and image pulling will be done directly on Minikube's built-in registry using `minikube docker-env`.

   This allowed me to remove the ugly Minikube hacks for making sure that the image is transferred correctly.

9. Set `DOCKER_CONFIG` explicitly in order to be able to do `docker login`. This does not seem to work when using the default docker config.json in the context of `minikube docker-env` because Minikube's docker won't have access to the host system's credentials store. Make docker executable used in the Makefile to be configurable using DOCKER environment variable. Makes it easier to switch between Minikube, Colima, etc.

10. Replaced wrong interpolation of `$..._RESOURCES_DEFAULT` variabales with `$..._RESOURCES` in the deployment manifests.

11. Removed `assemble-kubeconfig` script and moved its functionality into a new function in lib.sh. Reworked the SERVER_OVERRIDE hack that is necessary in some situations. The new function for assembling the kubeconfig retrieves the actual IP address under which the K8s API server is reachable from within the cluster by spawning a Pod and retrieving the address from its environment. This implies that `KUBECONF_CLUSTER_SERVER_OVERRIDE` is now a boolean setting.

12. `CLUSTER_NAME` is not configurable anymore, it is taken from the kubeconfig directly (nicer experience when deploying to Infra, for instance).

13. Prevent operator from being torn down in down.sh, which allows reusing an existing operator, speeding up bootstrapping.

14. Added logging messages.

15. Remove implicit `minikube start` during the bootstrapping. The user will need to make sure that whatever cluster they use that it is available (and in the current context).

16. Remove half-baked and potentially dangerous (because of leaks) debugging mechanism. If that sort of debugging is needed, just add a `set -x` locally into the script, for example into the `init` function.

17. Apply more shellcheck lint fixes.


.openshift-ci/tests/e2e-test.sh
```diff
@@ -15,24 +15,28 @@ $KUBECTL -n "$ACSMS_NAMESPACE" logs -l application=fleet-manager -c fleet-manage
...
-    FAIL=1
-fi
-T1=$(date "+%s")
-DELTA=$((T1 - T0))
-
-if [[ $FAIL == 0 ]]; then
-    log
-    log "** E2E TESTS FINISHED SUCCESSFULLY ($DELTA seconds) **"
-    log
+if [[ "$SKIP_TESTS" == "true" ]]; then
```
mtesseract, line 19: This is sometimes useful for development.

.openshift-ci/tests/e2e-test.sh
```diff
@@ -15,24 +15,28 @@ $KUBECTL -n "$ACSMS_NAMESPACE" logs -l application=fleet-manager -c fleet-manage
...
+    log "Skipping tests"
 else
-    log
-    log "** E2E TESTS FAILED ($DELTA seconds) **"
-    log
+    log "Next: Executing e2e tests"
```
mtesseract, line 22: Code on the last has been moved into this `else` branch.

.openshift-ci/tests/e2e.sh
```diff
@@ -22,23 +20,28 @@ if [[ -n "$OPENSHIFT_CI" ]]; then
         log "Got secret ${secret_name}"
         export "${secret_name}"="${secret_value}"
     done
+    export STATIC_TOKEN="${FLEET_STATIC_TOKEN:-}"
```
mtesseract, line 23: Single point where we convert OpenShift specific settings into our internally used representation.

.openshift-ci/tests/e2e.sh
```diff
@@ -22,23 +20,28 @@ if [[ -n "$OPENSHIFT_CI" ]]; then
...
+log "Cluster type: ${CLUSTER_TYPE}"
+log "Cluster name: ${CLUSTER_NAME}"
 log "Image: ${FLEET_MANAGER_IMAGE}"
 if [[ "$SPAWN_LOGGER" == "true" ]]; then
     LOG_DIR=$(mktemp -d)
     export LOG_DIR
     log "Log directory: ${LOG_DIR}"
 fi
 
-if [[ -n "$OPENSHIFT_CI" ]]; then
```
mtesseract, line 36: Has been moved up.

.openshift-ci/tests/e2e.sh
```diff
@@ -48,11 +51,6 @@ OCM)
     ;;
 
 STATIC_TOKEN)
-    # FLEET_STATIC_TOKEN is the name of the secret in Vault,
```
mtesseract, line 51: Has also been moved up ("single point").

.openshift-ci/tests/e2e.sh
```diff
@@ -63,24 +61,20 @@ esac
 log
 
 if [[ "$INHERIT_IMAGEPULLSECRETS" == "true" ]]; then
-    if [[ -z "${QUAY_IO_USERNAME:-}" ]]; then
-        die "QUAY_IO_USERNAME needs to be set"
+    if [[ -z "${QUAY_USER:-}" ]]; then
```
mtesseract, line 64: Using standard variables already used in Makefile instead of new ones.

.openshift-ci/tests/e2e.sh
```diff
@@ -63,24 +61,20 @@ esac
...
-KUBE_CONFIG=$(assemble-kubeconfig | yq e . -j - | jq -c . -)
-export KUBE_CONFIG
-
 # Configuration specific to this e2e test suite:
 export ENABLE_DB_PORT_FORWARDING="false"
 
 bootstrap.sh
 
-if [[ -z "$OPENSHIFT_CI" ]]; then
+if [[ "$CLUSTER_TYPE" != "openshift-ci" ]]; then
```
mtesseract, line 77: See first point in the PR description. This simplifies the overall logic.

.openshift-ci/tests/e2e.sh
```diff
@@ -149,6 +148,7 @@ fi
 
 if [[ "$FINAL_TEAR_DOWN" == "true" ]]; then
     down.sh
+    delete "${MANIFESTS_DIR}/rhacs-operator" || true
```
mtesseract, line 151: Not anymore done in down.sh, because down.sh is already called in the beginning to clean the table and we want to be able to reuse running operators.

Makefile
```diff
@@ -744,16 +754,12 @@ deploy/token-refresher:
...
-		docker run -it -v $(DOCS_DIR)/mermaid-diagrams-source:/data -v $(DOCS_DIR)/images:/output minlag/mermaid-cli -i /data/`basename $${f}` -o /output/`basename $${f} .mmd`.png; \
+		$(DOCKER) run -it -v $(DOCS_DIR)/mermaid-diagrams-source:/data -v $(DOCS_DIR)/images:/output minlag/mermaid-cli -i /data/`basename $${f}` -o /output/`basename $${f} .mmd`.png; \
 	done
 .PHONY: docs/generate/mermaid
 
 tag:
 	@echo "$(image_tag)"
 .PHONY: tag
 
-image-tag:
```
mtesseract, line 755: Not needed anymore.

dev/env/defaults/00-defaults.env
```diff
@@ -1,30 +1,33 @@
...
 export STACKROX_OPERATOR_NAMESPACE_DEFAULT="stackrox-operator"
-export OPENSHIFT_CI="${OPENSHIFT_CI:-}"
-export FLEET_MANAGER_IMAGE_DEFAULT="$(make -s -C "$GITROOT" image-tag)"
-export KUBECONFIG_DEFAULT="$HOME/.kube/config"
+export FLEET_MANAGER_IMAGE_DEFAULT="fleet-manager:$(make -s -C "$GITROOT" tag)"
 export ENABLE_DB_PORT_FORWARDING_DEFAULT="false"
 export ENABLE_FM_PORT_FORWARDING_DEFAULT="false"
 export OPENSHIFT_MARKETPLACE_DEFAULT="false"
-export INSTALL_OPERATOR_DEFAULT="false"
+export INSTALL_OPERATOR_DEFAULT="true"
```
mtesseract, line 21: We need operators nowadays.

dev/env/defaults/00-defaults.env
```diff
@@ -45,8 +48,10 @@ export DUMP_LOGS_DEFAULT="false"
 export OPERATOR_SOURCE_DEFAULT="quay"
 export INSTALL_OLM_DEFAULT="true"
 export FINAL_TEAR_DOWN_DEFAULT="false"
+export DOCKER_CONFIG_DEFAULT="${GITROOT}/.docker"
+export SKIP_TESTS_DEFAULT="false"
 
-export KUBECONF_CLUSTER_SERVER_OVERRIDE_DEFAULT=""
+export KUBECONF_CLUSTER_SERVER_OVERRIDE_DEFAULT="false"
```
mtesseract, line 54: This is now a boolean.

dev/env/defaults/01-minikube.env
```diff
@@ -0,0 +1,5 @@
+# shellcheck shell=bash
```
mtesseract, line 1: Smart defaulting: If the cluster name is minikube, then we derive from this the cluster type, which does not need to be specified by the user.

dev/env/defaults/10-operator-source.env
```diff
@@ -0,0 +1,6 @@
+# shellcheck shell=bash
```
mtesseract, line 1: again, smart defaulting logic.

dev/env/defaults/20-imagepullsecrets.env
```diff
@@ -0,0 +1,7 @@
+# shellcheck shell=bash
+
+if [[ ("${FLEET_MANAGER_IMAGE:-$FLEET_MANAGER_IMAGE_DEFAULT}" =~ ^quay.io/) ||
```
mtesseract, line 3: auto-sense setting for inheriting IPS.

dev/env/defaults/cluster-type-infra-openshift/env
```diff
@@ -0,0 +1,9 @@
+export OPENSHIFT_MARKETPLACE_DEFAULT="true"
```
mtesseract, line 1: default setting for deploying to infra.

dev/env/defaults/cluster-type-minikube/env
```diff
@@ -1,16 +1,7 @@
 export MINIKUBE="${MINIKUBE:-minikube}"
-export CLUSTER_NAME_DEFAULT="minikube"
-export KUBECONF_CLUSTER_SERVER_OVERRIDE_DEFAULT="https://minikube:8443"
+export KUBECONF_CLUSTER_SERVER_OVERRIDE_DEFAULT="true"
 export ENABLE_DB_PORT_FORWARDING_DEFAULT="true"
 export ENABLE_FM_PORT_FORWARDING_DEFAULT="true"
 export OPERATOR_SOURCE_DEFAULT="quay"
 export INHERIT_IMAGEPULLSECRETS_DEFAULT="true" # pragma: allowlist secret
 export INSTALL_OLM_DEFAULT="true"
-
```
mtesseract, line 9: Wasn't really need in my opinion.

dev/env/scripts/apply
```diff
@@ -9,16 +9,12 @@ source "${GITROOT}/dev/env/scripts/lib.sh"
 init
 
 apply_res() {
-    if [[ "$DEBUG" != "trace" ]]; then
-        disable_debugging
-    fi
     local path="$1"
     local short_path
-    short_path=$(echo "$path" | sed -e "s|^${GITROOT}/||;")
+    short_path=${path/#${GITROOT}/}
```
mtesseract, line 14: shellcheck suggested to use bash builtins for stripping the prefix.

dev/env/scripts/bootstrap.sh
```diff
@@ -38,79 +39,85 @@ inject_ips() {
...
+    create-imagepullsecrets
     inject_ips "$ACSMS_NAMESPACE" "default" "quay-ips"
     inject_ips "$STACKROX_OPERATOR_NAMESPACE" "default" "quay-ips"
 fi
 
 if [[ "$INSTALL_OPERATOR" == "true" ]]; then
     if [[ "$INSTALL_OLM" == "true" ]]; then
         # Setup OLM
-        if ! { operator-sdk olm status 2>&1 | grep -q "no existing installation found"; }; then
+        if { operator-sdk olm status 2>&1 | grep -q "no existing installation found"; }; then
```
mtesseract, line 50: This was a bug.

dev/env/scripts/bootstrap.sh
```diff
@@ -38,79 +39,85 @@ inject_ips() {
...
             operator-sdk olm install
+        else
+            log "OLM already installed..."
         fi
     fi
 
-    if [[ "$OPERATOR_SOURCE" == "quay" ]]; then
-        apply "${MANIFESTS_DIR}"/rhacs-operator/quay/01*
-    fi
+    if is_pod_ready "$STACKROX_OPERATOR_NAMESPACE" "app=rhacs-operator"; then
```
mtesseract, line 58: Don't try to reinstall OLM when it seems to be running already.

dev/env/scripts/bootstrap.sh
```diff
@@ -38,79 +39,85 @@ inject_ips() {
...
 fi
 
-load_image_into_minikube() {
-    local img="$1"
-
-    if $MINIKUBE image ls | grep -q "^${img}$"; then
-        true
-    else
-        $DOCKER pull "${img}" && $DOCKER save "${img}" | $MINIKUBE ssh --native-ssh=false docker load
+if [[ "$CLUSTER_TYPE" == "minikube" ]]; then
```
mtesseract, line 105: Make sure we are logged in into registry.

dev/env/scripts/lib.sh
```diff
@@ -34,53 +32,54 @@ dump_env() {
...
+    GITROOT_DEFAULT=$(git rev-parse --show-toplevel)
+    export GITROOT=${GITROOT:-$GITROOT_DEFAULT}
     set -eu -o pipefail
-    if [[ "$DEBUG" == "trace" ]]; then
-        set -x
-    fi
 
-    source "${GITROOT}/dev/env/defaults/env"
+    # For reading the defaults we need access to the
+    CLUSTER_NAME=$(get_current_cluster_name)
```
mtesseract, line 40: We now base some defaulting logic on the cluster name (in the case of minikube). Therefore we need to initialize the cluster name here already before line 43.

dev/env/scripts/lib.sh
```diff
@@ -110,67 +109,67 @@ init() {
...
         fi
-        sleep 2
+        sleep 1
     done
+
+    log "Timed out waiting for container ${container_name} to appear for pod ${pod_selector} in namespace ${namespace}"
+    return 1
+}
+
+is_pod_ready() {
```
mtesseract, line 145: We use this for checking if the operator is already running.

dev/env/scripts/lib.sh
```diff
@@ -110,67 +109,67 @@ init() {
...
+    local container_name="$3"
+    local timeout="120s"
 
     log "Waiting for pod ${pod_selector} within namespace ${namespace} to become ready..."
-    wait_for_container_to_appear "$namespace" "$pod_selector" "$container"
-    for i in $(seq 10); do
-        if $KUBECTL -n "$namespace" wait --timeout=5s --for=condition=ready pod -l "$pod_selector" 2>/dev/null >&2; then
-            break
-        fi
+    wait_for_container_to_appear "$namespace" "$pod_selector" "$container_name" || return 1
```
mtesseract, line 164: Simplified!

dev/env/scripts/lib.sh
```diff
@@ -200,3 +200,39 @@ wait_for_default_service_account() {
         return 1
     fi
 }
+
```
mtesseract, line 203: Simplified content of the previous `assemble-kubeconfig` script.

dev/env/scripts/up.sh
```diff
@@ -35,50 +41,47 @@ if [[ -z "$OPENSHIFT_CI" ]]; then
         log "Trying to pull image '${FLEET_MANAGER_IMAGE}'..."
         $DOCKER pull "$FLEET_MANAGER_IMAGE"
     fi
-fi
 
-if [[ "$CLUSTER_TYPE" == "minikube" ]]; then
```
mtesseract, line 40: I was able to get rid of this hack.

mtesseract:
Just rebased from main.

.openshift-ci/tests/e2e.sh
```diff
@@ -9,11 +9,9 @@ export GITROOT
...
 source "${GITROOT}/dev/env/scripts/lib.sh"
 
-log
-log "** Entrypoint for ACS MS E2E Tests **"
-log
-
-if [[ -n "$OPENSHIFT_CI" ]]; then
+if [[ "${OPENSHIFT_CI:-}" == "true" ]]; then
+    # We are running in an OpenShift CI context, configure accordingly.
+    log "Executing in OpenShift CI context"
```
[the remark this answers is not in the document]
mtesseract: `set -x` is completely removed to be on the safe side here. If somebody needs that kind of tracing, one has to add `set -x` manually into the init function.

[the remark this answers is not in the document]
mtesseract: And, yes, OpenShift CI should redact all secrets coming from Vault. The only secrets that I'm aware of that we could incidentally leak would be a kubeconfig.

dev/env/README.md
```diff
@@ -0,0 +1,63 @@
...
+                 --apiserver-port=8443 \
+                 --embed-certs=true \
+                 --delete-on-failure=true \
+                 --driver=hyperkit # For example
+```
+
+and that the `docker` CLI is in `PATH` (if not, export `DOCKER=...` accordingly). Furthermore, prepare your environment by setting:
+* `QUAY_USER`
+* `QUAY_TOKEN`
+* `STATIC_TOKEN` for `AUTH_TYPE=STATIC_TOKEN` or `OCM_TOKEN` for `AUTH_TYPE=OCM`
```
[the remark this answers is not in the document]
mtesseract: actually that works out of the box for minikube and infra?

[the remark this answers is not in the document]
johannes94: maybe when using the entry point script but when following the readme and first executing bootstrap.sh I get an error that CLUSTER_TYPE is not set.

dev/env/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Then do:
+
+```
+$ dev/env/scripts/bootstrap.sh
```
johannes94, line 65: I would change the README to default executing `.openshift-ci/test/e2e.sh` and explain other component script in another section. I think this is easier to understand for people that just want to execute the e2e tests

mtesseract: done

johannes94 (approved):
LGTM

PR #213, opened 2022-07-20
Fix central name collision and improve logs to easier identify central

mtodor:
## Description

During testing on `dp-02`, I deployed 2 centrals with the same name but with different users. Imagine a scenario in production:
1. User "Company1" creates central with the name "staging-cluster"
2. User "Company2" creates central with the name "staging-cluster"
In the database, we will have 2 central with different Ids, and everything should work properly.

We have discovered on `dp-02` that this case makes problems.
- fleetshard synchronizer constantly keeps updating centrals with the same name
- that leads to deployment updates and sets the number of replicas to 3 for the scanner
- that resets state that was previously set by horizontal pod autoscaller
- horizontal pod autoscaller adjusts the number of replicas again

All that causes the creation and termination of new pods, over and over again. After some time Kubernetes API server starts to throttle, and the cluster is not behaving properly anymore.

**Changes**

- primary change is in `fleetshard/pkg/runtime/runtime.go` - where we now use unique identifier `central.Id` instead of `central.Metadata.Name` for list of reconcilers
- also, several logs are updated to allow easier identification of related central. The name is simply not sufficient.

- ~[ ] Documentation added if necessary~

## Test manual

1. Create a staging OSD data plane cluster
2. Call fleet manager API with the static token to create a new central. An example:
```
curl --location --request POST "http://localhost:8000/api/rhacs/v1/centrals?async=true" --header "Content-Type: application/json" --header "Accept: application/json" --header "Authorization: Bearer ${STATIC_TOKEN}" --data-raw "{\"name\":\"test-central\",\"cloud_provider\":\"aws\",\"region\":\"us-east-1\",\"multi_az\":true}" --verbose
```
3. Execute the same call, but with an OCM token. An example:
```
curl --location --request POST "http://localhost:8000/api/rhacs/v1/centrals?async=true" --header "Content-Type: application/json" --header "Accept: application/json" --header "Authorization: Bearer $(ocm token)" --data-raw "{\"name\":\"test-central\",\"cloud_provider\":\"aws\",\"region\":\"us-east-1\",\"multi_az\":true}" --verbose
```
4. After that, monitor HPA
```
k get hpa --all-namespaces
```
You should notice that the number of replicas is going up and down between 2 and 3.

Check on pods can also be helpful
```
k get pods --all-namespaces | rg scanner
```


SimonBaeumer (approved):
Thank you, @mtodor!

PR #214, opened 2022-07-20
Don't swallow bootstrapping error logs

mtesseract:
## Description

Until now the bootstrapping was executed directly from within `e2e.sh` before any loggers are started, which means that if the bootstrapping fails then we don't see why it failed.

This PR shuffles the order so that `e2e.sh` takes care of starting loggers before bootstrapping (which is executed as part of the inner `e2e-test.sh` script).

## Test manual

```
# To run tests locally run:
```


kovayur (approved):
LGTM

PR #216, opened 2022-07-21
[test env] Simplify operator group and increase wait timeout

mtesseract:
## Description

Accidentally I have tried to deploy with `CLUSTER_TYPE=openshift` on my local Minikube. After deploying with `CLUSTER_TYPE=minikube` later I noticed that the operator does not come up. The reason was that for the different cluster type profiles a different operator group was created, which is not allowed. I noticed that there is no reason to have two different manifests for different operator groups and hence streamlined that.

Then the operator *still* did not come up (in time), because apparently the waiting timeout (60s by default) was not sufficient. Therefore I have increased it to 3m.

## Test manual

```
# To run tests locally run:
```


dev/env/scripts/bootstrap.sh
```diff
@@ -59,6 +59,9 @@ if [[ "$INSTALL_OPERATOR" == "true" ]]; then
         log "Skipping installation of operator since the operator seems to be running already"
     else
         log "Installing operator"
+
+        apply "${MANIFESTS_DIR}"/rhacs-operator/[0-9]* # This installs the operator-group.
```
vladbologa, line 63: does this apply again the namespace object, already applied on line 29?

vladbologa: also on lines 86 and 89, it could be 0[3]* now instead of 0[23]*, I think

mtesseract: Regarding the first comment: Yes, it does. But that is fine, if the ns already exists it is a no-op. I would suggest to keep it this way because it introduces less magic (of the form "this manifest only at this point" and "this other manifest only at that other point" -- the reason we need to apply the ns already before is to allow the logger to run early).

mtesseract: Regarding the second comment: You are right, that looked weird. I have simplified that.

vladbologa: Thanks, looks clearer now.

vladbologa (approved):
lgtm!

PR #237, opened 2022-07-28
ROX-11862: Remove redundant conversion layer (ManagedCentralCR)

kovayur:
## Description
This is a refactoring PR which removes the unnecessary conversion layer - ManagedCervice CR. We don't have such CRD.

## Checklist (Definition of Done)

- [ ] Unit and integration tests added
- [ ] Documentation added if necessary
- [ ] CI and all relevant tests are passing

## Test manual
Tested on a dedicated OSD cluster

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup OCM_OFFLINE_TOKEN=<ocm-offline-token> OCM_ENV=development
make verify lint binary test test/integration
```

PR #238, opened 2022-07-28
ROX-11862: Match DNS name to the ADR

kovayur:
## Description
This PR aligns the routes creation implementation with URI scheme we decided to use in the ADR.
Previously there was a heuristic approach to determine a hostname using a prefix and subdomain. After this change, UI and data hostnames can be set separately on the fleetmanager with any pattern we want.

## Test manual
Tested manually on a dedicated OSD cluster

```
# To run tests locally run:
```

### DIFF
--- a/internal/dinosaur/pkg/services/data_plane_dinosaur.go
+++ b/internal/dinosaur/pkg/services/data_plane_dinosaur.go
@@ -318,23 +318,22 @@ func (d *dataPlaneDinosaurService) persistDinosaurRoutes(dinosaur *dbapi.Central
 	return nil
 }
 
-func buildRoutes(routesInRequest []dbapi.DataPlaneCentralRouteRequest, dinosaur *dbapi.CentralRequest, clusterDNS string) ([]dbapi.DataPlaneCentralRoute, error) {
-	routes := []dbapi.DataPlaneCentralRoute{}
-	dinosaurHost := dinosaur.Host
-	for _, r := range routesInRequest {
-		if strings.HasSuffix(r.Router, clusterDNS) {
-			router := dbapi.DataPlaneCentralRoute{
-				Router: r.Router,
-			}
-			if r.Prefix != "" {
-				router.Domain = fmt.Sprintf("%s-%s", r.Prefix, dinosaurHost)
-			} else {
-				router.Domain = dinosaurHost
-			}
-			routes = append(routes, router)
-		} else {
-			return nil, errors.Errorf("router domain is not valid. router = %s, expected domain = %s", r.Router, clusterDNS)
-		}
+func buildRoutes(routesInRequest dbapi.DataPlaneCentralRoutesRequest, dinosaur *dbapi.CentralRequest, clusterDNS string) ([]dbapi.DataPlaneCentralRoute, error) {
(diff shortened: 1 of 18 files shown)


kovayur:
* main
  * **PR #237** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/237" target="_blank">Graphite</a> 
    * **PR #238** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/238" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/238?utm_source=stack-comment).

johannes94:
The code changes look good to me. One thing I wanted to make sure is that we are aware of the impact of the change with the DataPlaneCentralStatusRoutes to an object instead of a list. Previously the concept was very generic and we could basically create all kind of different DNS names just by putting different routes in the request from fleetshard-sync to fleet-manager. Fleetshard-sync was aware of what routes should be created and fleet-managers job was just to create the corresponding CNAME records for that. This way only fleetshard-sync would have to change to register a new route and DNS name. The new approach is much easier to understand, since it is tailored exactly to our usecase. But we have the trade off that with the new approach if we'd like to add another route in the future we'd always have to change both fleet-manager and fleetshard-sync. What do you think about that?

kovayur:
Thanks for the thoughtful review @johannes94. I can share the concern and indeed it was a trade-off in favor of simplicity. Need to say that now if we need to add another route, we need to change not only the fleethard sync and the fleet-manager but also the Private API schema, which I don't really like.
On second thought, maybe something more flexible should be done here, I agree.

johannes94 (commented):
Looks good. I'm going to test this together with #239 in our dev environment. If this works I'll approve.

internal/dinosaur/pkg/services/data_plane_dinosaur.go
```diff
@@ -302,9 +302,7 @@ func (d *dataPlaneDinosaurService) persistDinosaurRoutes(dinosaur *dbapi.Central
 	routesInRequest := dinosaurStatus.Routes
 	var routes []dbapi.DataPlaneCentralRoute
 
-	var routesErr error
-	baseClusterDomain := strings.TrimPrefix(clusterDNS, fmt.Sprintf("%s.", constants2.DefaultIngressDNSNamePrefix))
-	if routes, routesErr = buildRoutes(routesInRequest, dinosaur, baseClusterDomain); routesErr != nil {
+	if routesErr := validateRouters(routesInRequest, dinosaur, clusterDNS); routesErr != nil {
```
johannes94, line 305: This change led to routes not being stored, since the routes variable is never set. I'd suggest to remove the `routes` variable above and us `routesInRequest` in all function calls further down the line.

johannes94 (changes requested):

internal/dinosaur/pkg/services/data_plane_dinosaur.go
kovayur: Right, thanks for checking! Fixed.

johannes94 (approved):

PR #241, opened 2022-07-28
added DNS e2e tests

johannes94:
## Description
Add tests for creating routes when route resource is available and also for creating and deleting CNAME record in AWS Route53 when Route53 credentials are available.

## Test manual

- Follow instructions in `howto-run-e2e-tests.md`
- Run against a cluster without openshift routes and make sure DNS and route test is skipped
- Run against a cluster having openshift routes crd using Route53 credentials

```
# To run tests locally run:
```


kurlov (commented):
Looks great!
Have you run it against staging?
The reason I wonder is that we use the same `e2e/e2e_test.go` file for staging e2e

johannes94:
> Looks great! Have you run it against staging? The reason I wonder is that we use the same `e2e/e2e_test.go` file for staging e2e

I did not but we should. By default the DNS tests are skipped if a few env variables are not set. We should configure the test against stage to ensure DNS tests are also executed. I'd need a starting point/doc where I can see how to do that, can you point me to something? @kurlov

e2e/e2e_test.go
```diff
@@ -124,12 +181,67 @@ var _ = Describe("Central", func() {
...
+}
+
+func removeLastChar(s string) string {
+	return s[:len(s)-1]
+}
+
+func getHostedZone(central *public.CentralRequest) (*route53.HostedZone, error) {
+	hostedZones, err := route53Client.ListHostedZones(&route53.ListHostedZonesInput{})
+	if err != nil {
+		return nil, err
```
mtesseract, line 233: I don't know why this is not complained about by the wrapcheck linter (https://github.com/tomarrell/wrapcheck).

e2e/e2e_test.go
```diff
@@ -92,6 +97,58 @@ var _ = Describe("Central", func() {
...
+			if !routesEnabled {
+				Skip(skipRouteMsg)
+			}
+
+			central := getCentral(createdCentral, client)
+
+			var reencryptRoute *openshiftRouteV1.Route
+			Eventually(func() error {
+				reencryptRoute, err = routeService.FindReencryptRoute(context.Background(), namespaceName)
+				return err
```
mtesseract, line 111: I don't know why this is not complained about by the wrapcheck linter (https://github.com/tomarrell/wrapcheck).

e2e/e2e_test.go
```diff
@@ -92,6 +97,58 @@ var _ = Describe("Central", func() {
...
+			Eventually(func() int {
+				records, err = route53Client.ListResourceRecordSets(&route53.ListResourceRecordSetsInput{
+					HostedZoneId:    rhacsZone.Id,
+					StartRecordName: &central.Host,
+				})
+				Expect(err).ToNot(HaveOccurred())
+				return len(records.ResourceRecordSets)
+			}).WithTimeout(waitTimeout).WithPolling(defaultPolling).Should(Equal(1))
+
+			Expect(len(records.ResourceRecordSets)).To(Equal(1))
```
mtesseract, line 141: Isn't this tested already in l.139?

e2e/e2e_test.go
```diff
@@ -92,6 +97,58 @@ var _ = Describe("Central", func() {
...
+			recordSet := records.ResourceRecordSets[0]
+			Expect(len(recordSet.ResourceRecords)).To(Equal(1))
+			record := recordSet.ResourceRecords[0]
+
+			name := removeLastChar(*recordSet.Name)
```
mtesseract, line 146: Is this a trailing dot being removed or something? Would be helpful to add a comment about this.

e2e/e2e_test.go
```diff
@@ -124,12 +181,67 @@ var _ = Describe("Central", func() {
...
-	return central.Status
+	return central
+}
+
+func centralStatus(createdCentral *public.CentralRequest, client *fleetmanager.Client) string {
+	return getCentral(createdCentral, client).Status
+}
+
+func isDNSEnabled(routesEnabled bool) bool {
+	hasRoute53Creds := os.Getenv("ROUTE53_ACCESS_KEY") != "" &&
```
mtesseract, line 220: We are already doing `os.Getenv` on the route53 credentials in `e2e_test_suite.go` -- how about reusing the information already retrieved in one place at the other place?

e2e/e2e_test.go
```diff
@@ -124,12 +181,67 @@ var _ = Describe("Central", func() {
...
+
+func getHostedZone(central *public.CentralRequest) (*route53.HostedZone, error) {
+	hostedZones, err := route53Client.ListHostedZones(&route53.ListHostedZonesInput{})
+	if err != nil {
+		return nil, err
+	}
+
+	var rhacsZone *route53.HostedZone
+	for _, zone := range hostedZones.HostedZones {
+		// Omit the . at the end of hosted zone name
```
mtesseract, line 238: Oh, there it mentions a trailing dot, ok!

e2e/e2e_test.go
```diff
@@ -124,12 +181,67 @@ var _ = Describe("Central", func() {
...
+		name := removeLastChar(*zone.Name)
+		if strings.Contains(central.Host, name) {
+			rhacsZone = zone
+			break
+		}
+	}
+
+	return rhacsZone, nil
```
mtesseract, line 246: Shouldn't this report some sort of "not found" error in case the zone is not found?

e2e/e2e_test.go
```diff
@@ -124,12 +181,67 @@ var _ = Describe("Central", func() {
...
+	hostedZones, err := route53Client.ListHostedZones(&route53.ListHostedZonesInput{})
+	if err != nil {
+		return nil, err
+	}
+
+	var rhacsZone *route53.HostedZone
+	for _, zone := range hostedZones.HostedZones {
+		// Omit the . at the end of hosted zone name
+		name := removeLastChar(*zone.Name)
+		if strings.Contains(central.Host, name) {
```
mtesseract, line 240: Why exactly do we use `Contains` and don't test for equality here?

mtesseract (commented):
Looking good, just some minor remarks and questions.

e2e/e2e_test.go
johannes94: Values could look like this:
```
central.Host = "acs-<id>.rhacs-dev.com"
name = rhacs-dev.com
```
At least this was the case before Yuries changes with the passthrough route. Now I think host field could also be rhacs-dev.com. I'd leave it this way because it does work like we expect and we would break this test if the hostname is changed to some sub domain in the future.

e2e/e2e_test.go
johannes94: Correct, good catch. Fixed.

dev/env/scripts/up.sh
```diff
@@ -66,6 +66,7 @@ log "Database is ready."
 
 # Deploy MS components.
 log "Deploying fleet-manager"
+export FLEET_MANAGER_COMMAND="/usr/local/bin/fleet-manager serve --force-leader --api-server-bindaddress=0.0.0.0:8000 --health-check-server-bindaddress=0.0.0.0:8083 --kubeconfig=/secrets/kubeconfig --enable-central-external-certificate=$ENABLE_CENTRAL_EXTERNAL_CERTIFICATE --central-domain-name='$CENTRAL_DOMAIN_NAME'"
```
mtesseract, line 69:
```suggestion
export FLEET_MANAGER_COMMAND="/usr/local/bin/fleet-manager serve --force-leader --api-server-bindaddress=0.0.0.0:8000 --health-check-server-bindaddress=0.0.0.0:8083 --kubeconfig=/secrets/kubeconfig --enable-central-external-certificate='$ENABLE_CENTRAL_EXTERNAL_CERTIFICATE' --central-domain-name='$CENTRAL_DOMAIN_NAME'"
```

mtesseract (approved):
lgtm

PR #243, opened 2022-07-29
ROX-11407: Support for Configurable Resource Limits for Central & Scanner

mtesseract:
## Description

* Extends the API for modelling resource settings.
* Introduces new admin endpoint for creating centrals with non-default resource settings.

There are plenty of generated files, which should, of course, be filtered out while reviewing.
Also, my editor has (unfortunately) streamlined the used quotation marks ("..." vs '...') in the API YAMLs, hence the diff to these files is unnecessary large.

Implementing the `Update` call for modifying resource limits will be done in a separate PR.

Test has been added.

## Test manual

```
# To run tests locally run:
```


openapi/fleet-manager-private.yaml
```diff
@@ -22,7 +22,7 @@ tags:
 
 paths:
   # Endpoints for data plane communications
-  '/api/rhacs/v1/agent-clusters/{id}/status':
```
kovayur, line 25: I'm wondering if we need a pre-commit hook for [yaml formatting](https://github.com/adrienverge/yamllint)

kovayur (approved):
LGTM, Good job @mtesseract

openapi/fleet-manager-private.yaml
mtesseract: great idea

mtesseract:
@kovayur, I had to rework the PR due to some Issues I have found. Let me explain.

* I did the following change:
```
-	Central     api.JSON `json:"central"` // Schema is defined by #/components/schemas/CentralSpec from the public API.
-	Scanner     api.JSON `json:"scanner"` // Schema is defined by #/components/schemas/ScannerSpec from the public API.
+	Central     api.JSON `json:"central"` // Schema is defined by dbapi.CentralSpec
+	Scanner     api.JSON `json:"scanner"` // Schema is defined by dbapi.ScannerSpec
 ```
 
 which means that the internal representation of the `central` and `scanner` spec is defined by `dbapi.CentralSpec` and dbapi.ScannerSpec` instead of the OpenAPI generated models. I find this more logical (API models only for conversion at service boundaries) and also more reliable, because the models in `dbapi` are more strongly typed.
 
* I have streamlined the conversion functions in `resourceutil.go` and the validation functions in `internal/dinosaur/pkg/handlers/dinosaur.go`(there was also a bug in there).
 
 * `internal/dinosaur/pkg/presenters/dinosaur.go`: Only attempt to unmarshal `central` and `scanner` in case they are non-empty. Same in `internal/dinosaur/pkg/presenters/managedcentral.go`.
 
* Fix a rebase error in `internal/dinosaur/pkg/presenters/managedcentral.go`.
 
 * Fix a bug in `route_loader.go` which caused the wrong dinosaur handler to react to admin calls.
 
 
* Then finally: The internal refactoring of the types triggered a rather mysterious issue in `internal/dinosaur/pkg/config/dataplane_cluster_config.go`: due to a transitive (non-obvious) dependency on the k8s `controller-runtime` package we were hit by https://github.com/kubernetes-sigs/controller-runtime/issues/878. This caused `fleet-manager` to panic during startup. Until this is fixed in upstream I have added an ugly workaround, which is still the best I was able to come up with in the above file. It checks if the flag has already been defined and if so it modifies it to be in the shape that our application code expects. This seems to work nicely.

kovayur (approved):

kovayur:
Thanks for the explanation @mtesseract 

> I find this more logical (API models only for conversion at service boundaries)
Yes, this is more coherent, I agree

kovayur (approved):

kovayur (approved):

PR #257, opened 2022-08-02
CI E2E Tests: Make sure that RUN_AUTH_E2E variable is initialized

mtesseract:
## Description

If it is not set in the environment the e2e script will abort with an error.

I noticed some linter complaints (shellcheck) and have fixed the issues in one go.
And I have removed a secrets from the baseline and added the pragma annotation in the shell script directly to reduce merge conflicts in the future.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/.openshift-ci/tests/e2e.sh
+++ b/.openshift-ci/tests/e2e.sh
@@ -9,6 +9,8 @@ export GITROOT
 # shellcheck source=/dev/null
 source "${GITROOT}/dev/env/scripts/lib.sh"
 
+export RUN_AUTH_E2E="false"
+
 if [[ "${OPENSHIFT_CI:-}" == "true" ]]; then
     # We are running in an OpenShift CI context, configure accordingly.
     log "Executing in OpenShift CI context"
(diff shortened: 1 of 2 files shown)


.openshift-ci/tests/e2e.sh
```diff
@@ -27,7 +29,6 @@ if [[ "${OPENSHIFT_CI:-}" == "true" ]]; then
     export GOARGS="-mod=mod" # For some reason we need this in the offical base images.
     # When running in OpenShift CI, ensure we also run the auth E2E tests.
     RUN_AUTH_E2E="true"
-    export RUN_AUTH_E2E
```
dhaus67, line 30: Won't this now always have `RUN_AUTH_E2E` set to false since we don't update the environment variable and not run the auth E2E tests?

mtesseract: Of course. 🤦

dhaus67 (approved):
Thanks for fixing this!

PR #258, opened 2022-08-02
Test Env: Add additional wait during setting up of the operator

mtesseract:
## Description

Locally I was able to reproduce a bootstrapping issue (on completely fresh Colima & Rancher Desktop clusters) where the operator did not come up because the newly created subscription ended up in a broken state:

```
  Conditions:
    Last Transition Time:  2022-08-02T10:13:20Z
    Message:               all available catalogsources are healthy
    Reason:                AllCatalogSourcesHealthy
    Status:                False
    Type:                  CatalogSourcesUnhealthy
    Message:               error using catalog operatorhubio-catalog (in namespace olm): failed to list bundles: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 10.43.35.218:50051: i/o timeout"
    Reason:                ErrorPreventedResolution
    Status:                True
    Type:                  ResolutionFailed
  Last Updated:            2022-08-02T10:13:20Z
Events:                    <none>
```

When I investigated I found out that the service 10.43.35.218:50051 was in fact reachable at the time I tried to connect to it. Hence I am thinking we have a race here: It seems it can happen that the subscription fails early because the catalogsource pod is *not yet* ready. So I have implemented a waiting step which makes sure that the operatorhubio-catalog pod is healthy before the subscription is created. With this change I attempted to reproduce the problem multiple times, but did not succeed.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/dev/env/scripts/bootstrap.sh
+++ b/dev/env/scripts/bootstrap.sh
@@ -81,15 +81,24 @@ if [[ "$INSTALL_OPERATOR" == "true" ]]; then
                     jq -r '.items[].metadata.name' | grep -q '^rhacs-operator$' && break
                 sleep 1
             done
-        fi
 
-        if [[ "$OPERATOR_SOURCE" == "quay" ]]; then
+            # It seems that before creating the subscription (part of the next apply call) all catalog sources need to be healthy,
+            # otherwise the subscription will end up in the following state:
+            # Conditions:
+            #   Message:               all available catalogsources are healthy
+            #   Reason:                AllCatalogSourcesHealthy
+            #   Status:                False
+            #   Type:                  CatalogSourcesUnhealthy
+            #   Message:               error using catalog operatorhubio-catalog (in namespace olm): failed to list bundles: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 10.43.96.123:50051: i/o timeout"
+            #   Status:                True
+            #   Type:                  ResolutionFailed
+            #
+            # Therefore we wait for the operatorhubio-catalog/registry-server container to become ready.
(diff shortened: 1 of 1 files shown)


dev/env/scripts/bootstrap.sh
```diff
@@ -81,22 +81,33 @@ if [[ "$INSTALL_OPERATOR" == "true" ]]; then
...
-        elif [[ "$OPERATOR_SOURCE" == "marketplace" ]]; then
-            apply "${MANIFESTS_DIR}"/rhacs-operator/marketplace/*.yaml
-        fi
 
-        if [[ "$OPERATOR_SOURCE" == "quay" ]]; then
             # Apparently we potentially have to wait longer than the default of 60s sometimes...
             wait_for_resource_to_appear "$STACKROX_OPERATOR_NAMESPACE" "serviceaccount" "rhacs-operator-controller-manager" 180
             inject_ips "$STACKROX_OPERATOR_NAMESPACE" "rhacs-operator-controller-manager" "quay-ips"
 
             # Wait for rhacs-operator pods to be created. Possibly the imagePullSecrets were not picked up yet, which is why we respawn them:
```
dhaus67, line 106: (not necessarily to be done within this PR, but in general a worthwhile improvement IMO):
We could create a modified version of `wait_for_container_to_appear` which does not require the container to be running and use it here instead of `sleep 2` (if for whatever reason creating of the rhacs-operator pods takes longer than expected)

dhaus67 (approved):

dev/env/scripts/bootstrap.sh
mtesseract: good point. There is still room for improvement in dealing with all these depending steps.

PR #259, opened 2022-08-02
Add shellcheck pre-commit hook

dhaus67:
## Description

Inspired by [this PR](https://github.com/stackrox/acs-fleet-manager/pull/257) which fixed some shellcheck issues. This will introduce the shellcheck hook, ensuring all our scripts adhere to the same standard.

## Test manual

```
pre-commit run shellcheck --all-files # should succeed
```


dhaus67:
* main
  * **PR #259** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/259" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/259?utm_source=stack-comment).

mtesseract:
you can read my thoughts.

.githooks/pre-commit
```diff
@@ -17,7 +17,8 @@ staged_go_files=$(git diff --cached --name-only | grep ".go$" | awk -F/ '{print
 go_vet_results=""
 for file in $staged_go_files
 do
-    go_vet_results+=$(go vet ./${file}/... 2>&1 >/dev/null)
+    # shellcheck disable=SC3024
```
mtesseract, line 20: Couldn't you get (properly) rid of this by using `#!/usr/bin/env bash` as she-bang instead of POSIX `sh`?
Any reason to stick with `sh`?

dev/env/scripts/lib.sh
```diff
@@ -208,6 +210,7 @@ wait_for_resource_to_appear() {
 
     log "Waiting for ${kind}/${name} to be created in namespace ${namespace}"
 
+    # shellcheck disable=SC2034
     for i in $(seq "$seconds"); do
```
mtesseract, line 214: Wouldn't this be properly fixed by
```suggestion
    for _ in $(seq "$seconds"); do
```

dev/env/scripts/lib.sh
```diff
@@ -244,7 +247,7 @@ assemble_kubeconfig() {
     if [[ "$KUBECONF_CLUSTER_SERVER_OVERRIDE" == "true" ]]; then
         local server
         $KUBECTL delete pod alpine >/dev/null 2>&1 || true
-        server=$($KUBECTL run --rm -it alpine --quiet --image=alpine --restart=Never -- sh -c 'echo $KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT' | tr -d '\r')
+        server=$($KUBECTL run --rm -it alpine --quiet --image=alpine --restart=Never -- sh -c "echo $KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT" | tr -d '\r')
```
mtesseract, line 250: I think this is wrong. We need to interpolation within the pod, not outside?

mtesseract (changes requested):

scripts/fmcurl
```diff
@@ -14,4 +14,4 @@ shift
 # Normalize
 resource=$(echo "$resource" | sed -e 's/^\///;')
 
-curl -LH "Authorization: Bearer ${OCM_TOKEN}" http://localhost:8000/api/${resource} "$@" | jq .
+curl -LH "Authorization: Bearer ${OCM_TOKEN}" http://localhost:8000/api/"${resource}" "$@" | jq .
```
kovayur, line 17:
```suggestion
curl -LH "Authorization: Bearer ${OCM_TOKEN}" "http://localhost:8000/api/${resource}" "$@" | jq .
```

mtesseract:
@dhaus67 Shall we try to get this merged?

.githooks/pre-commit
dhaus67: No reason I'd be aware of - moved to bash.

dhaus67:
@mtesseract Yeah, sorry. Lost track on this PR after PTO

dev/env/scripts/lib.sh
mtesseract: I believe the current state of the code is wrong:
```
server=$($KUBECTL run --rm -it alpine --quiet --image=alpine --restart=Never -- sh -c echo $KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT | tr -d '\r')
 ```
 
 We need the single quotes to prevent early interpolation.

mtesseract: I mean, as it was before?

dhaus67: this is awkward. now it should be the correct command again (as in before).

mtesseract (approved):

mtesseract:
🎉

PR #262, opened 2022-08-03
ROX-11987: Migrate admin API middleware to internal SSO

dhaus67:
## Description

This PR migrates the middleware used for the admin API to make use of the internal SSO (auth.redhat.com) based on the option decided within [this ADR](https://github.com/stackrox/architecture-decision-records/blob/main/managed_service/ADR-0011-fleet-manager-middleware-admin-api.md).

The PR will do the following:
- Add configuration options for the role middleware used by the admin API, allowing to configure the required groups for API access.
- Add the IAM config for the internal SSO, allowing to validate tokens issued by auth.redhat.com (and its staging counterpart).
- Add configuration files for the roles based on different environments (dev/stage & prod), as well as allowing auth.stage.redhat.com.

What's been done outside of this PR as administrative work:
- The existing rover group `acs-general-engineering`, which exists within both `rover.redhat.com/groups & rover.stage.redhat.com/groups`, was re-used to authorize access in dev / staging environments.
- The rover groups `acs-fleet-manager-admin-full, acs-fleet-manager-admin-write, acs-fleet-manager-admin-read` have been created, owners are `cgorman, ebenshet, aruklets, dhaus, syethadk, misberne` temporarily, as ownership will be moved to SREs later on.

What's not been done within this PR:
- Currently, testing can only be done limited. Fleet manager expects a token issued by auth.redhat.com, which is not as easy to get due to not having client credentials / offline tokens such as the ocm token. I'm currently asking around how to best test this and will create a follow-up which will most likely add an automated E2E test to the existing auth test suite.

## Test manual

```
# Create a static token which has the correct issuer (i.e. auth.redhat.com) and the correct realm_roles assigned based on the documentation under `docs/development/test-locally-static-token.md`

# Try and access the admin API via curl and the static token as bearer
curl -H "Authorization: Bearer ${static_token}" http://<fleet manager endpoint>:8000/api/rhacs/v1/admin/dinosaurs

# Check that the request is executed successfully
```


dhaus67:
* main
  * **PR #262** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/262" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/262?utm_source=stack-comment).

akoserwal:
@dhaus67 For stage, are you planning to use auth stage as an issuer?

dhaus67:
@akoserwal yea, we have configured it within `config/additional-sso-issuers`. The rover groups within `config/role-authz-roles-dev` also include rover groups which are staging only.

akoserwal:
If we can use different groups but use auth prod only? wdyt?

dhaus67:
Hm, different groups can already be done - you can configure arbitrary amount of groups.
I just added the auth.stage.redhat.com in case we ever need it within staging - I don't have too strong of an opinion on it, if we don't deem it necessary we can also drop it 🤷

internal/dinosaur/pkg/routes/route_loader.go
```diff
@@ -54,6 +54,7 @@ type options struct {
 	AccessControlListMiddleware *acl.AccessControlListMiddleware
 	AccessControlListConfig     *acl.AccessControlListConfig
 	FleetShardAuthZConfig       *auth.FleetShardAuthZConfig
+	RoleAuthZConfig             *auth.RoleAuthZConfig
```
mtodor, line 57: Just a thought: Maybe it would be valuable to adjust the name of this variable to indicate that it's used for `admin` routes.
```suggestion
	AdminRoleAuthZConfig             *auth.AdminRoleAuthZConfig
```
That would pull a lot of configuration renaming. Including `pkg/auth/roles_authz.go` and options there, and so on. Middleware itself is ok because it's generic middleware that validates roles.

My main concern is that configurations are starting to be intertwined between admin/internal/customer API. And if you look at command options, it's hard to grasp what is used in the end for specific API authorization. It's not something we should tackle right now, but maybe we should consider some refactoring in the future.

mtodor (approved):
This looks nice! I've just added a few thoughts, nothing that changes functionality.

internal/dinosaur/pkg/routes/route_loader.go
dhaus67: I think it's worthwhile to do it within this PR, I've aligned the naming to the `fleetshard-authz` middleware, i.e. re-naming config files + flags.

PR #265, opened 2022-08-04
Disable admin password once sso.redhat.com auth provider is initialised

ivan-degtiarenko:
## Description
The goal is to disable admin-password generation(and thus login via admin/password in Central) after `sso.redhat.com` auth provider is created.
* auth provider should be created once
* admin-password can be disabled multiple times - idempotent change
The PR is done as part of https://issues.redhat.com/browse/ROX-11018

## Test manual

1. Deployed and created Central request to fleet-manager
2. Verified that:
* central is initialised
* `central-htpasswd` secret is deleted


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -63,7 +65,7 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 	remoteCentralName := remoteCentral.Metadata.Name
 	remoteCentralNamespace := remoteCentral.Metadata.Namespace
 
-	if !changed && !r.createAuthProvider && remoteCentral.RequestStatus == centralConstants.DinosaurRequestStatusReady.String() {
+	if !changed && r.authInitStatus == AuthInitCompleted && remoteCentral.RequestStatus == centralConstants.DinosaurRequestStatusReady.String() {
```
dhaus67, line 68: Should we factor those out into separate helper functions instead of big if's? We do something similar in the image enricher with enrichtment context, I think it might be worthwhile here to make the intent clear (also, if there's anything else we may add in the future which makes this more complex).

fleetshard/pkg/central/reconciler/init_auth_status.go
```diff
@@ -0,0 +1,17 @@
+package reconciler
+
+// AuthInitStatus represents state machine for auth initialisation in Managed Central.
+// There are two possible execution flows for this state machine:
+// CreateRhSso -> DisableAdminPassword -> AuthInitCompleted
+// or
+// AuthInitCompleted
+type AuthInitStatus int
```
dhaus67, line 8: nit: do you think its worthwhile to expose this? I don't see any other usages and not sure where it'd make sense.

dhaus67 (commented):
I assume the issue discussed in the slack thread w.r.t. the update has been resolved, right?

fleetshard/pkg/central/reconciler/init_auth_status.go
```diff
@@ -0,0 +1,17 @@
...
+
+const (
+	// CreateRhSso signals reconciler to create sso.redhat.com auth provider.
+	CreateRhSso AuthInitStatus = iota
```
SimonBaeumer, line 12:
```suggestion
	CreateRHSSOAuthStatus AuthInitStatus = iota
```

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -126,16 +131,22 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 		// TODO(create-ticket): implement update logic
 		glog.Infof("Update central %s/%s", central.GetNamespace(), central.GetName())
 
-		err = r.incrementCentralRevision(central)
+		err = r.incrementCentralRevision(&existingCentral)
 		if err != nil {
 			return nil, err
 		}
+		(&existingCentral).Spec = *central.Spec.DeepCopy()
```
misberner, line 138:
```suggestion
		existingCentral.Spec = *central.Spec.DeepCopy()
```

SimonBaeumer (dismissed):
Can you determine the necessary logic to setup the auth provider sequentially without a state machine?
A state machine adds a lot of complexity, especially if debugging a specific execution path.

Implementing it without a state machine could look like this:

 - Check if admin password is enabled -> create auth provider
 - Auth provider exists and admin password enabled -> disable admin password
 - Admin password disabled -> nothing to do

misberner (changes requested):
I think overall, this approach is not how a reconciler is supposed to work.

A reconciler has an input spec which gives rise to a desired state. The reconciler looks at the _actual_ state and performs changes to the effect that the actual state changes into the desired state.

A key concept is that the reconciler is, by itself, mostly stateless - the desired state should simply depend on the input spec, and the actions should only depend on the actual vs desired state. If we want any kind of stateful logic, this state has to be externalized to the actual (cluster) state, it cannot be kept in the in-memory state of the reconciler (what if the reconciler crashes and comes back up?)

I think this is, in part, already an issue with the existing code. Rather than having a `createAuthProvider` boolean field which is set to `false` after creation, the adequate setting should be `wantsAuthProvider` and should trigger the creation of an auth provider _if one does not yet exist_ (and if one exists with the wrong settings, fixes those).

I have no idea what would happen in the current code if the reconciler crashes and comes back up. Would it just try to create another auth provider with the same name and then fail? Would it always be stuck in `createAuthProvider == true`? In the current code, that might mean that while we set up the auth provider initially and then disable password generation, after a crash we would enable password generation again and it would remain enabled because we can't proceed past `CreateRhSso` due to creation failures.

dhaus67:
* main
  * **PR #265** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/265" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/265?utm_source=stack-comment).

misberner (changes requested):

fleetshard/pkg/central/client/client.go
```diff
@@ -37,6 +38,18 @@ type AuthProviderResponse struct {
 	ID string `json:"id"`
 }
 
+// GetLoginAuthProviderResponse ...
```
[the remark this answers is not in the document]
dhaus67: Don't see why we wouldn't do this to be honest. I've changed it to use the respective proto definitions.

fleetshard/pkg/central/client/client.go
```diff
@@ -102,3 +140,26 @@ func (c *Client) SendAuthProviderRequest(ctx context.Context, authProviderReques
...
+	if err != nil {
+		return nil, errors.Wrap(err, "sending GetLoginAuthProviders request to central")
+	} else if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		return nil, acsErrors.NewErrorFromHTTPStatusCode(resp.StatusCode, "failed to GetLoginAuthProviders from central %s/%s", c.central.Metadata.Namespace, c.central.Metadata.Name)
+	}
+
+	defer func() {
+		err := resp.Body.Close()
+		if err != nil {
+			glog.Warningf("Attempt to close GetLoginAuthProviders response failed: %s", err)
```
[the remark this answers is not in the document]
dhaus67: Used utils.IgnoreError (seems to be in line what we typically do within this code base / stackrox).

fleetshard/pkg/central/reconciler/init_auth.go
```diff
@@ -53,6 +57,50 @@ var (
...
+func isCentralReady(ctx context.Context, client ctrlClient.Client, central private.ManagedCentral) (bool, error) {
+	deployment := &appsv1.Deployment{}
+	err := client.Get(ctx,
+		ctrlClient.ObjectKey{Name: "central", Namespace: central.Metadata.Namespace},
+		deployment)
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("retrieving central deployment resource from Kubernetes: %w", err)
```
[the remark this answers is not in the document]
dhaus67: Looking at other code, I'd say we use `errors.Wrap`, refactored it to use that.

fleetshard/pkg/central/client/client.go
```diff
@@ -49,56 +48,122 @@ func NewCentralClient(central private.ManagedCentral, address, pass string) *Cli
...
 }
 
-// SendGroupRequest ...
+func (c *Client) createRequest(ctx context.Context, requestMessage proto.Message, method, path string) (*http.Request, error) {
+	var body io.Reader
+	if requestMessage != nil {
+		marshaller := jsonpb.Marshaler{
+			EmitDefaults: true,
+			Indent:       "  ",
+		}
```
misberner, line 84:
```suggestion
		marshaller := jsonpb.Marshaler{}
```
we're talking to a machine, no need for redundant information and pretty formatting

fleetshard/pkg/central/client/client.go
```diff
@@ -49,56 +48,122 @@ func NewCentralClient(central private.ManagedCentral, address, pass string) *Cli
...
+		jsonString, err := marshaller.MarshalToString(requestMessage)
+		if err != nil {
+			return nil, errors.Wrap(err, "marshalling new request to central")
+		}
+		body = strings.NewReader(jsonString)
```
misberner, line 89:
```suggestion
		jsonBytes, err := marshaller.Marshal(requestMessage)
		if err != nil {
			return nil, errors.Wrap(err, "marshalling new request to central")
		}
		body = bytes.NewReader(jsonBytes)
```
no need for string here

fleetshard/pkg/central/client/client.go
```diff
@@ -49,56 +48,122 @@ func NewCentralClient(central private.ManagedCentral, address, pass string) *Cli
...
+		defer utils.IgnoreError(resp.Body.Close)
+	}
+
+	if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		return nil, acsErrors.NewErrorFromHTTPStatusCode(resp.StatusCode,
+			"failed to create auth provider for central %s/%s",
+			c.central.Metadata.Namespace, c.central.Metadata.Name)
+	}
+
+	authProvider := &storage.AuthProvider{}
```
misberner, line 138: Please factor this out by:
1. rename `SendRequestToCentral` to `SendRequestToCentralRaw` (it might be useful eventually)
2. (re-)introduce `SendRequestToCentral` with an additional `responseOut proto.Message` parameter and returning a single `error` only. This method will call `SendRequestToCentralRaw`, and perform everything that this method does, only that it will not allocate a message in the highlighted line, but just use `responseOut`. If `responseOut` is `nil`, no unmarshaling should be performed (see below).

`SendAuthProviderRequest` then becomes:
```
func (c *Client) SendAuthProviderRequest(ctx context.Context, authProviderRequest *storage.AuthProvider) (*storage.AuthProvider, error) {
	var authProviderResponse storage.AuthProvider
	err := c.SendRequestToCentral(ctx, authProviderRequest, http.MethodPost, "/v1/authProviders", &authProviderResponse)
	if err != nil {
		return nil, err
	}
	return &authProviderResponse, nil
}
```

Gracefully handling `nil` `responseOut` arguments frees you from having to deal with closing response bodies for cases where you are not interested in them. For example:

```
func (c *Client) SendGroupRequest(ctx context.Context, groupRequest *storage.Group) error {
	resp, err := c.SendRequestToCentral(ctx, groupRequest, http.MethodPost, "/v1/groups", nil)
	if err != nil {
		return errors.Wrapf(err, "failed to create group for central %s/%s",
			c.central.Metadata.Namespace, c.central.Metadata.Name)
	}

	return nil
}
```

misberner (changes requested):

fleetshard/pkg/central/client/client.go
```diff
@@ -49,56 +47,115 @@ func NewCentralClient(central private.ManagedCentral, address, pass string) *Cli
...
+	responseMessage proto.Message) error {
+	resp, err := c.SendRequestToCentralRaw(ctx, requestMessage, method, path)
 	if err != nil {
-		return errors.Wrap(err, "sending new group to central")
+		return err
+	}
+
+	// Only required to close the response body if we plan on unmarshalling it. If no response message is set, we won't
+	// unmarshal and can hence skip closing the body.
+	if responseMessage != nil && resp.Body != nil {
```
misberner, line 90: I have no idea why my pending comment doesn't show up in review, but just in case it doesn't show up after submitting, please see my other comment regarding response body. This must always be closed.

misberner (approved):
Please fix the response body leak before merging

fleetshard/pkg/central/client/client.go
misberner: OK GItHub just dropped the pending comment altogether -.- response bodies must always be closed because otherwise we leak resources. And they are always non-`nil` if `err == nil`

dhaus67: Resolving in favor of the above comment.

fleetshard/pkg/central/client/client.go
```diff
@@ -49,56 +47,115 @@ func NewCentralClient(central private.ManagedCentral, address, pass string) *Cli
...
+// non-successful HTTP status code.
+func (c *Client) SendRequestToCentral(ctx context.Context, requestMessage proto.Message, method, path string,
+	responseMessage proto.Message) error {
+	resp, err := c.SendRequestToCentralRaw(ctx, requestMessage, method, path)
 	if err != nil {
-		return errors.Wrap(err, "sending new group to central")
+		return err
+	}
+
+	// Only required to close the response body if we plan on unmarshalling it. If no response message is set, we won't
```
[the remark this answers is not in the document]
dhaus67: Fixed this.

misberner (approved):

PR #267, opened 2022-08-04
ROX-12072: Route update optimizations

kovayur:
## Description

This change adds a few optimizations dedicated to reduce a number of status reports back to the fleet manager. The changes are:
1. Skip sending an update when central is **ready** and not changed (revert to the previous state)
2. Skip sending an update when central is **installing** and not changed.
3. Do not return and print errors when central is not ready (secrets, deployments) not created.
4. Return "skippable" error instead of `nil` to indicate that deletion is in progress
5. Minor fixes to cache and pointer reference

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -222,12 +222,6 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 		r.hasAuthProvider = true
 	}
 
-	// Setting the last central hash must always be executed as the last step.
-	// defer can't be used for this call because it is also executed after the reconcile failed.
-	if err := r.setLastCentralHash(remoteCentral); err != nil {
-		return nil, errors.Wrapf(err, "setting central reconcilation cache")
-	}
-
 	// TODO(create-ticket): When should we create failed conditions for the reconciler?
 	status := readyStatus()
 	// Do not report routes statuses if:
@@ -240,8 +234,8 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 		}
 	}
 
-	// Cache MUST be updated right before returning "ready" result.
-	// Otherwise, the hash will be saved without sending the actual status (in case of error).
+	// Setting the last central hash must always be executed as the last step.
+	// defer can't be used for this call because it is also executed after the reconcile failed.
 	if err := r.setLastCentralHash(remoteCentral); err != nil {
 		return nil, errors.Wrapf(err, "setting central reconcilation cache")
 	}
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -52,7 +52,7 @@ type CentralReconciler struct {
 func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private.ManagedCentral) (*private.DataPlaneCentralStatus, error) {
 	// Only allow to start reconcile function once
 	if !atomic.CompareAndSwapInt32(r.status, FreeStatus, BlockedStatus) {
-		return nil, errors.New("Reconciler still busy, skipping reconciliation attempt.")
+		return nil, ErrBusy
 	}
 	defer atomic.StoreInt32(r.status, FreeStatus)
 
@@ -63,10 +63,10 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 
 	remoteCentralName := remoteCentral.Metadata.Name
 	remoteCentralNamespace := remoteCentral.Metadata.Namespace
+	remoteCentralReady := remoteCentral.RequestStatus == centralConstants.DinosaurRequestStatusReady.String()
 
-	if !changed && r.wantsAuthProvider == r.hasAuthProvider && remoteCentral.RequestStatus == centralConstants.DinosaurRequestStatusReady.String() {
-		glog.Infof("Central %s/%s not changed, skipping reconciliation", remoteCentralNamespace, remoteCentralName)
-		return r.readyStatus(ctx, remoteCentralNamespace)
+	if !changed && r.wantsAuthProvider == r.hasAuthProvider && remoteCentralReady {
+		return nil, ErrCentralNotChanged
 	}
 
 	monitoringExposeEndpointEnabled := v1alpha1.ExposeEndpointEnabled
@@ -148,7 +148,7 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 		if deleted {
 			return deletedStatus(), nil
 		}
-		return nil, nil
+		return nil, ErrDeletionInProgress
 	}
 
 	if err := r.ensureNamespaceExists(remoteCentralNamespace); err != nil {
@@ -191,6 +191,9 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 
 	if r.useRoutes {
 		if err := r.ensureRoutesExist(ctx, remoteCentral); err != nil {
+			if errors.Is(err, k8s.ErrCentralTLSSecretNotFound) {
+				return installingStatus(), nil
+			}
 			return nil, errors.Wrapf(err, "updating routes")
 		}
 	}
@@ -208,7 +211,7 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 	// 1. Auth provider is already created
 	// 2. OR reconciler creator specified auth provider not to be created
 	// 3. OR Central request is in status "Ready" - meaning auth provider should've been initialised earlier
-	if r.wantsAuthProvider && !r.hasAuthProvider && remoteCentral.RequestStatus != centralConstants.DinosaurRequestStatusReady.String() {
+	if r.wantsAuthProvider && !r.hasAuthProvider && !remoteCentralReady {
 		err = createRHSSOAuthProvider(ctx, remoteCentral, r.client)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


kovayur:
* main
  * **PR #267** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/267" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/267?utm_source=stack-comment).

mtesseract:
@kovayur , can you provide a brief description for this PR?

kovayur:
> @kovayur , can you provide a brief description for this PR?
@mtesseract thanks for pointing this out. I forgot to put a description, fixed now.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -186,30 +186,35 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
-		return nil, errors.Wrapf(err, "setting central reconcilation cache")
-	}
-
 	// Check whether deployment is ready.
 	centralReady, err := isCentralReady(ctx, r.client, remoteCentral)
 	if err != nil {
 		return nil, err
 	}
-	if !centralReady {
+	if !centralReady || !centralTLSSecretFound {
```
mtesseract, line 205: Just to confirm, is it possible that `centralTLSSecretFound == false` and `centralReady == true` hold at the same time?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -218,18 +223,23 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
-}
-
-func (r *CentralReconciler) readyStatus(ctx context.Context, namespace string) (*private.DataPlaneCentralStatus, error) {
 	status := readyStatus()
-	if r.useRoutes {
-		routesStatuses, err := r.getRoutesStatuses(ctx, namespace)
+	// Do not report routes statuses if:
+	// 1. Routes are not used on the cluster
+	// 2. Central request is in status "Ready" - assuming that routes are already reported and saved
+	if r.useRoutes && !remoteCentralReady {
```
mtesseract, line 230: We are not able (currently) to correctly handle the case where `r.useRoutes` is modified, right?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -218,18 +223,23 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
 		if err != nil {
 			return nil, err
 		}
-		status.Routes = routesStatuses
 	}
+
+	// Cache MUST be updated right before returning "ready" result.
+	// Otherwise, the hash will be saved without sending the actual status (in case of error).
+	if err := r.setLastCentralHash(remoteCentral); err != nil {
+		return nil, errors.Wrapf(err, "setting central reconcilation cache")
```
mtesseract, line 240: nit: adding the cluster ID?
```suggestion
		return nil, errors.Wrapf(err, "updating reconcilation cache for cluster %s", remoteCentral.Id)
```

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -148,15 +148,8 @@ func TestReconcileLastHashSetOnSuccess(t *testing.T) {
...
 
-	status, err := r.Reconcile(context.TODO(), managedCentral)
-	require.NoError(t, err)
-	assert.Equal(t, "Ready", status.Conditions[0].Type)
-	assert.Equal(t, "True", status.Conditions[0].Status)
-
-	central := &v1alpha1.Central{}
-	err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: centralName, Namespace: centralNamespace}, central)
-	require.NoError(t, err)
-	assert.Equal(t, "4", central.Annotations[revisionAnnotationKey])
```
mtesseract, line 159: We can keep the check for the revision being `4`, no?

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -198,7 +191,7 @@ func TestReconcileDelete(t *testing.T) {
 
 	// trigger deletion
 	statusTrigger, err := r.Reconcile(context.TODO(), deletedCentral)
-	require.NoError(t, err)
```
mtesseract, line 201: Wondering if this could potentially lead to a flake in case the central is completely deleted unexpectedly fast?

fleetshard/pkg/runtime/runtime.go
```diff
@@ -107,7 +107,11 @@ func (r *Runtime) Start() error {
 
 func (r *Runtime) handleReconcileResult(central private.ManagedCentral, status *private.DataPlaneCentralStatus, err error) {
 	if err != nil {
-		glog.Errorf("error occurred %s/%s: %s", central.Metadata.Namespace, central.Metadata.Name, err.Error())
+		if centralReconciler.IsSkippable(err) {
+			glog.Infof("Skip sending the status for central %s/%s: %v", central.Metadata.Namespace, central.Metadata.Name, err)
+		} else {
+			glog.Errorf("Unexpected error occurred %s/%s: %s", central.Metadata.Namespace, central.Metadata.Name, err.Error())
```
mtesseract, line 113:
```suggestion
			glog.Errorf("Unexpected error occurred for central %s/%s: %v", central.Metadata.Namespace, central.Metadata.Name, err.Error())
```

mtesseract (commented):
Good job! just a few remarks.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -63,10 +63,10 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 
 	remoteCentralName := remoteCentral.Metadata.Name
 	remoteCentralNamespace := remoteCentral.Metadata.Namespace
+	remoteCentralReady := remoteCentral.RequestStatus == centralConstants.DinosaurRequestStatusReady.String()
```
SimonBaeumer, line 66: Can you add to function to the managed central struct `IsReady`?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -186,30 +186,35 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
+			}
 		}
 	}
 
-	if err := r.setLastCentralHash(remoteCentral); err != nil {
-		return nil, errors.Wrapf(err, "setting central reconcilation cache")
-	}
-
 	// Check whether deployment is ready.
 	centralReady, err := isCentralReady(ctx, r.client, remoteCentral)
```
SimonBaeumer, line 201: Can you rename this function to `isCentralDeploymentReady`? This function was not introduced in your pull request but could cause confiusion with the variable name `remoteCentralReady`.

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
mtesseract: This is generated from OpenAPI, not sure what the best approach is here.

mtesseract: We should look at the pre-commit configuration in that case. The working assumption there is that everything in those folders is auto-generated. Not sure if this assumption is also used in other places.

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: This is already done by caller (runtime): 
```
			glog.Errorf("Unexpected error occurred for central %s/%s: %v", central.Metadata.Namespace, central.Metadata.Name, err.Error())
```

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: Not sure I understand the question. `r.useRoutes` is not expected to change at runtime.

fleetshard/pkg/central/reconciler/reconciler_test.go
kovayur: You are right. Reverted.

fleetshard/pkg/central/reconciler/reconciler_test.go
kovayur: Good point. I think this is currently not possible because we check the existence of the object on a next reconcile attempt.

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: Yes, it is possible. I'm going to rename `centralReady` => `centralDeploymentReady` to avoid confusion

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: Done

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: Yes, this makes sense. Done.

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: Agree with @mtesseract, adding a new file to the autogenerated package may cause difficulties. For example, the package is [wiped on generation](https://github.com/stackrox/acs-fleet-manager/blob/main/Makefile#L396).
I extracted functions in the same file for now.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -63,10 +63,10 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 
 	remoteCentralName := remoteCentral.Metadata.Name
 	remoteCentralNamespace := remoteCentral.Metadata.Namespace
+	remoteCentralReady := isRemoteCentralReady(remoteCentral)
 
-	if !changed && r.wantsAuthProvider == r.hasAuthProvider && remoteCentral.RequestStatus == centralConstants.DinosaurRequestStatusReady.String() {
-		glog.Infof("Central %s/%s not changed, skipping reconciliation", remoteCentralNamespace, remoteCentralName)
-		return r.readyStatus(ctx, remoteCentralNamespace)
+	if !changed && r.wantsAuthProvider == r.hasAuthProvider && remoteCentralReady {
```
mtesseract, line 68: nit: I guess from a readability perspective it wouldn't be required to have a separate variable `remoteCentralReady`, since `isRemoteCentralReady(remoteCentral)` is easy to read as well and could be used inline.

fleetshard/pkg/central/reconciler/reconciler.go
mtesseract: Yeah, sorry, could have elaborated a little bit more. I was wondering what would happen if `useRoutes` is toggled and fleetshard-sync would be restarted. Probably not something that would occur in daily operations, but I was wondering from a data consistency perspective if in such a scenario routes (actual routes or empty routes) would need to be resent to fleet-manager.

mtesseract (approved):

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: I think it is part of a bigger problem I also raised before. At this moment we don't track stale resources on k8s cluster. 
I can imagine a case when fleetshard-sync was unavailable for some time and there were changes in the fleet manager, i.e. central was removed. The associated namespace with managed central will remain on the cluster even after fleetshard-sync becomes available.
I think we should track and alert such "stale" resources. Metrics may help us with that.

fleetshard/pkg/central/reconciler/reconciler.go
kovayur: done

SimonBaeumer (approved):

PR #268, opened 2022-08-04
ROX-12046: Add logging helm chart for data plane logging stack

mtodor:
## Description

This PR adds a logging stack to the terraforming helm template for the data plane.

**Notes**
- We have added post-install/update hooks to avoid the problem of missing custom resource definitions. The problem here is that custom resource definitions are installed after the Subscription is applied for the logging operator. Usually, it takes some time for it to be applied. And that can cause problems that custom resource definition does not exist when the helm is creating custom resource. With helm post hook, custom resources will be created after everything else is successfully deployed. *still to be properly tested*
- All logging stack is deployed into the `openshift-logging` namespace. This is a namespace defined for it.
- Some resources have predefined names. i.e. `instance` - please check docs: https://docs.openshift.com/container-platform/4.10/logging/cluster-logging-external.html#cluster-logging-collector-log-forward-cloudwatch_cluster-logging-external

## Test manual

The best way to test it is to use **THE GUIDE** to create a staging OSD cluster: `docs/development/setup-developer-osd-cluster.md`
The only important difference is in step 12. where we should add additional options for logging stack.

Required access keys can be found in BitWarden: `CloudWatch - Development Environment Access Key`

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/README.md
+++ b/dp-terraform/helm/rhacs-terraform/README.md
@@ -45,4 +45,4 @@ helm upgrade --install rhacs-terraform \
 helm uninstall rhacs-terraform --namespace rhacs
 ```
 
-See internal wiki for an example file `~/.rh/terraform-values.yaml`.
+See internal wiki for an example file `~/acs-terraform-values.yaml`.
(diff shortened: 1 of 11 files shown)


mtodor:
* main
  * **PR #268** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/268" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/268?utm_source=stack-comment).

stehessel (approved):

PR #277, opened 2022-08-05
Add SOP

kurlov:
## Description

Add runbook to the service. Alerts will be linked to this SOP


kurlov:
No longer relevant. Moved to the separate repo

PR #286, opened 2022-08-11
ROX-12052: Add script commands for quickly adding/removing hostnames locally

kovayur:
## Description
This change simplifies routes configuration on vanilla k8s. With new script commands you can easily and interactivally add and remove exposed hosts locally  

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/scripts/ingress-router.sh
+++ b/scripts/ingress-router.sh
@@ -8,13 +8,19 @@ ROUTER_RBAC_YAML="https://raw.githubusercontent.com/openshift/router/master/depl
 ROUTER_CRD_YAML="https://raw.githubusercontent.com/openshift/api/master/route/v1/route.crd.yaml"
 ROUTER_DEPLOYMENT_YAML="https://raw.githubusercontent.com/openshift/router/master/deploy/router.yaml"
 APPS_OPENSHIFT_CRD_YAML="$SCRIPT_DIR/../dev/env/manifests/ingress-router/dummy.crd.yaml"
+HOSTCTL_PROFILE="acs"
 
 usage() {
   echo "Usage: $(basename "$0") [-h | --help] <command> [<args>]"
   echo ""
   echo "Available commands:"
-  echo "  deploy               Deploys ingress router on an active k8s cluster in kubectl context"
-  echo "  undeploy             Un-deploys ingress router on an active k8s cluster in kubectl context"
+  echo "  deploy                                                          Deploys ingress router on an active k8s cluster in kubectl context (requires kubectl)"
+  echo "  undeploy                                                        Un-deploys ingress router on an active k8s cluster in kubectl context (requires kubectl)"
+  echo "  host                                                            Helper commands for exposing Routes locally"
+  echo "                                                                  See more: https://github.com/stackrox/acs-fleet-manager/blob/main/docs/development/test-locally-route-hosts.md"
+  echo "    host add (--profile=<profile> | -p <profile>) <hostname>      Adds the selected hostname to /etc/hosts (requires: kubectl, jq, hostctl, fzf). Profile: hostctl profile. Default: 'acs'"
+  echo "    host remove (--profile=<profile> | -p <profile>) <hostname>   Removes the selected hostname from /etc/host (requires: kubectl, jq, hostctl, fzf). Profile: hostctl profile. Default: 'acs'"
+
   if [[ -n "${1:-}" ]]; then
     echo ""
     echo >&2 "Error: $1"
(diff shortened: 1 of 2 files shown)


kovayur:
* main
  * **PR #286** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/286" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/286?utm_source=stack-comment).

mtesseract:
Do I understand this right that the modification of some host name data base (e.g. `/etc/hosts`) is required for getting the e2e test suite back working on vanilla k8s?

I must admit that I don't really like the idea of having to update some system configuration files for running e2e tests.

Are there alternatives to this approach? For example, would it be possible to spin up a pod with coredns or some other DNS server that would provide discoverability for routes?
@kovayur

vladbologa:
> Do I understand this right that the modification of some host name data base (e.g. `/etc/hosts`) is required for getting the e2e test suite back working on vanilla k8s?
> 
> I must admit that I don't really like the idea of having to update some system configuration files for running e2e tests.
> 
> Are there alternatives to this approach? For example, would it be possible to spin up a pod with coredns or some other DNS server that would provide discoverability for routes? @kovayur

I agree, but I think we can live with this for now, at least until the service preview. Would opening a ticket and fixing it later be acceptable @mtesseract ?

scripts/ingress-router.sh
```diff
@@ -43,13 +49,35 @@ list_manifests_reversed() {
   list_manifests | tac
 }
 
+add_host() {
+  local host context
+  context=$(kubectl config current-context)
```
mtesseract, line 54: Suggestion: Let's do something like `KUBECTL=${KUBECTL:-kubectl}` and use that instead of `kubectl` directly.
Reason: Happens regularly for me that I have to juggle different kubectl versions for accessing a specific K8s cluster and doing `export KUBECTL=...` seems to be an established way of supporting this use-case.

scripts/ingress-router.sh
```diff
@@ -43,13 +49,35 @@ list_manifests_reversed() {
...
+  host=$(kubectl get routes -l 'app.kubernetes.io/managed-by=rhacs-fleetshard' --all-namespaces -o json | jq -r '.items[].spec.host' | fzf --header "Using context $context")
```
mtesseract, line 55: In case this `kubectl get` command fails, e.g. on my system right now:
```
error: the server doesn't have a resource type "routes"
```

the output of `fzf` is a bit mangled (due to the attempted printing to stderr). I guess it would be acceptable to simply use `set -eo pipefail` for now so that the whole command will fail if the `kubectl get` fails instead of showing a mangled/dysfunctional `fzf` prompt.

scripts/ingress-router.sh
```diff
@@ -43,13 +49,35 @@ list_manifests_reversed() {
...
+  sudo hostctl add domains "$HOSTCTL_PROFILE" "$host"
+}
+
+remove_host() {
+  local host
+  host=$(hostctl list "$HOSTCTL_PROFILE" -o json | jq -r ".[].Host" | fzf)
```
mtesseract, line 61:
```suggestion
  host=$(hostctl list "$HOSTCTL_PROFILE" -o json | jq -r ".[].Host" | fzf --header "Using context $context")
```

Or something similar?

scripts/ingress-router.sh
```diff
@@ -72,6 +100,21 @@ deploy)
 undeploy)
   undeploy
   ;;
+host)
+  subcommand=$2
+  host=$3
+  case $subcommand in
+  add)
+    add_host "$host"
```
mtesseract, line 108: This argument `$host` does not seem to be used?
Same for the `remove_host` below.

Looking at the usage info
```
  echo "    host add (--profile=<profile> | -p <profile>) <hostname>      Adds the selected hostname to /etc/hosts (requires: kubectl, jq, hostctl, fzf). Profile: hostctl profile. Default: 'acs'"
  echo "    host remove (--profile=<profile> | -p <profile>) <hostname>   Removes the selected hostname from /etc/host (requires: kubectl, jq, hostctl, fzf). Profile: hostctl profile. Default: 'acs'"
```

it seems like this subcommand should also allow a non-interactive style (without fzf?), but it looks like it doesn't?

kovayur:
> Do I understand this right that the modification of some host name data base (e.g. `/etc/hosts`) is required for getting the e2e test suite back working on vanilla k8s?
> 
> I must admit that I don't really like the idea of having to update some system configuration files for running e2e tests.
> 
> Are there alternatives to this approach? For example, would it be possible to spin up a pod with coredns or some other DNS server that would provide discoverability for routes? @kovayur

First of all, this script is currently NOT required to run E2E tests on vanilla k8s unless we want to access central in E2E tests. It is mainly for manual testing to make sure that the newly created central is accessible. Without this script, routes can still be created, but you won't be able to resolve central host and open it in a browser.

I agree, DNS solution looks more mature, however there a few arguments for using a given script:
1. The DNS server requires resources, and one of the main reasons I switched from CRC is to minimize the resources required for running managed services locally.
2. Adding a DNS server is also a system configuration update. Of course, this should be done once, but I don't expect that this script will be used often.
3. Looks like the same approach is used in CRC, at least I see the records in my `/etc/hosts` from there

scripts/ingress-router.sh
kovayur: This does not call `kubectl` just prints the list from `/etc/hosts`. I could add a profile here.

scripts/ingress-router.sh
kovayur: You are right. Fixed.

scripts/ingress-router.sh
kovayur: I see, but there is already `set -eo pipefail`. I don't know why it doesn't work as expected, but I splitted the commands, this should help.

scripts/ingress-router.sh
kovayur: Done

mtesseract (approved):

PR #288, opened 2022-08-15
Add missing openapi endpoint & prevent formatting of generated files

mtesseract:
## Description

```

    Add previously forgotten POST endpoint to the admin API, which is already implemented in code.
    Regenerate api files.
    Fix pre-commit configuration to not modify generated files.
    Exclude two specific secrets which end up in the generated files (unfortunately the previously added #pragma annotation did not carry over to the generated files).
    Run pre-commit run --all-files
```

## Test manual

```
# To run tests locally run:
```


internal/dinosaur/pkg/api/public/api/openapi.yaml
```diff
@@ -815,7 +815,7 @@ components:
       value:
         id: 1iSY6RQ3JKI8Q0OTmjQFd3ocFRg
         kind: central
-        href: /api/rhacs/v1/centrals/1iSY6RQ3JKI8Q0OTmjQFd3ocFRg # pragma: allowlist secret
```
kovayur, line 818: Why don't you want to keep comments here?

.pre-commit-config.yaml
```diff
@@ -13,7 +13,13 @@ repos:
     rev: v1.2.0
     hooks:
       - id: detect-secrets
-        args: ["--baseline", ".secrets.baseline"]
+        args:
+          - --baseline
+          - .secrets.baseline
+          - --exclude-secrets
```
kovayur, line 19: 1) I was trying to avoid placing exclusions here because sometimes we may want to use `detect-secrets` out of `pre-commit`. Adding more arguments makes it harder to use `detect-secrets`. For example, I used it to recreate a baseline from scratch. If you think that's not the case - I'm fine with that.
2) I think the exclusion added here is too specific. I can imagine that with such level of details the list of exclusion here may drastically grow. Also this increases the number of locations where we should track the exclusions. I'd keep only exclusion paths here as you did for the `trailing-whitespace`.

To conclude: If we don't want to have comments in yaml files, I think it's ok to keep it in the `.secrets.baseline`

.pre-commit-config.yaml
```diff
@@ -13,7 +13,13 @@ repos:
     rev: v1.2.0
     hooks:
       - id: detect-secrets
-        args: ["--baseline", ".secrets.baseline"]
+        args:
+          - --baseline
```
kovayur, line 17: nit: I think the quotes here will increase readability.

internal/dinosaur/pkg/api/public/api/openapi.yaml
kovayur: Ah, I see. Can we exclude the whole generated path ?

mtesseract: That is also possible, yes. The reasons for me doing it this way are the following:

1. If we exclude by path then it is (theoretically!) possible to add an actual secret in an excluded file and commit it without `detect-secrets` being able to catch it. If we only allow-list specific (non-secret) entropy strings, then this is prevented.

2. I do not really anticipate that we will need to continuously maintain the list of exclusions. This would only be required in situations where we introduce new high-entropy strings (or something similar which is caught be detect-secret's built-in heuristics) in the OpenAPI source YAMLs.

3. This approach does not rely on the secrets baseline, which does not seem to be a good fit for generated files whose line numbers could change when working on API changes.

.pre-commit-config.yaml
mtesseract: Just to be clear: I think the best solution (by far) would be if YAML comments would be carried over by the openapi generator! What I am doing here is working around this apparent limitation.

Regarding your first point: I haven't thought about this use-case. Good point.

Regarding your second point: Again, to be clear, what is presented here in this PR is just my workaround for dealing with generated (OpenAPI) files. I am not suggesting to use this exclude mechanism on a bigger scale -- in the vast majority of situations we can simply use pragma comments.

mtesseract:
@kovayur Thanks to your concerns I've been able to simplifiy this. Apparently detect-secrets ignores uuids. Not entirely sure if that is a good thing or a bad thing. :-/ However, I have replaced the high-entropy strings in the OpenAPI YAML with UUIDs and removed the secret-specific excludes.

internal/dinosaur/pkg/api/public/api/openapi.yaml
kovayur: Yes, you are right, we may accidentally commit to the generated file and it will not be tracked by the pre-commit hook.

kovayur (approved):
LGTM

internal/dinosaur/pkg/api/public/api/openapi.yaml
vladbologa: But don't we have a separate hook that checks that the generated files are up-to-date?

PR #291, opened 2022-08-15
ROX-12074: Implement endpoint for updating central resource settings

mtesseract:
## Description

Implement support for  updating Central resource configuration in the`dinosaur` PATCH endpoint in the Admin API (will be renamed to `central` in a follow-up PR).

I found it too cumbersome to convert between the existing API model for resource requirements and the upstream one from `corev1`. The former used explicit fields for `Cpu` and `Memory` while the latter one used a generic map under the hood. It turned out to be tricky to get the right semantics for serialization/deserialization in the presence of zero-values, hence I streamlined this so that the API models for the resource requirements now also use a generic map.

- [x] ~~Documentation added if necessary~~

## Test manual

```
# To run tests locally run:
```


internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -48,14 +49,15 @@ func (h adminDinosaurHandler) Create(w http.ResponseWriter, r *http.Request) {
...
-			ValidateScannerSpec(ctx, &dinosaurRequest, "scanner", convDinosaur),
+			ValidateCentralSpec(ctx, &dinosaurRequest, convDinosaur),
+			ValidateScannerSpec(ctx, &dinosaurRequest, convDinosaur),
 		},
 		Action: func() (interface{}, *errors.ServiceError) {
 			svcErr := h.service.RegisterDinosaurJob(convDinosaur)
 			if svcErr != nil {
 				return nil, svcErr
 			}
+			// TODO(mclasmeier): Do we need PresentDinosaurRequestAdminEndpoint?
```
johannes94, line 60: Good question, I was about to comment something similar. I think we should have a dedicated type as input to the AdminCreate Endpoint and also probably return a dedicated response here. One reason is that the whole code base is separated into this different types for different APIs and it felt unusual that the new admin endpoints reuse types of other parts than the admin API. Second reason is that if we don't want to expose specific admin fields to the public API we would have to change it anyways. 3rd is if we change something for admin API, we also change the public API and we don't know for sure if tools or GUIs connecting to that API can handle the change without being modified.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -141,7 +143,59 @@ func (h adminDinosaurHandler) Delete(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return fmt.Errorf("updating CentralSpec within CentralRequest: %w", err)
+	}
+
+	err = new.SetScannerSpec(scannerSpec)
+	if err != nil {
+		return fmt.Errorf("updating ScannerSpec within CentralRequest: %w", err)
+	}
+
+	// Disabled this for now, since it is unclear as of now what our specific requirements
```
johannes94, line 181: Nit: I like the YAGNI approach https://martinfowler.com/bliki/Yagni.html#:~:text=Yagni%20originally%20is%20an%20acronym,'t%20gonna%20need%20it%22. So I would just remove it, since we have it in the git history anyways if we really need it.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -609,49 +609,67 @@ func (k *dinosaurService) Updates(dinosaurRequest *dbapi.CentralRequest, fields
 	return nil
 }
 
+// Disabled this for now.
```
johannes94, line 612: Same for this.

johannes94 (changes requested):
Good Job, left a few comments. One thing I encountered when executing e2e tests locally was that we need to set STATIC_TOKEN_ADMIN before running them. Would be nice if you could add it to `howto-run-e2e-tests.md`

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -48,14 +49,15 @@ func (h adminDinosaurHandler) Create(w http.ResponseWriter, r *http.Request) {
 			ValidateDinosaurClaims(ctx, &dinosaurRequest, convDinosaur),
 			ValidateCloudProvider(&h.service, convDinosaur, h.providerConfig, "creating central requests"),
 			handlers.ValidateMultiAZEnabled(&dinosaurRequest.MultiAz, "creating central requests"),
-			ValidateCentralSpec(ctx, &dinosaurRequest, "central", convDinosaur),
-			ValidateScannerSpec(ctx, &dinosaurRequest, "scanner", convDinosaur),
+			ValidateCentralSpec(ctx, &dinosaurRequest, convDinosaur),
```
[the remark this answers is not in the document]
mtesseract: done, thanks

internal/dinosaur/pkg/services/dinosaur.go
mtesseract: done

internal/dinosaur/pkg/handlers/admin_dinosaur.go
mtesseract: removed it

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -141,7 +160,139 @@ func (h adminDinosaurHandler) Delete(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return fmt.Errorf("updating ScannerSpec within CentralRequest: %w", err)
+	}
+
+	*request = new
+	return nil
+}
+
+// Update a Central instance.
 func (h adminDinosaurHandler) Update(w http.ResponseWriter, r *http.Request) {
```
johannes94, line 296: This endpoint return "null" as a response with a status code 200. Although the update of resources requirements works, "null" as a response indicates to me that something went wrong. Probably the reason is that the presenter used is not implemented.

johannes94 (approved):
LGTM except the comment. IMO we can merge and fix that in a follow up.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
mtesseract: Will do in a follow-up, thanks!

PR #293, opened 2022-08-16
Fix reconcile cache

SimonBaeumer:
## Description

The reconcile cache should be set after all steps were successfully completed. Here it was set before the SSO provider was initiliazed.
This can lead to skipped reconciles if the SSO registration fails, the reconciler returns. On the next reconcile the Central CR is still the same and the SSO config is never created.
Fixable by restarting fleetshard-sync.


johannes94 (approved):
LGTM

kovayur:
Same change in #267 :)

kovayur (approved):

PR #294, opened 2022-08-16
TEMPORARY Add rollback command to drop stage RDS

kurlov:
## Description

This command will drop all schemas.
This is a temporary change to pristine stage RDS. Stage DP clusters were recreated so we should clean stage DB as well.

### DIFF
--- a/templates/service-template.yml
+++ b/templates/service-template.yml
@@ -979,6 +979,7 @@ objects:
             command:
             - /usr/local/bin/fleet-manager
             - migrate
+            - rollback-all
             - --db-host-file=/secrets/rds/db.host
             - --db-port-file=/secrets/rds/db.port
             - --db-user-file=/secrets/rds/db.user


templates/service-template.yml
```diff
@@ -979,6 +979,7 @@ objects:
             command:
             - /usr/local/bin/fleet-manager
             - migrate
+            - rollback-all
```
machi1990, line 982: I think `rolling forward` would have been a safety approach here.

i.e adding another migration that cleans up the DP clusters given the cluster ids (these can be obtained).

Another option that has been discussed for our case is to introduce some DP clusters management API (to import clusters, deprovision clusters) etc which might enable for the kind of operation we are targeting to do with the rollback.

kurlov: You are right but we want to pristine our stage RDS because it was heavily used for development and we would like to not have a migration for that period and start from scratch.

machi1990: ack.

ebensh (approved):

PR #295, opened 2022-08-16
ROX-11990: Split host in public API to UI and Data

kovayur:
## Description
This PR splits `host` field of `CentralRequest` in Public API to `uiHost` and `dataHost`.

## Test manual

E2E tests
```
# To run tests locally run:
```


kovayur:
* main
  * **PR #295** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/295" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/295?utm_source=stack-comment).

mtesseract:
@kovayur Given the e2e failure, this is not yet ready for review, I suppose?

e2e/e2e_test.go
```diff
@@ -384,15 +371,29 @@ func getHostedZone(central *public.CentralRequest) (*route53.HostedZone, error)
...
+		HostedZoneId:    rhacsZone.Id,
+		StartRecordName: &central.UiHost,
+	})
+	Expect(err).ToNot(HaveOccurred())
+	dataRecords, err := route53Client.ListResourceRecordSets(&route53.ListResourceRecordSetsInput{
+		HostedZoneId:    rhacsZone.Id,
+		StartRecordName: &central.DataHost,
+	})
+	Expect(err).ToNot(HaveOccurred())
+	return len(uiRecords.ResourceRecordSets) + len(dataRecords.ResourceRecordSets)
```
mtesseract, line 398: nit: As caller of this function we can not differentiate between, for example, the case where `len(uiRecords.ResourceRecordSets) == 1 &&  len(dataRecords.ResourceRecordSets) == 1` and `len(uiRecords.ResourceRecordSets) == 0 &&  len(dataRecords.ResourceRecordSets) == 2`.

Besides that: 👍

mtesseract (approved):

e2e/e2e_test.go
kovayur: You are right, reworked

internal/dinosaur/pkg/api/public/model_central_request.go
```diff
@@ -27,7 +27,8 @@ type CentralRequest struct {
 	Region       string    `json:"region,omitempty"`
 	Owner        string    `json:"owner,omitempty"`
 	Name         string    `json:"name,omitempty"`
-	Host         string    `json:"host,omitempty"`
+	UiHost       string    `json:"ui_host,omitempty"`
```
mtesseract, line 30: @sachaudh FYI: This will be merged soon.

e2e/e2e_test.go
```diff
@@ -68,8 +68,14 @@ var _ = Describe("Central", func() {
...
-		adminAuth, err := envtokenauth.CreateAuth("STATIC_TOKEN_ADMIN")
-		Expect(err).ToNot(HaveOccurred())
+		var adminAuth fleetmanager.Auth
+		if adminToken := os.Getenv("STATIC_TOKEN_ADMIN"); adminToken != "" {
+			adminAuth, err = envtokenauth.CreateAuth("STATIC_TOKEN_ADMIN")
+			Expect(err).ToNot(HaveOccurred())
+		} else {
+			adminAuth = auth
+		}
+
```
SimonBaeumer, line 78: I don't really understand this. Why was this change necessary?

internal/dinosaur/pkg/api/admin/private/model_central_request.go
```diff
@@ -27,7 +27,8 @@ type CentralRequest struct {
 	Region       string    `json:"region,omitempty"`
 	Owner        string    `json:"owner,omitempty"`
 	Name         string    `json:"name,omitempty"`
-	Host         string    `json:"host,omitempty"`
+	UiHost       string    `json:"ui_host,omitempty"`
+	DataHost     string    `json:"data_host,omitempty"`
```
SimonBaeumer, line 31: WDYTH about `SensorHost` instead? This would reduce the mapping overhead for this route

```suggestion
	SensorHost     string    `json:"data_host,omitempty"`
```

e2e/e2e_test.go
```diff
@@ -165,25 +170,21 @@ var _ = Describe("Central", func() {
...
-			name := removeLastChar(*recordSet.Name)
-			Expect(name).To(Equal(central.Host))
-			Expect(*record.Value).To(Equal(reencryptIngress.RouterCanonicalHostname))
-
+			centralDomainNames := getCentralDomainNamesSorted(central)
+			dnsRecordsLoader := newDNSRecordsLoader(rhacsZone, centralDomainNames)
+			Eventually(dnsRecordsLoader.loadDNSRecords).
+				WithTimeout(waitTimeout).
+				WithPolling(defaultPolling).
+				Should(HaveLen(len(centralDomainNames)), "Started at %s", time.Now())
```
SimonBaeumer, line 178: Is here asserted that `len > 0`? If possible an assertion on the specific expected length would be great.

e2e/e2e_test.go
```diff
@@ -165,25 +170,21 @@ var _ = Describe("Central", func() {
...
-			recordSet := records.ResourceRecordSets[0]
-			Expect(len(recordSet.ResourceRecords)).To(Equal(1))
-			record := recordSet.ResourceRecords[0]
-
-			// Omit the . at the end of hosted zone name
-			name := removeLastChar(*recordSet.Name)
-			Expect(name).To(Equal(central.Host))
-			Expect(*record.Value).To(Equal(reencryptIngress.RouterCanonicalHostname))
-
+			centralDomainNames := getCentralDomainNamesSorted(central)
```
SimonBaeumer, line 173: Nice test!

e2e/e2e_test.go
```diff
@@ -233,15 +233,12 @@ var _ = Describe("Central", func() {
...
-				})
-				Expect(err).ToNot(HaveOccurred())
-				return len(records.ResourceRecordSets)
-			}).WithTimeout(waitTimeout).WithPolling(defaultPolling).Should(Equal(0))
+			centralDomainNames := getCentralDomainNamesSorted(central)
+			dnsRecordsLoader := newDNSRecordsLoader(rhacsZone, centralDomainNames)
+			Eventually(dnsRecordsLoader.loadDNSRecords).
+				WithTimeout(waitTimeout).
+				WithPolling(defaultPolling).
+				Should(BeEmpty(), "Started at %s", time.Now())
```
SimonBaeumer, line 241: Why are you printing a "Start at %s" here?

e2e/e2e_test.go
```diff
@@ -428,15 +425,67 @@ func getHostedZone(central *public.CentralRequest) (*route53.HostedZone, error)
...
 	}
 
 	return rhacsZone, nil
 }
+
+func getCentralDomainNamesSorted(central *public.CentralRequest) []string {
+	centralUIDomain := central.UiHost + "."
+	centralDataDomain := central.DataHost + "."
+
+	if centralUIDomain > centralDataDomain {
```
SimonBaeumer, line 445: Please use strings.Sort is more idiomatic
https://gobyexample.com/sorting

e2e/e2e_test.go
```diff
@@ -428,15 +425,67 @@ func getHostedZone(central *public.CentralRequest) (*route53.HostedZone, error)
...
+		return []string{centralDataDomain, centralUIDomain}
+	}
+	return []string{centralUIDomain, centralDataDomain}
+}
+
+type dnsRecordsLoader struct {
```
SimonBaeumer, line 451: This struct + functions look to me not related to e2e testing but only called from them. Can you extract them into their own package?

e2e/e2e_test.go
```diff
@@ -428,15 +425,67 @@ func getHostedZone(central *public.CentralRequest) (*route53.HostedZone, error)
...
+}
+
+func (loader *dnsRecordsLoader) loadDNSRecords() []*route53.ResourceRecordSet {
+	idx := 0
+	loadingRecords := true
+	nextRecord := &loader.centralDomainNames[idx]
+	result := make([]*route53.ResourceRecordSet, 0, len(loader.centralDomainNames))
+
+loading:
+	for loadingRecords {
```
SimonBaeumer, line 471: Is this loop necessary because of `ListResourceRecordSets` only returns the first 300 record sets?

SimonBaeumer (changes requested):

e2e/e2e_test.go
kovayur: Just to reduce a number of env variables required to run e2e tests locally. Now it is required to set both `STATIC_TOKEN` and `STATIC_TOKEN_ADMIN`

e2e/e2e_test.go
kovayur: I added this to debug test timeout issues. This way I can understand when the test run and see what happened in container logs during this time. I think it will not harm if we leave it as is.

internal/dinosaur/pkg/api/admin/private/model_central_request.go
kovayur: I think it will cause a confusion because we're already using `dataHost` in a few places in the code, for example in a private API.
`data` term is also mentioned in the hostname (agreed in the ADR), and on the design presentation. So I'd like to keep is as is.

e2e/e2e_test.go
kovayur: No it is asserted that `len == 2`, where `2` is the length of `centralDomainNames`

e2e/e2e_test.go
kovayur: Done.

e2e/e2e_test.go
kovayur: Kudos to @johannes94 🎉

e2e/e2e_test.go
kovayur: Extracted.

e2e/e2e_test.go
kovayur: Correct.

e2e/e2e_test.go
```diff
@@ -68,8 +67,14 @@ var _ = Describe("Central", func() {
...
 		Expect(err).ToNot(HaveOccurred())
 
-		adminAuth, err := envtokenauth.CreateAuth("STATIC_TOKEN_ADMIN")
-		Expect(err).ToNot(HaveOccurred())
+		var adminAuth fleetmanager.Auth
+		if adminToken := os.Getenv("STATIC_TOKEN_ADMIN"); adminToken != "" {
+			adminAuth, err = envtokenauth.CreateAuth("STATIC_TOKEN_ADMIN")
+			Expect(err).ToNot(HaveOccurred())
+		} else {
+			adminAuth = auth
```
mtesseract, line 75: I don't know -- under which circumstances would this work? As far as I know tokens for the public and the admin API require at least different token issuers?

SimonBaeumer (approved):

e2e/dns/records_loader.go
```diff
@@ -0,0 +1,97 @@
...
+		CentralDomainNames: getCentralDomainNamesSorted(central),
+		rhacsZone:          rhacsZone,
+	}
+}
+
+// LoadDNSRecords loads DNS records from Route53
+func (loader *RecordsLoader) LoadDNSRecords() []*route53.ResourceRecordSet {
+	idx := 0
+	loadingRecords := true
+	nextRecord := &loader.CentralDomainNames[idx]
```
mtesseract, line 37: Maybe return empty value in case `len(CentralDomainNames)==0` or fail with an error?

e2e/dns/records_loader.go
```diff
@@ -0,0 +1,97 @@
...
+		output, err := loader.route53Client.ListResourceRecordSets(&route53.ListResourceRecordSetsInput{
+			HostedZoneId:    loader.rhacsZone.Id,
+			StartRecordName: nextRecord,
+		})
+		Expect(err).ToNot(HaveOccurred())
+
+		for _, recordSet := range output.ResourceRecordSets {
+			if *recordSet.Name == loader.CentralDomainNames[idx] {
+				result = append(result, recordSet)
+				idx = idx + 1
```
mtesseract, line 51:
```suggestion
				idx += 1
```

mtesseract (commented):
not through yet

e2e/dns/records_loader.go
```diff
@@ -0,0 +1,97 @@
...
+// LoadDNSRecords loads DNS records from Route53
+func (loader *RecordsLoader) LoadDNSRecords() []*route53.ResourceRecordSet {
+	idx := 0
+	loadingRecords := true
+	nextRecord := &loader.CentralDomainNames[idx]
+	result := make([]*route53.ResourceRecordSet, 0, len(loader.CentralDomainNames))
+
+loading:
+	for loadingRecords {
+		output, err := loader.route53Client.ListResourceRecordSets(&route53.ListResourceRecordSetsInput{
```
mtesseract, line 42: nit: from a `f(input)=output` perspective the naming makes sense, but in my opinion something like `result` or something like this would feel more natural.

e2e/dns/records_loader.go
```diff
@@ -0,0 +1,97 @@
...
+		nextRecord = output.NextRecordName
+	}
+	loader.LastResult = result
+	return result
+}
+
+func getHostedZone(route53Client *route53.Route53, central *public.CentralRequest) (*route53.HostedZone, error) {
+	hostedZones, err := route53Client.ListHostedZones(&route53.ListHostedZonesInput{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list hosted zones: %v", err)
```
mtesseract, line 67:
```suggestion
		return nil, fmt.Errorf("failed to list hosted zones: %w", err)
```

e2e/dns/records_loader.go
```diff
@@ -0,0 +1,97 @@
...
+		// Omit the . at the end of hosted zone name
+		name := removeLastChar(*zone.Name)
+		if strings.Contains(central.UiHost, name) {
+			rhacsZone = zone
+			break
+		}
+	}
+
+	if rhacsZone == nil {
+		return nil, fmt.Errorf("hosted zone for central host: %v not found", central.UiHost)
```
mtesseract, line 81:
```suggestion
		return nil, fmt.Errorf("failed to find Route53 hosted zone for Central UI host %v", central.UiHost)
```

e2e/e2e_test.go
```diff
@@ -160,30 +164,21 @@ var _ = Describe("Central", func() {
...
-			Expect(*record.Value).To(Equal(reencryptIngress.RouterCanonicalHostname))
-
+			dnsRecordsLoader := dns.NewRecordsLoader(route53Client, central)
+
+			Eventually(dnsRecordsLoader.LoadDNSRecords).
+				WithTimeout(waitTimeout).
+				WithPolling(defaultPolling).
+				Should(HaveLen(len(dnsRecordsLoader.CentralDomainNames)), "Started at %s", time.Now())
+
+			recordSets := dnsRecordsLoader.LastResult
```
mtesseract, line 174: I am still pretty new to Gomega & Ginkgo, but I've been wondering if it would simplify the code a bit if you would pass a function to `Eventually` calling `LoadDNSRecords` and directly doing the assertions there instead of storing the result in `LastResult` and accessing the saved result in a separate testing step. As in e.g. https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode (2nd example). WDYT?

e2e/e2e_test.go
```diff
@@ -160,30 +164,21 @@ var _ = Describe("Central", func() {
...
+			for idx, domain := range dnsRecordsLoader.CentralDomainNames {
+				recordSet := recordSets[idx]
+				Expect(len(recordSet.ResourceRecords)).To(Equal(1))
+				record := recordSet.ResourceRecords[0]
+				Expect(*recordSet.Name).To(Equal(domain))
+				Expect(*record.Value).To(Equal(reencryptIngress.RouterCanonicalHostname)) // TODO use route specific ingress
```
mtesseract, line 180: Maybe this is just me, but I'm not completely sure what the TODO refers to.

e2e/e2e_test.go
kovayur: Reverted. Which token is appropriate to use here for running tests locally?

e2e/dns/records_loader.go
kovayur: Done, returning empty array now

e2e/dns/records_loader.go
kovayur: Changed to `idx++`

e2e/dns/records_loader.go
kovayur: The naming comes from return type, which is `ListResourceRecordSetsOutput`

e2e/dns/records_loader.go
kovayur: Done.

e2e/e2e_test.go
kovayur: I think @johannes94's original idea was to retry until we receive the necessary records, then perform the assertion.
In your case, if the assertion fails, we'll continue retrying until the timeout.

e2e/e2e_test.go
kovayur: Added more information, hope it looks clearer now.

e2e/dns/records_loader.go
kovayur: Fixed.

e2e/e2e_test.go
mtesseract: There is one in BitWarden for accessing the Admin API.
I set `STATIC_TOKEN` and `STATIC_TOKEN_ADMIN` in my `$GITROOT/.envrc`, which works nicely.

mtesseract (approved):

PR #301, opened 2022-08-18
ROX-11614: Add PagerDuty secret to the observability helm chart

0x656b694d:
## Description

[ROX-11614](https://issues.redhat.com/browse/ROX-11614) - Setup ACS PagerDuty as backend for alerting rules.

## Test manual

```sh
$ helm template . --namespace rhacs --set "pagerduty.key=abcdef" | grep -A2 -B9 KEY
# Source: observability/templates/01-operator-03-secret-pagerduty.yaml
apiVersion: v1
kind: Secret
metadata:
  name: rhacs-pagerduty
  namespace: rhacs-observability
  labels:
    configures: observability-operator
stringData:
  PAGERDUTY_KEY: "abcdef"
type: Opaque
---
```

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-pagerduty.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-pagerduty.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: rhacs-pagerduty
+  namespace: {{ include "observability.namespace" . }}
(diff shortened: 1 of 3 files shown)


stehessel:
Is there one key for all clusters or individual keys per cluster? Can we add labels in PagerDuty to distinguish from what cluster the alert has been sent?

0x656b694d:
A key identify a PagerDuty ruleset, which we may create as many as we want. So different clusters using different keys may be handled by rules from different sets on PagerDuty.

stehessel (approved):
LGTM.

PR #304, opened 2022-08-19
Add script to pull secrets and terraform data plane clusters

ebensh:
This script resolves the problem of ensuring we are pulling the correct secrets for the correct environment / data plane cluster.

Tested:
Changed `helm install` to `helm template` in the script and set -x, ran the command with `helm diff upgrade` to ensure no differences against acs-stage-[dp-01](https://issues.redhat.com//browse/dp-01) (known to be working).


dp-terraform/helm/rhacs-terraform/fetch_too_many_secrets.sh
```diff
@@ -0,0 +1,68 @@
...
+OBSERVABILITY_GITHUB_ACCESS_TOKEN=$(bw get password eb7aecd3-b553-4999-b201-aebe01445822)
+OBSERVABILITY_OBSERVATORIUM_METRICS_CLIENT_ID="observatorium-rhacs-metrics-staging"
+OBSERVABILITY_OBSERVATORIUM_METRICS_SECRET=$(
+    bw get item 510c8ed9-ba9f-46d9-b906-ae6100cf72f5 | \
+    jq --arg OBSERVABILITY_OBSERVATORIUM_METRICS_CLIENT_ID "${OBSERVABILITY_OBSERVATORIUM_METRICS_CLIENT_ID}" \
+        '.fields[] | select(.name | contains($OBSERVABILITY_OBSERVATORIUM_METRICS_CLIENT_ID)) | .value' --raw-output
+)
+
+cat <<EOF > ${OUTPUT_FILE}
+fleetshardSync:
```
stehessel, line 52: Why do you use a values file for some of the parameters? I get how that's practical for manual applications of the Helm chart, but can't we set each variable individually in the Helm command? Then you also don't have to worry about the security aspect of writing secrets to a file in plain text.

dp-terraform/helm/rhacs-terraform/terraform_staging_cluster.sh
```diff
@@ -0,0 +1,32 @@
...
+  --debug \
+  --namespace rhacs \
+  --create-namespace \
+  --values=/home/$USER/tmp_secrets/secrets.yaml \
+  --set fleetshardSync.authType="RHSSO" \
+  --set fleetshardSync.fleetManagerEndpoint=${FM_ENDPOINT} \
+  --set fleetshardSync.clusterId=${CLUSTER_ID} \
+  --set acsOperator.enabled=true . \
+  --set acsOperator.source=rhacs-operators \
+  --set acsOperator.startingCSV=rhacs-operator.v3.71.0
```
stehessel, line 28: You removed some of the observability parameters again.

stehessel:
May I also suggest to use [helm secrets](https://github.com/jkroepke/helm-secrets) with a [vals](https://github.com/variantdev/vals) backend? That should give us plenty of secure cloud options for parameter and secret managers to choose from.

dp-terraform/helm/rhacs-terraform/terraform_staging_cluster.sh
ebensh: Thanks for the catch; added back.

dp-terraform/helm/rhacs-terraform/fetch_too_many_secrets.sh
ebensh: 1. It made testing easier because I could test the pulling/bw CLI work separately
2. I thought it would simplify the multiple environments problem, because the secrets file would have the same structure between environments and the terraform script would have the same structure, regardless of the environment. If I combine them into one file then I have to duplicate the value setting and whatever other logic goes in there, which seemed more error prone.

Regardless, I can see it was more useful when I was debugging / developing than when it's in a steady state, so I'll recombine them.

Done.

ebensh:
> May I also suggest to use [helm secrets](https://github.com/jkroepke/helm-secrets) with a [vals](https://github.com/variantdev/vals) backend? That should give us plenty of secure cloud options for parameter and secret managers to choose from.

Yes, this is the next step; I wanted to terraform before sorting out the new secret management to unblock everyone, so this is an in-between step.

ebensh:
@stehessel this is ready for another round of review now, please take a look. Thanks!

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -0,0 +1,99 @@
...
+  --set logging.aws.accessKeyId="${LOGGING_AWS_ACCESS_KEY_ID}" \
+  --set logging.aws.secretAccessKey="${LOGGING_AWS_SECRET_ACCESS_KEY}" \
+  --set observability.enabled=true \
+  --set observability.github.accessToken="${OBSERVABILITY_GITHUB_ACCESS_TOKEN}" \
+  --set observability.github.repository=https://api.github.com/repos/stackrox/rhacs-observability-resources/contents \
+  --set observability.gateway=https://observatorium-mst.api.stage.openshift.com \
+  --set observability.observatorium.metricsClientId="${OBSERVABILITY_OBSERVATORIUM_METRICS_CLIENT_ID}" \
+  --set observability.observatorium.metricsSecret="${OBSERVABILITY_OBSERVATORIUM_METRICS_SECRET}"
+
+# To delete all resources:
```
stehessel, line 97: Isn't `helm uninstall ...` the canonical deletion command?

ebensh: We were in a bad state before where the existing helm install on the cluster was different from what was in the template. I see this as the difference between rm -rf'ing paths and apt removing a package.

I've clarified the comment though.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -0,0 +1,99 @@
...
+esac
+
+# helm uninstall rhacs-terraform --namespace rhacs
+
+# To use ACS Operator from the OpenShift Marketplace:
+#   --set acsOperator.source=redhat-operators
+#   --set acsOperator.sourceNamespace=openshift-marketplace
+
+# helm template ... to debug changes
+helm install rhacs-terraform ./ \
```
stehessel, line 74: This is not idempotent, right? To execute the script a second time we would need `helm upgrade`, and we can add `--install` to encompass both install and upgrade.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -0,0 +1,99 @@
...
+    exit 2
+    ;;
+
+  *)
+    echo "Unknown environment ${ENVIRONMENT}"
+    exit 2
+    ;;
+esac
+
+# helm uninstall rhacs-terraform --namespace rhacs
```
stehessel: This comment seems a bit misplaced. Maybe combine it with the one at the bottom?

stehessel:
I think having a way to create a diff with the existing values can be useful, for example by using https://github.com/databus23/helm-diff.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
ebensh: Good point - changed.

ebensh:
> I think having a way to create a diff with the existing values can be useful, for example by using https://github.com/databus23/helm-diff.

The instructions for this are already here: https://github.com/stackrox/acs-fleet-manager/blob/main/dp-terraform/helm/rhacs-terraform/

I've added this method at the top of the file.

How far are we going to go with polishing the script? We could add a --dry-run flag, a way to spit out the diff, etc., but in the end that's just forwarding parameters to Helm; at that point I would rather go back to having a (chmod restricted) values file so that we can just invoke Helm directly, no?

stehessel:
You are right, let's not go overboard. If we add something like Helm secrets, we can just use a values file where secrets are encoded with their secret manager names. Then we can just remove the whole script again and just use plain Helm with different value files for each environment.

stehessel (approved):

PR #306, opened 2022-08-19
Read data plane cluster id for staging e2e from env

kurlov:
## Description

Use env based  data plane `cluster_id` in e2e staging tests. This env is mapped from `acs-fleet-manager-ci/e2e/data-plane-cluster-id` appSRE vault secret
`


kovayur (approved):

mtesseract:
Reading the ID from env is, of course, fine, but is the Vault the best place for this non-secret item?
I would believe that this would also cause the ID to be redacted in log output, potentially hindering debugging?

mtesseract (approved):

PR #323, opened 2022-08-23
add fleetshard-sync metrics endpoint

johannes94:
## Description
This PR adds a metrics endpoint to the fleetshard-sync component. The endpoint serves default metrics for Go applications, as-well as custom defined metrics required to monitor our SLI/SLOs. By default the server listen to the address `:8080` and can be configured with env variable "FLEETSHARD_METRICS_ADDRESS"

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/config/config.go
+++ b/fleetshard/config/config.go
@@ -19,6 +19,7 @@ type Config struct {
 	OCMRefreshToken      string        `env:"OCM_TOKEN"`
 	StaticToken          string        `env:"STATIC_TOKEN"`
 	CreateAuthProvider   bool          `env:"CREATE_AUTH_PROVIDER" envDefault:"false"`
+	MetricsAddress       string        `env:"METRICS_ADDRESS" envDefault:":8080"`
 }
 
 // GetConfig retrieves the current runtime configuration from the environment and returns it.
--- a/fleetshard/config/config.go
+++ b/fleetshard/config/config.go
@@ -19,7 +19,7 @@ type Config struct {
 	OCMRefreshToken      string        `env:"OCM_TOKEN"`
 	StaticToken          string        `env:"STATIC_TOKEN"`
 	CreateAuthProvider   bool          `env:"CREATE_AUTH_PROVIDER" envDefault:"false"`
-	MetricsAddress       string        `env:"METRICS_ADDRESS" envDefault:":8080"`
+	MetricsAddress       string        `env:"FLEETSHARD_METRICS_ADDRESS" envDefault:":8080"`
 }
 
 // GetConfig retrieves the current runtime configuration from the environment and returns it.
--- a/fleetshard/pkg/fleetshardmetrics/metrics.go
+++ b/fleetshard/pkg/fleetshardmetrics/metrics.go
@@ -31,8 +31,8 @@ func registerCustomMetrics(r prometheus.Registerer) {
 	r.MustRegister(fleetManagerRequests)
 }
 
-// IncrementsK8sRequests increments the metric counter for k8s requests
-func IncrementsK8sRequests() {
+// IncrementK8sRequests increments the metric counter for k8s requests
+func IncrementK8sRequests() {
 	k8sRequests.Inc()
 }
 
--- a/fleetshard/pkg/fleetshardmetrics/metrics.go
+++ b/fleetshard/pkg/fleetshardmetrics/metrics.go
@@ -2,27 +2,40 @@ package fleetshardmetrics
 
 import "github.com/prometheus/client_golang/prometheus"
 
+const metricsPrefix = "acs_fleetshard_"
+
 var (
+	k8sRequests               prometheus.Counter
+	k8sRequestErrors          prometheus.Counter
+	fleetManagerRequests      prometheus.Counter
+	fleetManagerRequestErrors prometheus.Counter
+)
+
+func init() {
+	initMetrics()
+}
+
+func initMetrics() {
 	k8sRequests = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "total_k8s_requests",
+		Name: metricsPrefix + "total_k8s_requests",
 		Help: "The total number of requests send to the target kubernetes cluster",
 	})
 
 	k8sRequestErrors = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "total_k8s_request_errors",
+		Name: metricsPrefix + "total_k8s_request_errors",
 		Help: "The total number of unexpected errors for requests send to the target kubernetes cluster",
 	})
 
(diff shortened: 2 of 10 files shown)


fleetshard/pkg/fleetshardmetrics/metrics_test.go
```diff
@@ -0,0 +1,83 @@
...
+		{
+			metricName:    "total_fleet_manager_request_errors",
+			incrementFunc: IncrementFleetManagerRequestErrors,
+		},
+	}
+	promParser := expfmt.TextParser{}
+
+	for _, tc := range tt {
+		t.Run(tc.metricName, func(t *testing.T) {
+			resetMetrics()
```
SimonBaeumer, line 39: Should this not be cleaned up after a test run instead of being required before?

```suggestion
			defer resetMetrics()
```

fleetshard/pkg/fleetshardmetrics/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+		Help: "The total number of requests send to fleet-manager",
+	})
+
+	fleetManagerRequestErrors = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "total_fleet_manager_request_errors",
+		Help: "The total number of request errors for requests send to fleet-manager",
+	})
+)
+
+func registerCustomMetrics(r prometheus.Registerer) {
```
SimonBaeumer, line 27: Why not introducing a singleton struct with receiver methods for this instead?
Accessibility is nearly the same but can offer cleaner interfaces for reconstructing/resetting metrics.

fleetshard/pkg/fleetshardmetrics/server.go
```diff
@@ -0,0 +1,24 @@
...
+	// Register default metrics to use a dedicated registry instead of prometheus.DefaultRegistry
+	// this makes it easier to isolate metric state when unit testing this package
+	registry.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
+	registry.MustRegister(prometheus.NewGoCollector())
+	registerCustomMetrics(registry)
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{}))
+
+	server := &http.Server{Addr: address, Handler: mux}
```
SimonBaeumer, line 22:
```suggestion
	return &http.Server{Addr: address, Handler: mux}
```

fleetshard/pkg/fleetshardmetrics/metrics_test.go
```diff
@@ -0,0 +1,83 @@
...
+			targetMetric, hasKey := metrics[tc.metricName]
+			assert.Truef(t, hasKey, "expected metrics to contain %s but it did not: %v", tc.metricName, metrics)
+
+			value := targetMetric.Metric[0].Counter.Value
+			assert.Equalf(t, 1.0, *value, "expected metric: %s to have value: %v", tc.metricName, value)
+		})
+	}
+}
+
+func resetMetrics() {
```
SimonBaeumer, line 63: This duplicates the initialization logic from `metrics.go`. Working with a singleton here is more clean (and still allows easy access) and gives a better interface to initialize.

fleetshard/pkg/fleetshardmetrics/server.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// NewMetricsServer returns the metrics server
+func NewMetricsServer(address string) *http.Server {
+	registry := prometheus.NewRegistry()
```
SimonBaeumer, line 12: For my understanding: Is a prometheus registry a set of collector which represent a specific metric, i.e. the fleet-manager requests which can be incremented independently?

fleetshard/config/config.go
```diff
@@ -19,6 +19,7 @@ type Config struct {
 	OCMRefreshToken      string        `env:"OCM_TOKEN"`
 	StaticToken          string        `env:"STATIC_TOKEN"`
 	CreateAuthProvider   bool          `env:"CREATE_AUTH_PROVIDER" envDefault:"false"`
+	MetricsAddress       string        `env:"METRICS_ADDRESS" envDefault:":8080"`
```
stehessel, line 22: Why `8080`? Isn't 9090 the default Prometheus port?

SimonBaeumer (changes requested):
Good job on the implementation!

I think collecting k8s error count does not give us much insights, instead collecting reconcile errors is a better indicator because k8s errors are often expected to inspect the cluster state (i.e. a resource was not found).

fleetshard/pkg/fleetshardmetrics/metrics.go
stehessel, line 27: What do you think about prefixing the metrics? For example I think Central uses a `rox_central_*` prefix for most of its metrics.

fleetshard/config/config.go
stehessel: Why did you go for `MetricsAddress` instead of `MetricsPort`?

stehessel (approved):
LGTM, just some minor comments 🚀

fleetshard/pkg/fleetshardmetrics/metrics.go
johannes94: Can we discuss this in more detail, so that I know i'm understanding correctly?

fleetshard/pkg/fleetshardmetrics/metrics_test.go
johannes94: I refactored the metrics initialization to drop the duplication here.

fleetshard/pkg/fleetshardmetrics/metrics.go
johannes94: added the prefix `acs_fleetshard_*`

SimonBaeumer: I am referring to the Singleton pattern where only one instance of an object exists during the lifetime of the service.
To create a singleton the functions must be part of a struct (so having a receiver). This struct than is initialized only **once** by calling a constructor/factory function.

Example:

```
var (
  service MetricsService
  once     sync.Once
)

func initialize() {
  // init struct here
  service = &MetricsService{}
}

func Singleton() MetricsService {
  once.Do(initialize)
  return service
}
```

You can find several examples in the StackRox code base how to implement Singletons in Go.

SimonBaeumer (changes requested):

SimonBaeumer (approved):
Good job @johannes94!

PR #334, opened 2022-08-25
ROX-12155: Move Central's IdP setup to a worker

rukletsov:
## Description
Prior to this PR, only "static" RHSSO setup for Central is possible, hence `CentralRequest` is augmented right upon being presented to fleetshard. This will not be possible once RHSSO setup becomes more involved, e.g., talking to some API.

In preparation to supporting such involved RHSSO setup, this PR moves augmentation code into a separate "worker", which configures RHSSO in a non-blocking way.

**Note**: it's unclear how to add an E2E test using currently available testing framework. What we probably need to add in follow-up PRs is an ability to query a deployed Central to be able to at least check if an auth provider has been configured as expected (without checking the configuration allows logging in).

## Test manual

Verified in the CI run status transitions of Central request up to Provisioning.

### DIFF
--- a/internal/dinosaur/pkg/api/dbapi/central_request_types.go
+++ b/internal/dinosaur/pkg/api/dbapi/central_request_types.go
@@ -52,6 +52,9 @@ type CentralRequest struct {
 	RoutesCreationID string `json:"routes_creation_id"`
 	// DeletionTimestamp stores the timestamp of the DELETE api call for the resource
 	DeletionTimestamp *time.Time `json:"deletionTimestamp"`
+
+	// All we need to integrate Central with an IdP.
+	AuthConfig
 }
 
 // CentralList ...
@@ -59,6 +62,13 @@ type CentralList []*CentralRequest
 
 // CentralIndex ...
 type CentralIndex map[string]*CentralRequest
+
+// AuthConfig keeps all we need to set up IdP for a Central instance.
+type AuthConfig struct {
+	ClientID     string `json:"idp_client_id"`
+	ClientSecret string `json:"idp_client_secret"`
+	Issuer       string `json:"idp_issuer"`
+}
 
 // Index ...
 func (l CentralList) Index() CentralIndex {
--- a/internal/dinosaur/pkg/api/dbapi/central_request_types.go
+++ b/internal/dinosaur/pkg/api/dbapi/central_request_types.go
@@ -22,7 +22,8 @@ type CentralRequest struct {
 	Owner          string `json:"owner" gorm:"index"` // TODO: ocm owner?
 	OwnerAccountID string `json:"owner_account_id"`
 	OwnerUserID    string `json:"owner_user_id"`
-	// The DNS host (domain) of the Central service
+	// Instance-independent part of the Central's hostname. For example, this
+	// can be `rhacs-dev.com`, `acs-stage.rhcloud.com`, etc.
 	Host           string `json:"host"`
 	OrganisationID string `json:"organisation_id" gorm:"index"`
 	FailedReason   string `json:"failed_reason"`
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/golang/glog"
+	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
 	"github.com/stackrox/acs-fleet-manager/pkg/shared"
 )
@@ -87,11 +88,11 @@ func (c *CentralConfig) ReadFiles() error {
 	// Check that all parts of static auth config are present.
 	if c.HasStaticAuth() {
 		if c.CentralIDPClientSecret == "" {
-			glog.Warningf("no client_secret specified for static client_id %q;"+
+			return errors.Errorf("no client_secret specified for static client_id %q;"+
 				" auth configuration is either incorrect or insecure", c.CentralIDPClientID)
 		}
 		if c.CentralIDPIssuer == "" {
-			glog.Errorf("no issuer specified for static client_id %q;"+
+			return errors.Errorf("no issuer specified for static client_id %q;"+
 				" auth configuration will likely not work properly", c.CentralIDPClientID)
 		}
 	}
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -83,6 +83,19 @@ func (c *CentralConfig) ReadFiles() error {
 	} else {
 		glog.Infof("Central's IdP client secret from file %q is missing", c.CentralIDPClientSecretFile)
 	}
+
+	// Check that all parts of static auth config are present.
+	if c.HasStaticAuth() {
+		if c.CentralIDPClientSecret == "" {
+			glog.Warningf("no client_secret specified for static client_id %q;"+
+				" auth configuration is either incorrect or insecure", c.CentralIDPClientID)
+		}
+		if c.CentralIDPIssuer == "" {
+			glog.Errorf("no issuer specified for static client_id %q;"+
+				" auth configuration will likely not work properly", c.CentralIDPClientID)
+		}
+	}
+
 	// TODO(ROX-11289): drop MaxCapacity
 	// MaxCapacity is deprecated and will not be used.
 	// Temporarily set MaxCapacity manually in order to simplify app start.
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -89,3 +89,12 @@ func (c *CentralConfig) ReadFiles() error {
 	c.MaxCapacity = MaxCapacityConfig{1000}
 	return nil
 }
+
+// HasStaticAuth returns true if the static auth config for Centrals has been
+// specified and false otherwise.
+func (c *CentralConfig) HasStaticAuth() bool {
+	// We don't look at other integral parts of the auth config like
+	// RhSsoIssuer or RhSsoClientSecret. Failure to provide a working auth
+	// configuration should not mask an intent to use static configuration.
+	return c.RhSsoClientID != ""
+}
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -96,5 +96,5 @@ func (c *CentralConfig) HasStaticAuth() bool {
 	// We don't look at other integral parts of the auth config like
 	// RhSsoIssuer or RhSsoClientSecret. Failure to provide a working auth
 	// configuration should not mask an intent to use static configuration.
-	return c.RhSsoClientID != ""
+	return c.CentralIDPClientID != ""
 }
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -849,7 +849,7 @@ func (k *dinosaurService) ListCentralsWithoutAuthConfig() ([]*dbapi.CentralReque
 
 	dbConn := k.connectionFactory.New().
 		Where("status IN (?)", status).
-		Where("client_id == ''")
+		Where("client_id LIKE ''")
 
 	var results []*dbapi.CentralRequest
 	if err := dbConn.Find(&results).Error; err != nil {
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -847,14 +847,13 @@ func (k *dinosaurService) ListCentralsWithoutAuthConfig() ([]*dbapi.CentralReque
 		dinosaurConstants.CentralRequestStatusPreparing,
 	}
 
-	dbConn := k.connectionFactory.New()
-	var results []*dbapi.CentralRequest
-	if err := dbConn.
+	dbConn := k.connectionFactory.New().
 		Where("status IN (?)", status).
-		Where("idp_client_id == ''").
-		Scan(&results).
-		Error; err != nil {
-		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to list dinosaur requests")
+		Where("idp_client_id == ''")
+
+	var results []*dbapi.CentralRequest
+	if err := dbConn.Find(&results).Error; err != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to list Central requests")
 	}
 	return results, nil
 }
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -29,8 +29,17 @@ import (
 )
 
 var (
-	dinosaurDeletionStatuses  = []string{dinosaurConstants.CentralRequestStatusDeleting.String(), dinosaurConstants.CentralRequestStatusDeprovision.String()}
-	dinosaurManagedCRStatuses = []string{dinosaurConstants.CentralRequestStatusProvisioning.String(), dinosaurConstants.CentralRequestStatusDeprovision.String(), dinosaurConstants.CentralRequestStatusReady.String(), dinosaurConstants.CentralRequestStatusFailed.String()}
+	dinosaurDeletionStatuses = []string{
+		dinosaurConstants.CentralRequestStatusDeleting.String(),
+		dinosaurConstants.CentralRequestStatusDeprovision.String(),
+	}
+
+	dinosaurManagedCRStatuses = []string{
+		dinosaurConstants.CentralRequestStatusProvisioning.String(),
+		dinosaurConstants.CentralRequestStatusDeprovision.String(),
+		dinosaurConstants.CentralRequestStatusReady.String(),
+		dinosaurConstants.CentralRequestStatusFailed.String(),
+	}
 )
 
 // DinosaurRoutesAction ...
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -847,7 +847,7 @@ func (k *dinosaurService) ListCentralsWithoutAuthConfig() ([]*dbapi.CentralReque
 		dinosaurConstants.CentralRequestStatusPreparing,
 	}
 
-	dbConn := k.connectionFactory.New().
+	dbQuery := k.connectionFactory.New().
 		Where("status IN (?)", status).
 		Where("client_id = ''")
 
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -334,7 +334,7 @@ func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralR
 		Meta: api.Meta{
 			ID: dinosaurRequest.ID,
 		},
-		Status:      dinosaurConstants.CentralRequestStatusProvisioning.String(),
+		Status: dinosaurConstants.CentralRequestStatusProvisioning.String(),
 	}
 	if err := k.Update(updatedCentralRequest); err != nil {
 		return errors.NewWithCause(errors.ErrorGeneral, err, "failed to update dinosaur request")
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -849,7 +849,7 @@ func (k *dinosaurService) ListCentralsWithoutAuthConfig() ([]*dbapi.CentralReque
 
 	dbConn := k.connectionFactory.New().
 		Where("status IN (?)", status).
-		Where("idp_client_id == ''")
+		Where("client_id == ''")
 
 	var results []*dbapi.CentralRequest
 	if err := dbConn.Find(&results).Error; err != nil {
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -102,6 +102,7 @@ type DinosaurService interface {
 	CountByStatus(status []dinosaurConstants.CentralStatus) ([]DinosaurStatusCount, error)
 	CountByRegionAndInstanceType() ([]DinosaurRegionCount, error)
 	ListDinosaursWithRoutesNotCreated() ([]*dbapi.CentralRequest, *errors.ServiceError)
+	ListDinosaursWithoutAuthConfig() ([]*dbapi.CentralRequest, *errors.ServiceError)
 	VerifyAndUpdateDinosaurAdmin(ctx context.Context, dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError
 	ListComponentVersions() ([]DinosaurComponentVersions, error)
 }
@@ -318,7 +319,15 @@ func (k *dinosaurService) AcceptDinosaurRequest(dinosaurRequest *dbapi.CentralRe
 func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError {
 	// Check if the request is ready to be transitioned to provisioning.
 
-	// TODO(alexr): check RH SSO is set up
+	// Check IdP config is ready.
+	//
+	// TODO(alexr): Shall this go into "preparing_dinosaurs_mgr.go"? Ideally,
+	//     all CentralRequest updating logic is in one place, either in this
+	//     service or workers.
+	if dinosaurRequest.AuthConfig.ClientID == "" {
+		// We can't provision this request, skip
+		return nil
+	}
 
 	// Update the fields of the CentralRequest record in the database.
 	updatedDinosaurRequest := &dbapi.CentralRequest{
@@ -823,6 +832,28 @@ func (k *dinosaurService) ListDinosaursWithRoutesNotCreated() ([]*dbapi.CentralR
 	dbConn := k.connectionFactory.New()
 	var results []*dbapi.CentralRequest
 	if err := dbConn.Where("routes IS NOT NULL").Where("routes_created = ?", "no").Find(&results).Error; err != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to list dinosaur requests")
+	}
+	return results, nil
+}
+
+// ListDinosaursWithoutAuthConfig returns all _relevant_ central requests with
+// no auth config.
+func (k *dinosaurService) ListDinosaursWithoutAuthConfig() ([]*dbapi.CentralRequest, *errors.ServiceError) {
+	// There is no value in augmenting auth config for central requests beyond
+	// 'Preparing'.
+	status := []dinosaurConstants.DinosaurStatus{
+		dinosaurConstants.DinosaurRequestStatusAccepted,
+		dinosaurConstants.DinosaurRequestStatusPreparing,
+	}
+
+	dbConn := k.connectionFactory.New()
+	var results []*dbapi.CentralRequest
+	if err := dbConn.
+		Where("status IN (?)", status).
+		Where("idp_client_id != ''").
+		Scan(&results).
+		Error; err != nil {
 		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to list dinosaur requests")
 	}
 	return results, nil
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -64,9 +64,9 @@ type DinosaurService interface {
 	HasAvailableCapacity() (bool, *errors.ServiceError)
 	// HasAvailableCapacityInRegion checks if there is capacity in the clusters for a given region
 	HasAvailableCapacityInRegion(dinosaurRequest *dbapi.CentralRequest) (bool, *errors.ServiceError)
-	// PrepareDinosaurRequest sets any required information (i.e. dinosaur host, sso client id and secret)
-	// to the Dinosaur Request record in the database. The dinosaur request will also be updated with an updated_at
-	// timestamp and the corresponding cluster identifier.
+	// AcceptDinosaurRequest transitions CentralRequest to 'Preparing'.
+	AcceptDinosaurRequest(dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError
+	// PrepareDinosaurRequest transitions CentralRequest to 'Provisioning'.
 	PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError
 	// Get method will retrieve the dinosaurRequest instance that the give ctx has access to from the database.
 	// This should be used when you want to make sure the result is filtered based on the request context.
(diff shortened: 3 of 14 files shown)


internal/dinosaur/pkg/api/dbapi/central_request_types.go
```diff
@@ -51,6 +52,9 @@ type CentralRequest struct {
 	RoutesCreationID string `json:"routes_creation_id"`
 	// DeletionTimestamp stores the timestamp of the DELETE api call for the resource
 	DeletionTimestamp *time.Time `json:"deletionTimestamp"`
+
+	// All we need to integrate Central with an IdP.
+	AuthConfig
```
misberner, line 57:
```suggestion
	AuthConfig AuthConfig `json:"authConfig"`
```
You don't seem to want to use this as an embedded struct field anyway

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -29,8 +29,17 @@ import (
...
-	dinosaurDeletionStatuses  = []string{dinosaurConstants.DinosaurRequestStatusDeleting.String(), dinosaurConstants.DinosaurRequestStatusDeprovision.String()}
-	dinosaurManagedCRStatuses = []string{dinosaurConstants.DinosaurRequestStatusProvisioning.String(), dinosaurConstants.DinosaurRequestStatusDeprovision.String(), dinosaurConstants.DinosaurRequestStatusReady.String(), dinosaurConstants.DinosaurRequestStatusFailed.String()}
+	dinosaurDeletionStatuses = []string{
+		dinosaurConstants.DinosaurRequestStatusDeleting.String(),
+		dinosaurConstants.DinosaurRequestStatusDeprovision.String(),
+	}
+
+	dinosaurManagedCRStatuses = []string{
+		dinosaurConstants.DinosaurRequestStatusProvisioning.String(),
+		dinosaurConstants.DinosaurRequestStatusDeprovision.String(),
```
misberner, line 39: Is it intentional that `Deprovision` occurs both here and in `dinosaurDeletionStatuses`?

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -261,14 +271,19 @@ func (k *dinosaurService) RegisterDinosaurJob(dinosaurRequest *dbapi.CentralRequ
 	return nil
 }
 
-// PrepareDinosaurRequest ...
-func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError {
+// AcceptDinosaurRequest sets any information about Central that does not
+// require blocking operations (e.g., requesting a RHSSO client). Upon success,
```
misberner, line 275:
```suggestion
// require blocking operations (as opposed to, e.g., requesting a RHSSO client). Upon success,
```

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -298,6 +312,36 @@ func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralR
...
+func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError {
+	// Check if the request is ready to be transitioned to provisioning.
+
+	// Check IdP config is ready.
+	//
+	// TODO(alexr): Shall this go into "preparing_dinosaurs_mgr.go"? Ideally,
+	//     all CentralRequest updating logic is in one place, either in this
+	//     service or workers.
+	if dinosaurRequest.AuthConfig.ClientID == "" {
+		// We can't provision this request, skip
```
misberner, line 327: So the status will never go into provisioning? What does that mean in practice?

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -792,6 +836,28 @@ func (k *dinosaurService) ListDinosaursWithRoutesNotCreated() ([]*dbapi.CentralR
...
+	status := []dinosaurConstants.DinosaurStatus{
+		dinosaurConstants.DinosaurRequestStatusAccepted,
+		dinosaurConstants.DinosaurRequestStatusPreparing,
+	}
+
+	dbConn := k.connectionFactory.New()
+	var results []*dbapi.CentralRequest
+	if err := dbConn.
+		Where("status IN (?)", status).
+		Where("idp_client_id != ''").
```
misberner, line 853: Shouldn't this be `==`?

internal/dinosaur/pkg/workers/dinosaurmgrs/accepted_dinosaurs_mgr.go
```diff
@@ -126,10 +126,10 @@ func (k *AcceptedDinosaurManager) reconcileAcceptedDinosaur(dinosaur *dbapi.Cent
 	}
 	dinosaur.DesiredCentralVersion = selectedDinosaurOperatorVersion.CentralVersions[len(selectedDinosaurOperatorVersion.CentralVersions)-1].Version
 
-	glog.Infof("Dinosaur instance with id %s is assigned to cluster with id %s", dinosaur.ID, dinosaur.ClusterID)
-	dinosaur.Status = constants2.DinosaurRequestStatusPreparing.String()
-	if err2 := k.dinosaurService.Update(dinosaur); err2 != nil {
-		return errors.Wrapf(err2, "failed to update dinosaur %s with cluster details", dinosaur.ID)
+	glog.Infof("Central instance with id %s is assigned to cluster with id %s", dinosaur.ID, dinosaur.ClusterID)
+
+	if err = k.dinosaurService.AcceptDinosaurRequest(dinosaur); err != nil {
```
misberner, line 131: Is it the normal procedure that workers call `AcceptDinsoaurRequest` directly? Just curious because there was no comparable call before

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,102 @@
...
+
+	centralRequests, listErr := k.centralService.ListDinosaursWithoutAuthConfig()
+	if listErr != nil {
+		errs = append(errs, errors.Wrap(listErr, "failed to list centrals without auth config"))
+	} else {
+		glog.V(5).Infof("%d central(s) need auth config to be added", len(centralRequests))
+	}
+
+	for _, cr := range centralRequests {
+		if err := augmentWithAuthConfig(cr, k.centralConfig); err != nil {
```
misberner, line 58:
```suggestion
		if err := k.augmentWithAuthConfig(cr); err != nil {
```
seems more fitting to turn that into a method?

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -54,9 +63,9 @@ type DinosaurService interface {
 	HasAvailableCapacity() (bool, *errors.ServiceError)
 	// HasAvailableCapacityInRegion checks if there is capacity in the clusters for a given region
 	HasAvailableCapacityInRegion(dinosaurRequest *dbapi.CentralRequest) (bool, *errors.ServiceError)
-	// PrepareDinosaurRequest sets any required information (i.e. dinosaur host, sso client id and secret)
-	// to the Dinosaur Request record in the database. The dinosaur request will also be updated with an updated_at
-	// timestamp and the corresponding cluster identifier.
+	// AcceptDinosaurRequest transitions CentralRequest to 'Preparing'.
+	AcceptDinosaurRequest(dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError
```
misberner, line 67:
```suggestion
	// AcceptCentralRequest transitions CentralRequest to 'Preparing'.
	AcceptCentralRequest(centralRequest *dbapi.CentralRequest) *errors.ServiceError
```
Let's keep the dinosaur count monotonically decreasing

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -92,6 +101,7 @@ type DinosaurService interface {
 	CountByStatus(status []dinosaurConstants.DinosaurStatus) ([]DinosaurStatusCount, error)
 	CountByRegionAndInstanceType() ([]DinosaurRegionCount, error)
 	ListDinosaursWithRoutesNotCreated() ([]*dbapi.CentralRequest, *errors.ServiceError)
+	ListDinosaursWithoutAuthConfig() ([]*dbapi.CentralRequest, *errors.ServiceError)
```
misberner, line 104:
```suggestion
	ListCentralsWithoutAuthConfig() ([]*dbapi.CentralRequest, *errors.ServiceError)
```
same here

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
misberner: Nvm, I see now why it isn't one.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,102 @@
...
+		r.AuthConfig.ClientSecret = centralConfig.RhSsoClientSecret
+		r.AuthConfig.Issuer = centralConfig.RhSsoIssuer
+
+		return nil
+	}
+
+	// Proceed with the dynamic path.
+	glog.V(7).Infoln("no static config found; requesting dynamic client")
+
+	// TODO(alexr): Talk to RHSSO dynamic client API.
```
misberner, line 99: I think it would be nicer to factor out the "static config exists?" check to run outside of this function, and call the same function for static vs dynamically provisioned config

misberner (changes requested):

internal/dinosaur/pkg/services/dinosaur.go
rukletsov: I believe so. `dinosaurManagedCRStatuses` is a collection of statuses that are used by `ListByClusterID()`, which sounds generic but apparently for some reason we're only interested in CentralRequests in some state. `ListByClusterID()` is only used to hand off work to fleetshard sync, so `DinosaurRequestStatusDeprovision` makes sense to me though I don't know how to verify which statuses are of interest to fleetshard sync.

internal/dinosaur/pkg/api/dbapi/central_request_types.go
rukletsov: I'm not sure about this. Look how any non-basic type is represented as `api.JSON`, e.g., `Routes`. I definitely don't want to store `AuthConfig` as JSON here nor I fancy figuring out why we don't simply use nested structs.

So yeah, I want to embed this. Would you suggest dropping `AuthConfig` when accessing its fields?

internal/dinosaur/pkg/services/dinosaur.go
rukletsov: My understanding of the fleet manager concept is that "workers" make updates to central requests, including status transitions. `PrepareDinosaurRequest()` is called from one of the worker that aims to transition some central requests into provisioning status. So what I think is going to happen here, every now and then `PrepareDinosaurRequest()` is called on a central request instance and is a no-op until after a different worker in `dinosaurs_auth_config.go` updates request's auth config.

internal/dinosaur/pkg/services/dinosaur.go
rukletsov: 🤦‍♂️

internal/dinosaur/pkg/workers/dinosaurmgrs/accepted_dinosaurs_mgr.go
rukletsov: That's what I picked up from other workers, e.g., [this](https://github.com/stackrox/acs-fleet-manager/blob/d3942c334d7b99e5b4167824d92bc538f698fa04/internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_dinosaurs_mgr.go#L107) and [this](https://github.com/stackrox/acs-fleet-manager/blob/d3942c334d7b99e5b4167824d92bc538f698fa04/internal/dinosaur/pkg/workers/dinosaurmgrs/preparing_dinosaurs_mgr.go#L75).

However, I agree that state transitioning logic is hard to comprehend.

internal/dinosaur/pkg/services/dinosaur.go
rukletsov: Hehe :). Yeah, I was on the fence here. I didn't spawn any dinosaurs in the newly created file, but here I'm not sure which strategy makes most sense: being consistent and 🪓 all dinosaurs at once or use the methodical one-dinosaur-at-a-time approach.

docs/legacy/feature-flags.md
```diff
@@ -34,10 +34,18 @@ This lists the feature flags and their sub-configurations to enable/disable and
...
     - `central-tls-cert-file` [Required]: The path to the file containing the Central TLS certificate (default: `'secrets/central-tls.crt'`).
     - `central-tls-key-file` [Required]: The path to the file containing the Central TLS private key (default: `'secrets/central-tls.key'`).
-- **enable-evaluator-instance**: Enable the creation of one central evaluator instances per user  
-- **rhsso-client-id**: RHSSO client ID to pass to Central's auth config to set up RHSSO IdP
-- **rhsso-client-secret-file**: File containing RHSSO client secret to pass to Central's auth config to set up RHSSO IdP
-- **rhsso-issuer**: RHSSO issuer to pass to Central's auth config to set up RHSSO IdP
+- **enable-evaluator-instance**: Enable the creation of one central evaluator instances per user
+
+- **rhsso-***: A collection of flags describing _static_ auth config for Central.
+  If set, every Central will have the **same** IdP config—likely not what you want
```
ivan-degtiarenko, line 40:
```suggestion
  If set, every Central will have the **same** IdP config — likely not what you want
```

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -262,14 +272,19 @@ func (k *dinosaurService) RegisterDinosaurJob(dinosaurRequest *dbapi.CentralRequ
 	return nil
 }
 
-// PrepareDinosaurRequest ...
-func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError {
+// AcceptCentralRequest sets any information about Central that does not
+// require blocking operations (deducing namespace or instance hostname). Upon
```
ivan-degtiarenko, line 276: What do you consider blocking operation? `GetClusterDNS` can potentially end up in a request to OCM service, is it non-blocking?

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -793,6 +837,27 @@ func (k *dinosaurService) ListDinosaursWithRoutesNotCreated() ([]*dbapi.CentralR
 	return results, nil
 }
 
+// ListCentralsWithoutAuthConfig returns all _relevant_ central requests with
+// no auth config.
+func (k *dinosaurService) ListCentralsWithoutAuthConfig() ([]*dbapi.CentralRequest, *errors.ServiceError) {
+	// There is no value in augmenting auth config for central requests beyond
```
ivan-degtiarenko, line 843: Good thinking! I suggest we take it one step further and verify(not necessarily in this method) that there are no central requests without auth config in status provisioning and ready.

internal/dinosaur/pkg/workers/dinosaurmgrs/accepted_dinosaurs_mgr.go
```diff
@@ -126,10 +126,10 @@ func (k *AcceptedDinosaurManager) reconcileAcceptedDinosaur(dinosaur *dbapi.Cent
 	}
 	dinosaur.DesiredCentralVersion = selectedDinosaurOperatorVersion.CentralVersions[len(selectedDinosaurOperatorVersion.CentralVersions)-1].Version
 
-	glog.Infof("Dinosaur instance with id %s is assigned to cluster with id %s", dinosaur.ID, dinosaur.ClusterID)
-	dinosaur.Status = constants2.CentralRequestStatusPreparing.String()
-	if err2 := k.dinosaurService.Update(dinosaur); err2 != nil {
-		return errors.Wrapf(err2, "failed to update dinosaur %s with cluster details", dinosaur.ID)
+	glog.Infof("Central instance with id %s is assigned to cluster with id %s", dinosaur.ID, dinosaur.ClusterID)
+
+	if err = k.dinosaurService.AcceptCentralRequest(dinosaur); err != nil {
```
ivan-degtiarenko, line 131: I think the name might be confusing: every central request passed down to `reconcileAcceptedDinosaur` is already accepted. Maybe something like `PreprocessCentralRequest`? I'm not so great at naming methods

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,111 @@
...
+	for _, cr := range centralRequests {
+		glog.V(5).Infof("augmenting Central %q with auth config", cr.Meta.ID)
+		// Auth config can either be:
+		//   1) static, i.e., the same for all Centrals,
+		//   2) dynamic, i.e., each Central has its own.
+		// In case of 1), all necessary information should be provided in
+		// CentralConfig. For 2), we need to request a dynamic client from the
+		// RHSSO API.
+
+		var augmentWithAuthConfigF func(r *dbapi.CentralRequest, centralConfig *config.CentralConfig) error
```
ivan-degtiarenko, line 66: nit: I'm not sure whether it's a more idiomatic way compared to having a function with `if-else` inside.

Considering that's the result of @misberner suggestion, let's ask him :)

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,111 @@
...
+	if centralConfig.RhSsoClientSecret == "" {
+		glog.Warningf("no client_secret specified for static client_id %q;" +
+			" auth configuration is either incorrect or insecure", centralConfig.RhSsoClientID)
+	}
+	if centralConfig.RhSsoIssuer == "" {
+		glog.Errorf("no issuer specified for static client_id %q;" +
+			" auth configuration will likely not work properly", centralConfig.RhSsoClientID)
+	}
+
+	r.AuthConfig.ClientID = centralConfig.RhSsoClientID
```
ivan-degtiarenko, line 96: Where do we save these modifications to the database?

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	return errs
+}
+
+// augmentWithStaticAuthConfig augments provided CentralRequest with static auth
+// config information, i.e., the same for all Centrals.
+func augmentWithStaticAuthConfig(r *dbapi.CentralRequest, centralConfig *config.CentralConfig) error {
+	glog.V(7).Infoln("static config found; no dynamic client will be requested from IdP")
```
ivan-degtiarenko, line 84: Why it's `glog.V(7)` here while `glog.V(5)` above?

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,111 @@
...
+	r.AuthConfig.ClientSecret = centralConfig.RhSsoClientSecret
+	r.AuthConfig.Issuer = centralConfig.RhSsoIssuer
+
+	return nil
+}
+
+// augmentWithDynamicAuthConfig performs all necessary rituals to obtain auth
+// configuration via RHSSO API.
+func augmentWithDynamicAuthConfig(_ *dbapi.CentralRequest, _ *config.CentralConfig) error {
+	glog.V(7).Infoln("")
```
ivan-degtiarenko, line 106: Even though this method will be rewritten when we will integrate with RH SSO API, let's log something meaningful here for now.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return errs
+}
+
+// augmentWithStaticAuthConfig augments provided CentralRequest with static auth
+// config information, i.e., the same for all Centrals.
+func augmentWithStaticAuthConfig(r *dbapi.CentralRequest, centralConfig *config.CentralConfig) error {
+	glog.V(7).Infoln("static config found; no dynamic client will be requested from IdP")
+
+	// TODO(alexr): Ideally this belongs in a config validation routine.
```
ivan-degtiarenko, line 86: What's stopping us from adding this validation here:
https://github.com/stackrox/acs-fleet-manager/blob/5a987736238906fc86aecc51aa3a70b3828cb648/internal/dinosaur/pkg/config/central.go#L65-L65?

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
```diff
@@ -0,0 +1,111 @@
+package dinosaurmgrs
```
ivan-degtiarenko, line 1: the filename does not reflect the fact this file contains `...Manager` code

ivan-degtiarenko (commented):
Great work! Left some comments.

I'm trying to understand how partially moving work from `Preparing` to `Accepted` status(state) might affect the overall flow. The important question is: is there a difference between failing while in the `Accepted` state vs in the `Preparing` state?  Unfortunately, I don't have a concrete answer for now.

docs/legacy/feature-flags.md
rukletsov: Generally, em-dash is not surrounded by spaces in English [1, 2].

[1] https://english.stackexchange.com/questions/214585/space-before-after-emdash-or-not
[2] https://www.merriam-webster.com/words-at-play/em-dash-en-dash-how-to-use

internal/dinosaur/pkg/services/dinosaur.go
rukletsov: Every network operation can block. If `GetClusterDNS()` can block, it should be taken out of this function and put into a separate worker.

internal/dinosaur/pkg/workers/dinosaurmgrs/accepted_dinosaurs_mgr.go
rukletsov: I aligned method names with status names. From what I got from the code, on request in status `StatusX`, `X()` is called. I agree this is confusing, but in Rome do as the Romans do...

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
rukletsov: Yeah, a call to `dinosaurService.Update()` is missing — I'd like to get the read query work first : )

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+	for _, cr := range centralRequests {
+		glog.V(5).Infof("augmenting Central %q with auth config", cr.Meta.ID)
+		// Auth config can either be:
+		//   1) static, i.e., the same for all Centrals,
+		//   2) dynamic, i.e., each Central has its own.
+		// In case of 1), all necessary information should be provided in
+		// CentralConfig. For 2), we need to request a dynamic client from the
+		// RHSSO API.
+
+		var augmentWithAuthConfigF func(r *dbapi.CentralRequest, centralConfig *config.CentralConfig) error
```
misberner, line 66: Hmm, is this really necessary? What benefit does it provide over calling `augmentWith(Static|Dynamic)Config` inline instead of assigning to this variable? You just need to declare `var err error` before the `if` statement.

misberner (changes requested):

docs/legacy/feature-flags.md
ivan-degtiarenko: Sorry for wasting your time on this, folks - it's not principal at all(should've marked with `nit`). If you feel like it looks nice, let's leave it as that

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
ivan-degtiarenko: I guess my question is more general: where do these numbers come from? Did you take some other piece of code as a reference to use numbers 5 and 7, or was it just a gut feeling?

I think it would be nice at some point to come up with a system which messages go to which level if it does not exist yet. Definitely not within this PR, but probably would be good to add it to our tech debt

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
ivan-degtiarenko: +1

internal/dinosaur/pkg/workers/dinosaurmgrs/accepted_dinosaurs_mgr.go
ivan-degtiarenko: Fair enough, let's not rework this within this PR

internal/dinosaur/pkg/services/dinosaur.go
ivan-degtiarenko: I'm not sure about the separate worker - I would argue having a worker per remote call is too granular a separation of concern. I would say it would be more simple and clear to move this call to `Prepare...` function.

ivan-degtiarenko (commented):
Small batch of comments

docs/legacy/feature-flags.md
rukletsov: Avoided em-dash and moved this to https://github.com/stackrox/acs-fleet-manager/pull/381.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
rukletsov: ¯\_(ツ)_/¯

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config.go
rukletsov: I think we could. It was not clear to me that it's a good place for verification but looking at other implementations of the interface, it probably is.

internal/dinosaur/pkg/config/central.go
```diff
@@ -83,9 +83,31 @@ func (c *CentralConfig) ReadFiles() error {
...
 	}
+
+	// Check that all parts of static auth config are present.
+	if c.HasStaticAuth() {
+		if c.CentralIDPClientSecret == "" {
+			glog.Warningf("no client_secret specified for static client_id %q;"+
+				" auth configuration is either incorrect or insecure", c.CentralIDPClientID)
+		}
+		if c.CentralIDPIssuer == "" {
+			glog.Errorf("no issuer specified for static client_id %q;"+
```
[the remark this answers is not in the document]
ivan-degtiarenko: +1

internal/dinosaur/pkg/services/dinosaur.go
ivan-degtiarenko: @rukletsov WDYT?

ivan-degtiarenko (approved):
One comment was left without an answer, otherwise LGTM

internal/dinosaur/pkg/services/dinosaur.go
rukletsov: I think an explicit call per status transition is good regardless. Regarding separate workers... I don't think it matters much. Goroutines are cheap, so should be the workers. Whether having one worker per logical task or one per status transition — does not matter much because the whole worker concept is a bit of a miss here.

We can refactor later.

internal/dinosaur/pkg/config/central.go
```diff
@@ -83,9 +83,31 @@ func (c *CentralConfig) ReadFiles() error {
 	} else {
 		glog.Infof("Central's IdP client secret from file %q is missing", c.CentralIDPClientSecretFile)
 	}
+
+	// Check that all parts of static auth config are present.
+	if c.HasStaticAuth() {
+		if c.CentralIDPClientSecret == "" {
+			glog.Warningf("no client_secret specified for static client_id %q;"+
```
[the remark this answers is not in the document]
rukletsov: No option here says or guarantees it's a code exchange, so the fact that client_secret is required as you mention in the comment is sort of non-local knowledge. However, I'm fine logging and returning an error here.

internal/dinosaur/pkg/config/central.go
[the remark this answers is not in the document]
rukletsov: 🤷‍♂️ this is not really fleet manager job to verify correctness of Central's config. It shall bypass what it gets, emitting advice where it thinks it might not be OK.

However, I don't have a strong opinion and will return error.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// CentralAuthConfigManager updates CentralRequests with auth configuration.
+type CentralAuthConfigManager struct {
+	workers.BaseWorker
+	centralService services.DinosaurService
+	centralConfig  *config.CentralConfig
+}
+
+var _ workers.Worker = &CentralAuthConfigManager{}
```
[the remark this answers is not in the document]
rukletsov: Yeah, it kinds is but we don't have to follow I think. Will adjust.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
```diff
@@ -0,0 +1,103 @@
...
+			err = augmentWithStaticAuthConfig(cr, k.centralConfig)
+		} else {
+			glog.V(7).Infoln("no static config found; attempting to obtain one from the IdP")
+			err = augmentWithDynamicAuthConfig(cr, k.centralConfig)
+		}
+		if err != nil {
+			errs = append(errs, errors.Wrap(err, "failed to augment central request with auth config"))
+		}
+
+		if err := k.centralService.Update(cr); err != nil {
```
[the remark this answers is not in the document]
rukletsov: It should be fine call `Update()` if `augmentWithXYZAuthConfig()` is not written by Machiavelli but agree let's not do it.

internal/dinosaur/pkg/config/central.go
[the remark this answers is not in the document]
rukletsov: Yeah, tests fall apart now because surely we didn't follow the practise in test...

misberner (approved):

PR #340, opened 2022-08-29
Delete leftover namespaces in down script

mtesseract:
## Description

When debugging issues in the e2e test suite it happens easily that we have to cleanup leftovers. In my opinion it makes sense to integrate this into `down.sh`.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/dev/env/scripts/down.sh
+++ b/dev/env/scripts/down.sh
@@ -18,7 +18,7 @@ delete "${MANIFESTS_DIR}/fleetshard-sync" || true
 central_namespaces=$($KUBECTL get namespace -o jsonpath='{range .items[?(@.status.phase == "Active")]}{.metadata.name}{"\n"}{end}' | grep '^rhacs-.*$')
 
 for namespace in $central_namespaces; do
-    kubectl delete namespace "$namespace" &
+    $KUBECTL delete namespace "$namespace" &
 done
 log "Waiting for leftover RHACS namespaces to be deleted... "
 for p in $(jobs -pr); do
--- a/dev/env/scripts/down.sh
+++ b/dev/env/scripts/down.sh
@@ -14,3 +14,13 @@ port-forwarding stop db 5432 || true
 delete "${MANIFESTS_DIR}/db" || true
 delete "${MANIFESTS_DIR}/fleet-manager" || true
 delete "${MANIFESTS_DIR}/fleetshard-sync" || true
+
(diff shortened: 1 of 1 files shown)


dev/env/scripts/down.sh
```diff
@@ -14,3 +14,13 @@ port-forwarding stop db 5432 || true
 delete "${MANIFESTS_DIR}/db" || true
 delete "${MANIFESTS_DIR}/fleet-manager" || true
 delete "${MANIFESTS_DIR}/fleetshard-sync" || true
+
+central_namespaces=$($KUBECTL get namespace -o jsonpath='{range .items[?(@.status.phase == "Active")]}{.metadata.name}{"\n"}{end}' | grep '^rhacs-.*$')
```
SimonBaeumer, line 18: Where is `$KUBECTL` var declared?

SimonBaeumer (changes requested):

dev/env/scripts/down.sh
mtesseract: https://github.com/stackrox/acs-fleet-manager/blob/mc%2Fdebug-e2e-local/dev/env/scripts/lib.sh#L89

SimonBaeumer: Link is broken, but I see where it is coming from.
Should the `kubectl` in line 21 then be replaced?
https://github.com/stackrox/acs-fleet-manager/pull/340/files#diff-399711d07af14f32ea9544fecfd0cd1690b9505d0c21964c883a1b5f859fa3d8R21

SimonBaeumer (approved):

dev/env/scripts/down.sh
SimonBaeumer: @mtesseract that was fast. 👍 Thank you!

mtesseract: Sorry, was on the wrong (non-pushed) branch... 🤦

mtesseract:
@SimonBaeumer I managed to introduce a bug in in this tiny PR: `grep` would fail with a non-zero exit code in case no such namespaces are found and due to `set -e` being active the script would be aborted immediately. The `|| true` seems to be the common practice for enforcing a success exit code irregardless of matches being found or not.

SimonBaeumer:
> @SimonBaeumer I managed to introduce a bug in in this tiny PR: `grep` would fail with a non-zero exit code in case no such namespaces are found and due to `set -e` being active the script would be aborted immediately. The `|| true` seems to be the common practice for enforcing a success exit code irregardless of matches being found or not.

Sounds good 👍

PR #349, opened 2022-08-30
ROX-12394: Configurable Tenant Defaults

mtesseract:
## Description

Rework resource defaulting: Instead of hard-coding the resource defaults, make them overwritable using environment variables.
Use the env package for unmarshalling directly into a corresponding struct.
Change the fleet-manager deployment in the test environment so that slimmer central tenants are created by default.

+ Some pre-commit fixes.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/internal/dinosaur/pkg/defaults/default_central_spec.go
+++ b/internal/dinosaur/pkg/defaults/default_central_spec.go
@@ -1,28 +1,57 @@
 package defaults
 
 import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/caarlos0/env/v6"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+// CentralDefaults ...
+type CentralDefaults struct {
+	MemoryRequest resource.Quantity `env:"MEMORY_REQUEST" envDefault:"250Mi"`
+	CPURequest    resource.Quantity `env:"CPU_REQUEST" envDefault:"250m"`
+	MemoryLimit   resource.Quantity `env:"MEMORY_LIMIT" envDefault:"4Gi"`
+	CPULimit      resource.Quantity `env:"CPU_LIMIT" envDefault:"1000m"`
+}
+
 var (
-	// CentralRequestMemory ...
-	CentralRequestMemory = resource.MustParse("250Mi")
-	// CentralRequestCPU ...
-	CentralRequestCPU = resource.MustParse("250m")
-	// CentralLimitMemory ...
-	CentralLimitMemory = resource.MustParse("4Gi")
-	// CentralLimitCPU ...
-	CentralLimitCPU = resource.MustParse("1000m")
+	// Central ...
+	Central CentralDefaults
 	// CentralResources ...
+	CentralResources corev1.ResourceRequirements
+)
+
+func init() {
+	defaults := CentralDefaults{}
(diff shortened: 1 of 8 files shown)


mtesseract:
@SimonBaeumer Any update here?

internal/dinosaur/pkg/defaults/default_scanner_spec.go
```diff
@@ -1,57 +1,87 @@
...
-			corev1.ResourceMemory: ScannerDbLimitMemory,
+			corev1.ResourceCPU:    Scanner.Db.CPULimit,
+			corev1.ResourceMemory: Scanner.Db.MemoryLimit,
 		},
 	}
-)
+}
+
+// PettyPrintScannerDefaults ...
+func PettyPrintScannerDefaults() []string {
```
johannes94, line 80: Did you mean PrettyPrint... instead of `Petty`?

johannes94: Apart from that, you could define a more generic function that takes an empty interface and a label and prints any object in that fashion instead of defining a function for scanner and central. The implementation itself would not differ much: (not tested)
```suggestion
func PettyPrintScannerDefaults(obj interface{}, label string) []string {
	bytes, err := json.MarshalIndent(obj, "", "  ")
	if err != nil {
		return nil
	}
	
	return append([]string{fmt.Sprintf("%s:", label)}, strings.Split(string(bytes), "\n")...)
}
```

internal/dinosaur/pkg/defaults/default_central_spec.go
```diff
@@ -1,28 +1,57 @@
...
-			corev1.ResourceMemory: CentralLimitMemory,
+			corev1.ResourceCPU:    Central.CPULimit,
+			corev1.ResourceMemory: Central.MemoryLimit,
 		},
 	}
-)
+}
+
+// PettyPrintCentralDefaults ...
+func PettyPrintCentralDefaults() []string {
```
johannes94, line 50:
```suggestion
func PrettyPrintCentralDefaults() []string {
```

johannes94 (changes requested):
Looks good to me at first glance. I'll approve as soon as I tested this and my requested changes were addressed.

internal/dinosaur/pkg/defaults/default_scanner_spec.go
mtesseract: Thanks!

pkg/environments/environment.go
```diff
@@ -135,11 +135,23 @@ func (env *Env) CreateServices() error {
 		return fmt.Errorf("modifying configuration: %w", err)
 	}
 	glog.Info("Active defaults for new Central tenants:")
-	for _, line := range defaults.PettyPrintCentralDefaults() {
-		glog.Info("  " + line)
+	prettyPrintedDefaults, err := defaults.PrettyPrintDefaults(&defaults.Central, "CentralDefaults")
+	if err != nil {
+		glog.Errorf("Failed to pretty-print Central defaults: %v", err)
```
johannes94, line 140: Nit: The 2 error handling/logging sections, look like a lot of duplication. I think we could also do it in a function instead.

johannes94 (approved):

pkg/environments/environment.go
mtesseract: done

PR #354, opened 2022-08-31
ROX-12410: Replace dinosaur with central in metrics

kurlov:
## Description

- Replace `dinosaur` naming with `central` in metrics
- Adopt Grafana dashboard accordingly


stehessel (approved):
Looks good!

PR #361, opened 2022-09-05
Allow creating tenant resources via Helm chart

misberner:
## Description

Add a mechanism to allow creating tenant resources through a Helm chart, avoiding the need to specify these as Go structs. The chart is currently empty, I will add resources to it in a follow-up PR, and at a later point we can also add existing extra resources like routes or even the central CR to it.

The approach in this PR uses the chart only for rendering, and then creates the resources individually. For reconciliation, it merely checks if a resource of the given kind and name exists, but does not attempt to actually reconcile any diverging spec. This matches what is currently done with other resources, such as routes.

Eventually, it would be nice to use a `helm upgrade` for reconciliation, or at least have some approach that mimics what Helm does, i.e., keep a manifest of all resources rather than relying on rendering output. The reason this was not done here is the fact that the Helm action client depends on Kubernetes's `client-go`, and no the controller-runtime client. These seem to have hopelessly diverged, and it isn't possible to use the existing unit tests with the controller-runtime `fake` client when the Helm action client is used.

- [x] ~Documentation added if necessary~

## Test manual

- Unit test


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -395,6 +414,76 @@ func (r *CentralReconciler) ensureCentralCRDeleted(ctx context.Context, central
 	return false, nil
 }
 
+func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remoteCentral private.ManagedCentral) error {
```
SimonBaeumer, line 417: Can you implement a small test for this function?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -395,6 +414,76 @@ func (r *CentralReconciler) ensureCentralCRDeleted(ctx context.Context, central
...
+		}
+		err = r.client.Create(ctx, obj)
+		if err != nil && !apiErrors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create object %s/%s of type %v: %w", key.Namespace, key.Name, obj.GroupVersionKind(), err)
+		}
+	}
+	return nil
+}
+
+func (r *CentralReconciler) ensureChartResourcesDeleted(ctx context.Context, remoteCentral private.ManagedCentral) (bool, error) {
```
SimonBaeumer, line 449: Can you implement a small test for this function?

SimonBaeumer (changes requested):
The change looks great, I like using the Helm approach to render the resources.

misberner:
@SimonBaeumer test added

misberner:
@SimonBaeumer @johannes94 could you please take another look?

SimonBaeumer (approved):

PR #363, opened 2022-09-05
fix memory leak in reconciler map of fleetshard-sync

johannes94:
## Description
fleetshard-sync hold a map in memory with instances of a Reconciler for each central. When a central gets deleted and no longer appears in the response from fleet-manager's API the Reconciler instance is not deleted. This leads to a growing Map of Reconciler instances and potential out of memory issues if many central instances are created and deleted 

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/runtime/runtime.go
+++ b/fleetshard/pkg/runtime/runtime.go
@@ -85,10 +85,6 @@ func (r *Runtime) Start() error {
 			return 0, err
 		}
 
-		// This map collects all central ids in the current list, it is later used to find and delete all reconcilers of
-		// centrals that are no longer in the GetManagedCentralList
-		startedReconcile := map[string]bool{}
-
 		// Start for each Central its own reconciler which can be triggered by sending a central to the receive channel.
 		glog.Infof("Received %d centrals", len(list.Items))
 		for _, central := range list.Items {
@@ -102,9 +98,9 @@ func (r *Runtime) Start() error {
 				status, err := reconciler.Reconcile(context.Background(), central)
 				r.handleReconcileResult(central, status, err)
 			}(reconciler, central)
-			startedReconcile[central.Id] = true
 		}
-		r.deleteStaleReconcilers(startedReconcile)
+
+		r.deleteStaleReconcilers(list)
 		return r.config.RuntimePollPeriod, nil
 	}, 10*time.Minute, backoff)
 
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/runtime/runtime.go
```diff
@@ -92,8 +96,9 @@ func (r *Runtime) Start() error {
 				status, err := reconciler.Reconcile(context.Background(), central)
 				r.handleReconcileResult(central, status, err)
 			}(reconciler, central)
+			startedReconcile[central.Id] = true
 		}
-
+		r.deleteStaleReconcilers(startedReconcile)
```
misberner, line 101: You don't really need the newly introduced map, you could just make this method take in a `[]private.ManagedCentral` and pass `list.Items` here

misberner: Ah, I see, then the lookup in there would no longer be easy. I'd still say that it's cleaner to assemble the map in `deleteStaleReconcilers`

johannes94: I did it that way to don't loop over the list one additional time. But I guess this doesn't reduce runtime that much in this case. I moved it to the deleteStaleReconcilers method.

internal/dinosaur/pkg/presenters/managedcentral.go
```diff
@@ -66,10 +66,10 @@ func (c *ManagedCentralPresenter) PresentManagedCentral(from *dbapi.CentralReque
 			},
 			Auth: private.ManagedCentralAllOfSpecAuth{
 				ClientSecret: c.centralConfig.RhSsoClientSecret, // pragma: allowlist secret
-				ClientId:    c.centralConfig.RhSsoClientID,
-				OwnerOrgId:  from.OrganisationID,
-				OwnerUserId: from.OwnerUserID,
-				Issuer:      c.centralConfig.RhSsoIssuer,
+				ClientId:     c.centralConfig.RhSsoClientID,
```
johannes94, line 69: This was automatically changed by pre commit hooks.

SimonBaeumer (approved):
LGTM, thanks!

fleetshard/pkg/runtime/runtime.go
```diff
@@ -134,6 +135,21 @@ func (r *Runtime) handleReconcileResult(central private.ManagedCentral, status *
 	}
 }
 
+func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
+	// This map collects all central ids in the current list, it is later used to find and delete all reconcilers of
+	// centrals that are no longer in the GetManagedCentralList
+	centralIds := map[string]bool{}
```
misberner, line 141: Super nit: the idiomatic way of implementing a `map`-based set in go is to make the value type a `struct{}`. You'd write `centralIds[central.Id] = struct{}{}` below, and change the check in l.147 to `if _, found := centralIds[key]; !found`. Alternatively, you can use a `StringSet` from `github.com/stackrox/rox/pkg/set`.

misberner (approved):

PR #370, opened 2022-09-07
Propagate Central's RH SSO config to fleet manager

rukletsov:
## Description
Ensure Central's RH SSO config is propagated to fleet manager in E2E tests.

Three respective secrets have been added to https://vault.ci.openshift.org/ui/vault/secrets/kv/show/selfservice/rhacs-ms-e2e-tests/credentials.

## Test manual

TODO(alexr): Look at CI output to see if the secret is being read.

### DIFF
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -42,7 +42,9 @@ func NewCentralConfig() *CentralConfig {
 		CentralDomainName:                "rhacs-dev.com",
 		CentralLifespan:                  NewCentralLifespanConfig(),
 		Quota:                            NewCentralQuotaConfig(),
-		RhSsoIssuer:                      "https://sso.redhat.com/auth/realms/redhat-external",
+		RhSsoClientID:                    "secrets/central.idp-client-id",
+		RhSsoClientSecret:                "secrets/central.idp-client-secret",
+		RhSsoIssuer:                      "secrets/central.idp-issuer",
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/dinosaur/pkg/config/central.go
```diff
@@ -42,7 +42,7 @@ func NewCentralConfig() *CentralConfig {
 		CentralDomainName:                "rhacs-dev.com",
 		CentralLifespan:                  NewCentralLifespanConfig(),
 		Quota:                            NewCentralQuotaConfig(),
-		RhSsoIssuer:                      "https://sso.redhat.com/auth/realms/redhat-external",
```
dhaus67, line 45: Why remove the `RhSsoIssuer` here? We don't explicitly set this for i.e. the production environment, so let's have this as a sane default and provide the possibility to override it.

internal/dinosaur/pkg/config/central.go
```diff
@@ -42,7 +42,7 @@ func NewCentralConfig() *CentralConfig {
...
+		RhSsoClientSecretFile:            "secrets/central.idp-client-secret",
```
dhaus67, line 45: It's a bit unfortunate that we are going to have a mismatch in names, i.e. `secrets/central.idp-client-secret` for the local one / E2E env and on stage / prod its going to be `secrets/central-rhsso-client-secret/clientSecret` instead of i.e. `secrets/central-idp-client-secret`.

But, I don't think it's something we need to change right now.

dhaus67 (approved):
One question regarding the rhsso issuer field, other than that LGTM and its also successfully working now:
```
[fleet-manager] I0907 16:44:00.254116       7 central.go:79] Central Red Hat OIDC client secret is configured.
```

internal/dinosaur/pkg/config/central.go
rukletsov: It's unfortunate to have defaults like this be baked into the code. But I think you have a point that it's already like this everywhere so we won't achieve much by improving one single occurrence. Will revert.

dhaus67 (approved):
The failure in E2E doesn't look like its related to this code (seems like the operator hasn't been installed..), so LGTM!

PR #379, opened 2022-09-08
ROX-12078: Extract central error in fleetshard-sync

ivan-degtiarenko:
## Description
To make it easier to debug future issues with calls from fleetshard-sync to central, let's add logging for HTTP responses in case there's an error.

## Test manual

None


fleetshard/pkg/central/client/client.go
```diff
@@ -103,6 +107,21 @@ func (c *Client) SendRequestToCentral(ctx context.Context, requestMessage proto.
 	return nil
 }
 
+func extractCentralError(resp *http.Response) string {
+	var data map[string]string
+	if resp == nil || resp.Body == nil {
```
dhaus67, line 112: Don't think we need the `resp == nil` check here, seems redundant as we do not check it in other places (i.e. `resp.StatusCode`) and IIRC `if err == nil then resp != nil` for `httpClient.Do(req)`

fleetshard/pkg/central/client/client.go
```diff
@@ -103,6 +107,21 @@ func (c *Client) SendRequestToCentral(ctx context.Context, requestMessage proto.
...
+		return couldNotParseReason
+	}
+	err := json.NewDecoder(resp.Body).Decode(&data)
+	if err != nil {
+		return couldNotParseReason
+	}
```
dhaus67, line 118: nit:
```suggestion
	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
		return couldNotParseReason
	}
```

fleetshard/pkg/central/client/client.go
```diff
@@ -103,6 +107,21 @@ func (c *Client) SendRequestToCentral(ctx context.Context, requestMessage proto.
 	return nil
 }
 
+func extractCentralError(resp *http.Response) string {
+	var data map[string]string
```
dhaus67, line 111: Let's use a struct for this instead of a map.
In case we want to get additional info in the future we can extend it, so it's more clear reading through which values we actually care about.

fleetshard/pkg/central/client/client_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExtractCentralError(t *testing.T) {
+	// 1. Error in returned response
+	json := `{"error":"error-message"}`
+	r := ioutil.NopCloser(bytes.NewReader([]byte(json)))
```
dhaus67, line 15: nit:
```suggestion
	r := ioutil.NopCloser(strings.NewReader((json))
```

dhaus67 (approved):
Just minor comments, otherwise LGTM.

fleetshard/pkg/central/client/client.go
dhaus67: Sure thing.

dhaus67 (approved):

PR #385, opened 2022-09-08
Fix pre-commit workflow

dhaus67:
## Description

Seems like the workflow that executes the action has been broken for quite some time (i.e. its running, but the hooks will be skipped).

https://github.com/stackrox/acs-fleet-manager/commit/3776f169f22dd19bc4729bdaf76776d5835afa82 seems to have introduced this behavior.

I've reverted back the changes for now, fixed the issues that are already on main and have not been addressed, and will create a follow-up to ensure the local hook is executed within pipeline.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -19,4 +19,3 @@ jobs:
         with:
           go-version: "1.17"
       - uses: pre-commit/action@v3.0.0
-        with:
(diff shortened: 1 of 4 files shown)


dhaus67:
* main
  * **PR #385** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/385" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/385?utm_source=stack-comment).

mtodor (approved):
Nice catch! LGTM

PR #387, opened 2022-09-09
Use RHACS operator version with memory leak fix

misberner:
## Description

This changes the operator version (for quay, not marketplace) to a version that includes the memory leak fix, hopefully resulting in fewer CI flakes. Temporarily switching to quay as a source for CI, because there isn't a version with the fix on Marketplace yet.

- [x] ~Documentation added if necessary~

## Test manual

N/A

### DIFF
--- a/dev/env/defaults/00-defaults.env
+++ b/dev/env/defaults/00-defaults.env
@@ -10,7 +10,7 @@ export CLUSTER_ID_DEFAULT="1234567890abcdef1234567890abcdef"
 export CLUSTER_DNS_DEFAULT="cluster.local"
 
 export IMAGE_REGISTRY_DEFAULT="quay.io/rhacs-eng"
-STACKROX_VERSION_TAG="3.70.0-nightly-20220707"
+STACKROX_VERSION_TAG="3.72.0-nightly-20220909"
 export STACKROX_OPERATOR_VERSION_DEFAULT="${STACKROX_VERSION_TAG}"
 export CENTRAL_VERSION_DEFAULT=$(echo "$STACKROX_VERSION_TAG" | sed -e 's/0-nightly/x-nightly/;')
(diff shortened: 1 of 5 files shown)


SimonBaeumer (approved):

misberner:
It looks like I cannot use the quay catalog source in CI, but due to the challenging debuggability of OSCI haven't really figured out why that is. The IPS gets created..

mtesseract:
I can look into this.

mtesseract:
Please have a look at branch `mc/use-newer-rhacs-operator`. This includes your changes, but the interesting part is commit `0315b8290ca1f0b556de656c1313c929b9657867`, which makes sure that a OLM-specific logic is not executed in case we are running on OpenShift but using the operator from Quay.

The other fix is in commit `90ccfbdc715554658dde0fe7f10fab4e64178543`, which does two things:
There was a bug before which caused an early termination of the script by executing a command which failed under `set -e`. So, I added the `if ...; then ...; fi` wrapping to protect around this. Furthermore, the `jq` & `grep` command did not take into account that there might be multiple catalog sources providing the `rhacs-operator`, so I made sure that this only cares about the operator from Quay.

Could you try adding these two commits to your PR?

misberner:
Thanks for looking into this, Moritz! I'll give it another try.

misberner:
@mtesseract go for it, even though I imagine that with 3.72.0 out, which has the fix, it should be much easier to just use that?

mtesseract:
I'm thinking this functionality (picking a non-marketplace operator for testing purposes) would be useful in a more general sense, not specifically because of the 3.72 fix.

mtesseract:
I would say this is superseded by https://github.com/stackrox/acs-fleet-manager/pull/511.

PR #399, opened 2022-09-13
ROX-12251: Create IdP for OSD cluster

dhaus67:
## Description

Add a script that will automate the creation of the IdP for OSD clusters and configure the allowed users that will be granted administrative rights.

The IdP will use (for staging) auth.redhat.com.

The reason there needs to be a list of allowed users comparatively to a rover group is due to non-support of OCM for group syncing (Kafka does it in the same way but with a sync set, which periodically syncs the access for this 
:-)). This should be enough for us though.

For production, the same OIDC client sadly cannot be re-used. That's due to the redirect_uri not allowing wildcards. For this, we need to order another one and update this script afterwards (for now, I also added the HTPasswd 
creation automation).

## Test manual

```
# Check out acs-stage-dp-01 and clear the IdP / users.
# Run the following, ensure the IdP is created and users are added to the correct group (i.e. cluster-admins), login also works successfully:
./dp-terraform/osd-cluster-idp-setup.sh stage acs-stage-dp-01
```

### DIFF
--- a/dp-terraform/osd-cluster-idp-setup.sh
+++ b/dp-terraform/osd-cluster-idp-setup.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+set -eo pipefail
+
+# Requires: `jq`
+# Requires: BitWarden CLI `bw`
+
+if [[ $# -ne 2 ]]; then
+    echo "Usage: $0 [environment] [cluster]" >&2
+    echo "Known environments: stage prod"
+    echo "Cluster typically looks like: acs-{environment}-dp-01"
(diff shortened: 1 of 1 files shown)


dhaus67:
* main
  * **PR #399** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/399" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/399?utm_source=stack-comment).

ebensh (approved):
I'm sad that my hacking around BitWarden as our secret management / never having cycles to actually finish setting up the proper solution is now going to propagate to yet another file.

Thanks for setting this up though :) Looks good!

PR #406, opened 2022-09-15
Add git commit sha and tag to fleetshard-sync deployment

SimonBaeumer:
## Description
This PR adds the git commit SHA to the fleetshard-sync deployment.

## Test manual

 - Manual testing

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
@@ -6,6 +6,8 @@ metadata:
   labels:
     app: fleetshard-sync
     managed-service.rhacs.redhat.com/update-trigger: {{ now }}
+    managed-service.rhacs.redhat.com/git-commit-sha: {{ .Values.fleetshardSync.gitCommitSHA }}
+    managed-service.rhacs.redhat.com/git-describe: {{ .Values.fleetshardSync.gitDescribe }}
 spec:
   replicas: 1
   selector:
--- a/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
@@ -7,7 +7,7 @@ metadata:
     app: fleetshard-sync
     managed-service.rhacs.redhat.com/update-trigger: {{ now }}
     managed-service.rhacs.redhat.com/git-commit-sha: {{ .Values.fleetshardSync.gitCommitSHA }}
-    managed-service.rhacs.redhat.com/git-describe: {{ .Values.fleetshardSync.gitDescribe }}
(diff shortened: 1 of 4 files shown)


dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -99,6 +99,9 @@ case $ENVIRONMENT in
     ;;
 esac
 
+GIT_COMMIT_HASH=$(git rev-parse HEAD)
```
ebensh, line 102: Nit - please be consistent with naming (HASH vs SHA)

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: fleetshard-sync
+    "managed-service.rhacs.redhat.com/git-commit-sha": {{ .Values.fleetshardSync.gitCommitSHA }}
```
ebensh, line 8: Is this consistent with labels exported by other ACS binaries, e.g. the Central set up by ACS Operator? It would be nice to have a single key for this across all deployments in the fleet to make simpler prometheus/grafana visualization

Edit: I assume this is the edit in the central reconciler annotation key. Is there any reason to add the managed-service namespace prefix on this here and diverge from other ACS deployments?

ebensh (changes requested):
Looks good, couple questions and nits

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
SimonBaeumer: done

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
SimonBaeumer: Typically we want to distinguish between annotations on different projects. For example on StackRox `platform.stackrox.io/---` is used as a key.

SimonBaeumer: Removed the `managed-service.` to be consistent with already existent services.

ebensh (approved):

stehessel:
@SimonBaeumer I stumbled upon this change when trying to update the deployed version of fleetshard-sync. Maybe I'm missing something, but I think there is a fundamental flaw here because we add side effects to a declarative configuration. This results in a mismatch between the git hash of the actual deployed image in `fleetshardSync.image` and the hash of the HEAD on which the engineers happens to execute the script - which then ends up in `fleetshardSync.gitCommitSHA`. 

Can you maybe explain the reason behind this change in the first place? As is it seems kind of broken to me.

stehessel:
I think we can keep the git hash, but we need to make sure it actually corresponds to the one in the image.

PR #408, opened 2022-09-15
ROX-12076: probe service

stehessel:
### Summary

Design doc: https://docs.google.com/document/d/1Vk-r97VokbhsHnZKcb8nAJO0M55NO9W0xtx0AQTavhA/edit

The probe service enables blackbox monitoring for fleet manaer. During each
probe run, it attempts to

1. create a Central instance and ensure it is in `ready` state.
2. verify that the Central instance is of type `standard` and that the Central UI is reachable.
3. deprovision the Central.

Requests against fleet manager are authenticated by a Red Hat SSO service account.

The probe service may run as a one-shot command (`probe run`), in which case a single probe
is executed. If the probe aborts or fails, the service exits with exit code 1. In
addition, the probe service may run in a continuous loop (`probe start`), in which case
probe runs are executed in an endless loop. Prometheus metrics capture the results of the probes.

The probe service implements graceful shutdown, which means upon receiving an interrupt signal, it
attempts to clean up remaining resources.

### Testing

The primary unit testing is done via a custom mocked fleet manager client (see `probe/pkg/fleetmanager/client_mock.go`). I felt this was simpler than a mock directly on http level via https://pkg.go.dev/net/http/httptest.

### TODOs / follow ups:

* Deployment
  - We need to deploy the probe service to a long running cluster, such that it may probe the stage/prod fleet manager deployed in the control plane. The current plan is to deploy the probe service to our data plane clusters for two reasons: a) It is outside of the control plane network. b) We have easy control over the monitoring setup. c) Deployment is easy via Helm.
* Liveness and readiness endpoints.
* Write-ahead log for requested Central instances to clean up in the event of an ungraceful shutdown.

## Test manual

Tested probe service with local fleet manager and OSD data plane cluster.

### DIFF
--- a/dp-terraform/add-on/rhacs-terraform/api/v1alpha1/terraform_types.go
+++ b/dp-terraform/add-on/rhacs-terraform/api/v1alpha1/terraform_types.go
@@ -7,14 +7,14 @@ import (
 type Terraform struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec TerraformSpec `json:"spec,omitempty"`
-	Status TerraformStatus `json:"status,omitempty"`
+	Spec              TerraformSpec   `json:"spec,omitempty"`
+	Status            TerraformStatus `json:"status,omitempty"`
 }
 
 type TerraformSpec struct {
 	FleetshardSync *FleetshardSyncSpec `json:"fleetshardSync,omitempty"`
-	AcsOperator *AcsOperatorSpec `json:"acsOperator,omitempty"`
-	Observability *ObservabilitySpec `json:"observability,omitempty"`
+	AcsOperator    *AcsOperatorSpec    `json:"acsOperator,omitempty"`
+	Observability  *ObservabilitySpec  `json:"observability,omitempty"`
 }
 
 type TerraformStatus struct {
@@ -22,35 +22,35 @@ type TerraformStatus struct {
 }
 
 type FleetshardSyncSpec struct {
-	OcmToken string `json:"ocmToken,omitempty"`
-	FleetManagerEndpoint string `json:"fleetManagerEndpoint,omitempty"`
-	ClusterId string `json:"clusterId,omitempty"`
-	RedHatSSO *RedHatSSOSpec `json:"redHatSSO,omitempty"`
+	OcmToken             string         `json:"ocmToken,omitempty"`
+	FleetManagerEndpoint string         `json:"fleetManagerEndpoint,omitempty"`
+	ClusterId            string         `json:"clusterId,omitempty"`
+	RedHatSSO            *RedHatSSOSpec `json:"redHatSSO,omitempty"`
 }
 
 type RedHatSSOSpec struct {
-	ClientId string `json:"clientId,omitempty"`
+	ClientId     string `json:"clientId,omitempty"`
 	ClientSecret string `json:"clientSecret,omitempty"`
 }
 
 type AcsOperatorSpec struct {
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled     bool   `json:"enabled,omitempty"`
 	StartingCSV string `json:"startingCSV,omitempty"`
 }
 
 type ObservabilitySpec struct {
-	Enabled bool `json:"enabled,omitempty"`
-	Github *GithubSpec `json:"github,omitempty"`
+	Enabled       bool               `json:"enabled,omitempty"`
+	Github        *GithubSpec        `json:"github,omitempty"`
 	Observatorium *ObservatoriumSpec `json:"observatorium,omitempty"`
 }
 
 type GithubSpec struct {
 	AccessToken string `json:"accessToken,omitempty"`
-	Repository string `json:"repository,omitempty"`
+	Repository  string `json:"repository,omitempty"`
 }
 
 type ObservatoriumSpec struct {
-	Gateway string `json:"gateway,omitempty"`
+	Gateway         string `json:"gateway,omitempty"`
 	MetricsClientId string `json:"metricsClientId,omitempty"`
-	MetricsSecret string `json:"metricsSecret,omitempty"`
+	MetricsSecret   string `json:"metricsSecret,omitempty"`
 }
--- a/probe/pkg/cmd/cmd.go
+++ b/probe/pkg/cmd/cmd.go
@@ -1,0 +1,117 @@
+package cmd
+
+import (
+	"net/http"
+	"os"
+	"os/signal"
+
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/stackrox/acs-fleet-manager/probe/config"
+	"github.com/stackrox/acs-fleet-manager/probe/pkg/fleetmanager"
+	"github.com/stackrox/acs-fleet-manager/probe/pkg/metrics"
+	"github.com/stackrox/acs-fleet-manager/probe/pkg/runtime"
+	"golang.org/x/sys/unix"
+)
+
+// Command builds the root CLI command.
+func Command() *cobra.Command {
+	c := &cobra.Command{
+		SilenceUsage: true,
+		Use:          os.Args[0],
+		Long:         "Probe is a service that verifies the availability of ACS fleet manager.",
+	}
+	c.AddCommand(
+		startCommand(),
+		runCommand(),
+	)
+	return c
+}
+
+func startCommand() *cobra.Command {
+	c := &cobra.Command{
+		SilenceUsage: true,
+		Use:          "start",
+		Short:        "Start a continuous loop of probe runs.",
+		Long:         "Start a continuous loop of probe runs.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return executeProbe(true, cmd, args)
+		},
+	}
+	return c
+}
+
+func runCommand() *cobra.Command {
+	c := &cobra.Command{
+		SilenceUsage: true,
+		Use:          "run",
+		Short:        "Run a single probe run.",
+		Long:         "Run a single probe run.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return executeProbe(false, cmd, args)
+		},
+	}
+	return c
+}
+
+func executeProbe(loop bool, cmd *cobra.Command, args []string) error {
+	glog.Infof("Probe service has been started.")
+
+	config, err := config.GetConfig()
+	if err != nil {
+		return errors.Wrap(err, "Failed to load configuration")
+	}
+
+	metricsServer := metrics.NewMetricsServer(config.MetricsAddress)
+	defer cleanupMetricsServer(metricsServer)
+	go func() {
+		if err := metricsServer.ListenAndServe(); err != nil {
+			glog.Errorf("Failed to serve metrics: %v", err)
+		}
+	}()
+
+	fleetManagerClient, err := fleetmanager.New(config)
+	if err != nil {
+		return errors.Wrap(err, "Failed to initialize probe")
+	}
+	runtime, err := runtime.New(config, fleetManagerClient)
+	if err != nil {
+		return errors.Wrap(err, "Failed to initialize probe")
+	}
+	defer cleanupRuntime(runtime)
+
+	// TODO: Add server for liveness and readiness probes here.
+
+	isErr := make(chan bool, 1)
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, os.Interrupt, unix.SIGTERM)
+
+	go runtime.Run(isErr, sigs, loop)
+
+	sig := <-sigs
+	if sig != nil {
+		glog.Infof("Caught %s signal.", sig)
+		isErr <- true
+	}
+	glog.Info("Probe service has been stopped.")
+
+	returnErr := <-isErr
+	if returnErr {
+		return errors.New("Probe service failed. Exiting with status code 1.")
+	}
+	return nil
+}
+
+func cleanupRuntime(runtime *runtime.Runtime) {
+	err := runtime.Stop()
+	if err != nil {
+		glog.Errorf("Failed to close runtime: %v", err)
+	}
+}
+
+func cleanupMetricsServer(metricsServer *http.Server) {
+	if err := metricsServer.Close(); err != nil {
+		glog.Errorf("Failed to close metrics server: %v", err)
+	}
+}
--- a/probe/pkg/metrics/metrics.go
+++ b/probe/pkg/metrics/metrics.go
@@ -1,0 +1,121 @@
+package metrics
+
+import (
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	prometheusNamespace = "acs"
+	prometheusSubsystem = "probe"
+)
+
+var (
+	metrics *Metrics
+	once    sync.Once
+)
+
+// Metrics holds the prometheus.Collector instances for the probe's custom metrics
+// and provides methods to interact with them.
+type Metrics struct {
+	startedRuns            prometheus.Counter
+	successfulRuns         prometheus.Counter
+	failedRuns             prometheus.Counter
+	lastSuccessTimestamp   prometheus.Gauge
+	lastFailureTimestamp   prometheus.Gauge
+	totalDurationHistogram prometheus.Histogram
+}
+
+// Register registers the metrics with the given prometheus.Registerer.
+func (m *Metrics) Register(r prometheus.Registerer) {
+	r.MustRegister(m.startedRuns)
+	r.MustRegister(m.successfulRuns)
+	r.MustRegister(m.failedRuns)
+	r.MustRegister(m.totalDurationHistogram)
+	r.MustRegister(m.lastFailureTimestamp)
+	r.MustRegister(m.lastSuccessTimestamp)
+}
+
+// IncStartedRuns increments the metric counter for started probe runs.
+func (m *Metrics) IncStartedRuns() {
+	m.startedRuns.Inc()
+}
+
+// IncSuccessfulRuns increments the metric counter for successful probe runs.
+func (m *Metrics) IncSuccessfulRuns() {
+	m.successfulRuns.Inc()
+}
+
+// IncFailedRuns increments the metric counter for failed probe runs.
+func (m *Metrics) IncFailedRuns() {
+	m.failedRuns.Inc()
+}
+
+// SetLastSuccessTimestamp sets timestamp for the last successful probe run.
+func (m *Metrics) SetLastSuccessTimestamp() {
+	m.lastSuccessTimestamp.SetToCurrentTime()
+}
+
+// SetLastFailureTimestamp sets timestamp for the last failed probe run.
+func (m *Metrics) SetLastFailureTimestamp() {
+	m.lastFailureTimestamp.SetToCurrentTime()
+}
+
+// ObserveTotalDuration sets the total duration gauge for probe runs.
+func (m *Metrics) ObserveTotalDuration(duration time.Duration) {
+	m.totalDurationHistogram.Observe(duration.Seconds())
+}
+
+// MetricsInstance return the global Singleton instance for Metrics
+func MetricsInstance() *Metrics {
+	once.Do(initMetricsInstance)
+	return metrics
+}
+
+func initMetricsInstance() {
+	metrics = newMetrics()
+}
+
+func newMetrics() *Metrics {
+	return &Metrics{
+		startedRuns: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "started_runs",
+			Help:      "The number of started probe runs.",
+		}),
+		successfulRuns: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "successful_runs",
+			Help:      "The number of successful probe runs.",
+		}),
+		failedRuns: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "failed_runs",
+			Help:      "The number of failed probe runs.",
+		}),
+		lastSuccessTimestamp: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "last_success_timestamp",
+			Help:      "The Unix timestamp of the last successful probe run.",
+		}),
+		lastFailureTimestamp: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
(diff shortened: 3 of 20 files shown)


probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+type Runtime struct {
+	config             *config.Config
+	fleetManagerClient fleetmanager.Client
+	request            *public.CentralRequestPayload
+	centralState       *public.CentralRequest
+}
+
+// New creates a new runtime.
+func New(config *config.Config, fleetManagerClient fleetmanager.Client) (*Runtime, error) {
+	centralName, err := newCentralName()
```
misberner, line 32: I don't have anything too concrete in mind, but I'm sure in the future someone _might_ want to run several probes in parallel, and in this case, the probes have to have some names to distinguish them (e.g., pod names). I think that rather than calling the centrals `probe-<random>`, they should be called `probe-<probe name>-<random>`.

I also think that it would make sense to generate a new name in each iteration

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+			isDone <- true
+		}()
+
+		select {
+		case <-ctxTimeout.Done():
+			glog.Errorf("Probe run timed out: %v", ctxTimeout.Err())
+			metrics.MetricsInstance().IncFailedRuns()
+			if loop {
+				time.Sleep(r.config.RuntimeRunWaitPeriod)
```
misberner, line 78: Generally speaking, avoid `time.Sleep` in production code, as it is not sensitive to context cancellations. You can use `concurrency.WaitWithTimeout(ctxTimeout, r.config.RuntimeRunWaitPeriod)`, where a `true` return value indicates that `ctxTimeout` was cancelled.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	glog.Infof("Deletion of Central instance %s requested.", r.centralState.Id)
+	if err != nil {
+		return errors.Wrapf(err, "Deletion of Central instance %s failed", r.centralState.Id)
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusAccepted {
+		return errors.Errorf("Deletion request to Central instance %s was not accepted: %s.", r.centralState.Id, response.Status)
+	}
+
+	err = r.ensureCentralState(constants.CentralRequestStatusDeprovision.String())
```
misberner, line 197: Should this function not wait (with a certain timeout) until the central is actually deleted?

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	err = r.deleteCentral(ctx)
+	return err
+}
+
+// Stop the probe run.
+// TODO: Add write ahead log to clean up after ungraceful restarts.
+func (r *Runtime) Stop() error {
+	if r.centralState == nil {
```
misberner, line 119: `r.centralState` is written to within `Run()`, yet there is no explicit synchronization here, so it's not legal to call `Stop()` while `Run()` is running - which seems to defeat its purpose. In order to stop an asynchronous goroutine, you must use channels (or a context, a `concurrency.Signal` or something like that).

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+
+// totalDurationTimer returns a function that measures the elapsed time between the
+// call to the timer and the call to the returned function. The result is tracked
+// as a Prometheus metric.
+func totalDurationTimer() func() {
+	start := time.Now()
+	return func() {
+		metrics.MetricsInstance().ObserveTotalDuration(time.Since(start))
+	}
+}
```
misberner, line 139: Reading the callsite `defer totalDurationTimer()()` makes it hard to see what's going on, both because the name isn't very descriptive, and also because of the double function call. I would propose to change it to
```suggestion
func recordElapsedTime(start time.Time) {
	metrics.MetricsInstance().ObserveTotalDuration(time.Since(start))
}
```
an then call it via `defer recordElapsedTime(time.Now())` (note that arguments to `defer`ed functions are evaluated eagerly)

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	glog.Infof("Creation of Central instance requested.")
+	if err != nil {
+		return errors.Wrap(err, "Creation of Central instance failed")
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusAccepted {
+		return errors.Errorf("Creation request to Central instance was not accepted: %s.", response.Status)
+	}
+
+	r.centralState = &central
```
misberner, line 153: I think setting a `struct` field to an intermediate state is dangerous, given the number of goroutines. I haven't gone through that exercise, but I'm pretty sure one could create a race condition from that (note that a program is only concurrency-correct if its correct even when the scheduler is arbitrarily malicious, i.e., only explicit synchronizations matter, no real-time considerations, as a scheduler could keep a goroutine paused at an arbitrary point for an arbitrary amount of time).

I think it would be better as well as feasible to just pass the central state directly to the respective methods (you do need it out-of-band to realize `Stop`, but as written above, this will need to be changed anyway to just trigger a signal/write to a channel that is then handled in the main loop, where the local `centralState` variable is available if an iteration is currently in progress).

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	return fmt.Sprintf("probe-%s", rndString), nil
+}
+
+func (r *Runtime) ensureCentralState(targetState string) error {
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), r.config.RuntimePollTimeout)
+	defer cancel()
+
+	isDone := make(chan bool, 1)
+
+	go r.pollCentral(ctxTimeout, isDone, targetState)
```
misberner, line 218: If you rewrite `pollCentral` to be fully sensitive to context cancellations (and that is possible), you do not need to run it in a goroutine.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return "", errors.Wrapf(err, "Reading random bytes for unique central name")
+	}
+	rndString := hex.EncodeToString(rnd)
+
+	return fmt.Sprintf("probe-%s", rndString), nil
+}
+
+func (r *Runtime) ensureCentralState(targetState string) error {
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), r.config.RuntimePollTimeout)
```
misberner, line 213: Why do you use `context.Background()` here, given that at most callsites, there's a non-background parent context available that should be threaded through

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+
+	select {
+	case <-ctxTimeout.Done():
+		return errors.Wrapf(ctxTimeout.Err(), "Central instance %s did not reach %s state", r.centralState.Id, targetState)
+	case <-isDone:
+		return nil
+	}
+}
+
+func (r *Runtime) pollCentral(ctx context.Context, isDone chan bool, targetState string) error {
```
misberner, line 228: Just a stylistic remark that will likely be obsolete after addressing comments: when passing channels between functions, it's common to be as specific as possible about the operations that will happen on them. Since this method only writes to the channel but will never read from it, one would write this as
```suggestion
func (r *Runtime) pollCentral(ctx context.Context, isDone chan<- bool, targetState string) error {
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+		defer response.Body.Close()
+		if response.StatusCode != http.StatusOK {
+			glog.Warningf("Central instance %s did not respond with status OK: %s.", r.centralState.Id, response.Status)
+			continue
+		}
+
+		r.centralState = &central
+		if r.centralState.Status == targetState {
+			glog.Infof("Central instance %s is in `%s` state.", r.centralState.Id, targetState)
+			isDone <- true
```
misberner, line 244: Note that generally speaking, unguarded (without an enclosing `select` with other alternative `case`s) writes to channel are deadlock-prone and should be avoided. They are only safe in limited circumstances, mostly when using buffered channels. The channel here is buffered, but it is created as such in a different function, and this method contains no method comment indicating that the channel _must_ be buffered.

misberner: Concretely, if the channel wasn't buffered, and you would hit the `<-ctxTimeout.Done()` case in `ensureCentralState`, this goroutine would block forever.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	return &Runtime{
+		config:             config,
+		fleetManagerClient: fleetManagerClient,
+		request:            request,
+	}, nil
+}
+
+// Run sets up time outs and metrics, and then executes either a single probe run
+// or a continuous loop of probe runs.
+func (r *Runtime) Run(isErr chan bool, sigs chan os.Signal, loop bool) {
```
misberner, line 53: I think this method signature unnecessarily breaks multiple aspects of abstraction:
1. `isErr` only seems to be used when `loop` is `false`, so maybe the non-`loop` case should be a separate method.
2. However, I do think that the loop case should not be handled at all. This type should only offer a method for a one-shot execution. Rather than weaving loop-or-not case distinction throughout this code, the loop execution should be built _on top of_ the one-shot execution.
3. `isErr` is a boolean, but what is the concrete error? With the above suggestion, you could rewrite this as `RunSingle(...) error`.
4. It seems that `sigs chan os.Signal` is overloaded/abused. You call `signal.Notify` on it. That instructs the Go runtime to write certain signals to this channel. However, in this method, you don't care about these signals - you never read from the channel. But you _close_ it when you're done. This is not what the channel is intended for, and could also lead to crashes: writing to a closed channel causes a panic. `signal.Notify` will still do what it's supposed to do after you close the channel. If a signal is delivered just after that happens, the Go runtime will attempt to write the received signal to the closed channel, which will result in a panic.

Concrete suggestions:
- Remove the `loop bool` parameter (and rename to `RunSingle`). Looping should be the responsibility of the caller (i.e., `pkg/cmd`), or, if you think that's adequate, a separate `RunLoop` method that calls `RunSingle`.
- Change this method (`RunSingle`) to return a concrete `error`.
- Take `sigs` and any reference to `os.Signal` out of the picture. Instead, pass a `ctx context.Context` as a general purpose cancellation mechanism. The caller can then cancel the context and call if a signal is received, a timeout expires, anything.

The one thing to consider, though, is that contexts typically don't go well together with clean shutdowns. A cancelled context basically instructs the method to return ASAP. Therefore, I think that cancellation on `SIGINT` or `SIGTERM` should still invoke `Stop()` (or `GracefulStop()`) rather than just cancelling the context.

probe/pkg/cmd/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	runtime, err := runtime.New(config, fleetManagerClient)
+	if err != nil {
+		return errors.Wrap(err, "Failed to initialize probe")
+	}
+	defer cleanupRuntime(runtime)
+
+	// TODO: Add server for liveness and readiness probes here.
+
+	isErr := make(chan bool, 1)
```
misberner, line 86: There are several issues with the asynchronous nature of this code. Please see my comments in `runtime.go`.

probe/pkg/cmd/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/spf13/cobra"
+	"github.com/stackrox/acs-fleet-manager/probe/config"
+	"github.com/stackrox/acs-fleet-manager/probe/pkg/fleetmanager"
+	"github.com/stackrox/acs-fleet-manager/probe/pkg/metrics"
+	"github.com/stackrox/acs-fleet-manager/probe/pkg/runtime"
+	"golang.org/x/sys/unix"
+)
+
+// Command builds the root CLI command.
+func Command() *cobra.Command {
```
misberner, line 19: `pkg/` is typically used for shared library code. The `cmd` package does not seem to fit that bill.

Also, per the [standard Go project layout](https://github.com/golang-standards/project-layout), the name `cmd` is commonly reserved for a _sibling_ directory of `pkg` that contains one directory for each binary. I.e., if you want to have `probe/pkg`, one would typically expect the `package main` in `probe/cmd/probe`.

misberner (changes requested):
(note: lots of comments, I would recommend to read them all first before starting to implement any changes)

This is a great start! However, especially in the `runtime` implementation, I think there is a bit of restructuring that needs to happen to make it actually concurrency-correct (more below).

First thing, though, is that I would like this PR to be split into several smaller PRs, this will make the review process as well as iterating based on review comments much more tractable. My suggestion is:
- Separate PR for stylistic/formatting changes (`terraform_types.go` and `tools.go`). For smaller PRs I find it okay to sneak it in, but for PRs of that size (where scrolling in Chrome becomes noticeably slow for me), any kind of avoidable noise should be avoided.
- One PR for the main binary + Makefile stuff + runtime logic.
- One follow-up PR for the prometheus metrics

You could split it up even further. For example, main package/cobra scaffolding + makefile targets with noop logic firsts, then the runtime logic. Or a separate PR for the metrics registration side, and then one following up on the runtime logic PR to actually populate the metrics. However, no need to go to the extreme; as long as you can get to a size of not more than 500 lines (w/o generated code) per PR, I'm content.

The runtime logic is where I mostly see the issues. The use of contexts is somewhat inconsequential, and you definitely use more goroutines than would be needed. My suggestion would be to attempt to rewrite this with _as little concurrency as possible_. Most of things should happen sequentially in the "main" goroutine (i.e., the one in which `Run()` is called), and then you'll need some mechanism to asynchronously stop the execution. I've commented with more details in the `runtime.go` file.

dp-terraform/add-on/rhacs-terraform/api/v1alpha1/terraform_types.go
```diff
@@ -7,50 +7,50 @@ import (
 type Terraform struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec TerraformSpec `json:"spec,omitempty"`
-	Status TerraformStatus `json:"status,omitempty"`
+	Spec              TerraformSpec   `json:"spec,omitempty"`
```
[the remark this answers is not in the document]
stehessel: This is just the result of a `make code/fix`. Yeah I can separate that out and add it to the pre-commit hook and CI.

probe/pkg/runtime/runtime.go
misberner: Another thought: somewhere around here, you mention a write-ahead log for better cleanup. To be honest, that seems an unnecessarily complex solution to me. If you make the probe name part of the central name, a probe can easily discover "stray" central instances belonging to it that should be deleted. And even if a probe was abruptly deleted and is not coming back (at least not with that name), a probe supervisor could look at the list of running probes (with their names) and delete any orphaned central instances. And if the list of running probe names cannot easily be obtained, one could even (hex-)encode a creation timestamp in the central name, and have a janitor-like job delete centrals that are more than x minutes old.

A WAL would IMHO only make sense if the probe did something with a system that was _not_ declarative in nature.

stehessel:
Thank you Malte for the extensive review. This kind of review is quite valuable to me and probably others, so great job 👍 . Let me try to follow your recommendations starting with the PR break up. I'll probably keep this one for reference and start new ones.

probe/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,121 @@
...
+func (m *Metrics) SetLastFailureTimestamp() {
+	m.lastFailureTimestamp.SetToCurrentTime()
+}
+
+// ObserveTotalDuration sets the total duration gauge for probe runs.
+func (m *Metrics) ObserveTotalDuration(duration time.Duration) {
+	m.totalDurationHistogram.Observe(duration.Seconds())
+}
+
+// MetricsInstance return the global Singleton instance for Metrics
```
dhaus67, line 71:
```suggestion
// MetricsInstance return the global Singleton instance for Metrics.
```

probe/pkg/metrics/server.go
```diff
@@ -0,0 +1,27 @@
+package metrics
+
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// NewMetricsServer returns the metrics server
```
dhaus67, line 10:
```suggestion
// NewMetricsServer returns the metrics server.
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+// Stop the probe run.
+// TODO: Add write ahead log to clean up after ungraceful restarts.
+func (r *Runtime) Stop() error {
+	if r.centralState == nil {
+		return nil
+	}
+	ctx := context.Background()
+	_, err := r.fleetManagerClient.DeleteCentralById(ctx, r.centralState.Id, true)
+	if err != nil {
+		return errors.Wrapf(err, "Clean up: Deletion of Central instance %s failed", r.centralState.Id)
```
dhaus67, line 125:
```suggestion
		return errors.Wrapf(err, "clean up: deletion of Central instance %s failed", r.centralState.Id)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+		metrics.MetricsInstance().ObserveTotalDuration(time.Since(start))
+	}
+}
+
+// Create a Central and verify that it transitioned to 'ready' state.
+func (r *Runtime) createCentral(ctx context.Context) error {
+	central, response, err := r.fleetManagerClient.CreateCentral(ctx, true, *r.request)
+	glog.Infof("Creation of Central instance requested.")
+	if err != nil {
+		return errors.Wrap(err, "Creation of Central instance failed")
```
dhaus67, line 146:
```suggestion
		return errors.Wrap(err, "creation of Central instance failed")
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusAccepted {
+		return errors.Errorf("Creation request to Central instance was not accepted: %s.", response.Status)
```
dhaus67, line 150:
```suggestion
		return errors.Errorf("creation request to Central instance was not accepted: %s", response.Status)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	r.centralState = &central
+	err = r.ensureCentralState(constants.CentralRequestStatusReady.String())
+	return err
+}
+
+// Verify that the Central instance has the expected properties and that the
+// Central UI is reachable.
+func (r *Runtime) verifyCentral(ctx context.Context) error {
+	if r.centralState.InstanceType != fleetmanager.StandardInstanceType {
+		return errors.Errorf("Central has wrong instance type. Expected %s, got %s.", fleetmanager.StandardInstanceType, r.centralState.InstanceType)
```
dhaus67, line 162:
```suggestion
		return errors.Errorf("central has wrong instance type. Expected %s, got %s", fleetmanager.StandardInstanceType, r.centralState.InstanceType)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	err := r.pingURL(r.centralState.CentralUIURL)
+	return err
+}
+
+func (r *Runtime) pingURL(url string) error {
+	request, err := http.NewRequest("HEAD", url, nil)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to create request for Central UI %s", r.centralState.Id)
```
dhaus67, line 172:
```suggestion
		return errors.Wrapf(err, "failed to create request for Central UI %s", r.centralState.Id)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	response, err := httpclient.HTTPClient.Do(request)
+	if err != nil {
+		return errors.Wrapf(err, "Central UI of instance %s not reachable", r.centralState.Id)
```
dhaus67, line 176:
```suggestion
		return errors.Wrapf(err, "central UI of instance %s not reachable", r.centralState.Id)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		return errors.Errorf("Central UI %s did not respond with status OK. Got: %s", r.centralState.Id, response.Status)
```
dhaus67, line 180:
```suggestion
		return errors.Errorf("central UI %s did not respond with status OK. Got: %s", r.centralState.Id, response.Status)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	return nil
+}
+
+// Delete the Central instance and verify that it transitioned to 'deprovision' state.
+func (r *Runtime) deleteCentral(ctx context.Context) error {
+	response, err := r.fleetManagerClient.DeleteCentralById(ctx, r.centralState.Id, true)
+	glog.Infof("Deletion of Central instance %s requested.", r.centralState.Id)
+	if err != nil {
+		return errors.Wrapf(err, "Deletion of Central instance %s failed", r.centralState.Id)
```
dhaus67, line 190:
```suggestion
		return errors.Wrapf(err, "deletion of Central instance %s failed", r.centralState.Id)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusAccepted {
+		return errors.Errorf("Deletion request to Central instance %s was not accepted: %s.", r.centralState.Id, response.Status)
```
dhaus67, line 194:
```suggestion
		return errors.Errorf("deletion request to Central instance %s was not accepted: %s", r.centralState.Id, response.Status)
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+
+	err = r.ensureCentralState(constants.CentralRequestStatusDeprovision.String())
+	return err
+}
+
+func newCentralName() (string, error) {
+	rnd := make([]byte, 8)
+	_, err := rand.Read(rnd)
+	if err != nil {
+		return "", errors.Wrapf(err, "Reading random bytes for unique central name")
```
dhaus67, line 205:
```suggestion
		return "", errors.Wrapf(err, "reading random bytes for unique central name")
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,249 @@
...
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), r.config.RuntimePollTimeout)
+	defer cancel()
+
+	isDone := make(chan bool, 1)
+
+	go r.pollCentral(ctxTimeout, isDone, targetState)
+
+	select {
+	case <-ctxTimeout.Done():
+		return errors.Wrapf(ctxTimeout.Err(), "Central instance %s did not reach %s state", r.centralState.Id, targetState)
```
dhaus67, line 222:
```suggestion
		return errors.Wrapf(ctxTimeout.Err(), "central instance %s did not reach %s state", r.centralState.Id, targetState)
```

PR #410, opened 2022-09-15
Fix IdP setup

dhaus67:
## Description

There was an issue related to the OIDC IdP setup, where the username-claim was not set. This lead to the user ID being a generated ID, hence the user did not have the correct permissions when logging in.

Besides setting the claim, I've also done the following:
- Run shellcheck against the script.
- Ensure the `acsms-admin` user is created in staging as well as HTPasswd provider.


dhaus67:
* main
  * **PR #410** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/410" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/410?utm_source=stack-comment).

docs/development/setup-osd-cluster-idp.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. `jq` installed.
+2. `bw` - BitWarden CLI installed.
+3. `ocm` installed.
+
+Additionally, you will require access to our BitWarden vault.
+
+## Creating the IdPs
+
+For each OSD cluster, you can create IdPs that will allow login to the OpenShift Console and map your user to a specific
```
ebensh, line 13: Nit: manual line breaks seem unnecessary

docs/development/setup-osd-cluster-idp.md
```diff
@@ -0,0 +1,103 @@
...
+group within the cluster, providing i.e. administrative rights.
+
+Based on the environment (staging or prod), the following choices exist for IdPs:
```
ebensh, line 16: would s/staging/stage/ for consistency

ebensh, line 16: It looks from the description below like _both_ are set up, so instead of "choices exist" would say "the following IdPs will be created:" or similar

docs/development/setup-osd-cluster-idp.md
```diff
@@ -0,0 +1,103 @@
...
+- stage:
+  1. Fetch required credentials from BitWarden.
+  2. Create the OIDC IdP for the cluster.
+  3. Create the user <-> group mapping for cluster-admins.
+  4. Create the HTPasswd IdP for the cluster.
+  5. Create the `acsms-admin` user and map it to cluster-admins group.
+- prod:
+  1. Fetch required credentials from BitWarden.
+  2. Create the HTPasswd IdP for the cluster.
+  3. Create the `acsms-admin` user and map it to cluster-admins group.
```
ebensh, line 41: Prod uses `acsms-prod-admin` currently.

I think I would, if possible, change the stage environment's HTPasswd admin account name to `acsms-stage-admin` (or change both to `acsms-admin-<stage|prod>`). This would be a change in the script, BitWarden, and worth announcing to the group. Thoughts? Good time / bad time / not worth it / just use one name instead and deal with BitWarden confusion? (but definitely not the same password ;))

docs/development/setup-osd-cluster-idp.md
```diff
@@ -0,0 +1,103 @@
...
+There's two options for deleting the user mappings from console.openshift.com:
+- deleting manually within the UI
+- calling the following command: `ocm delete user --group=cluster-admins <user id>`
+
+Additionally, you will have to clear the users within the OSD cluster.
+
+You can do so by running the following:
+```shell
+# Login to the cluster. This will automatically set the correct context for oc.
+ocm cluster login <cluster name>
```
ebensh, line 61: 1. Does this achieve the same effect as logging into the cluster and using the UI to get an oc login token `oc login --token=sha256~...`?
2. If so, what are you actually logged in to the cluster /as/? You didn't specify here an admin account, so when you run `oc get identity`, what authn/authz do you have?

ebensh: Not directly relevant to this script/PR, as long as it works - but this would simplify the terraform script too, obviating the need to login to the cluster's console and get a login token

docs/development/setup-osd-cluster-idp.md
```diff
@@ -0,0 +1,103 @@
...
+oc delete identity <identities>
+
+# Ensure the users are also cleaned up
+oc get users
+
+# Delete existing users
+oc delete users <user ids>
+```
+
+## Troubleshooting
```
ebensh, line 77: I just want to say I love this section. Thank you!

ebensh (changes requested):

docs/development/setup-osd-cluster-idp.md
dhaus67: 1. Yes, achieves the same thing as `oc login ...`. The ocm command will _actually_ invoke this, so it's a wrapper around it.
2. When running `ocm cluster login`, you will be asked to provide the HTPasswd credentials for logging in. This means you will be then afterwards be logged in as `acsms-(stage|prod)-admin` with the respective rights (cluster-admin).


Agreed, there's also another version of how to achieve this by doing the following:
```
export CLUSTER_API_URL=$(ocm get "/api/clusters_mgmt/v1/clusters/${CLUSTER_ID}" | jq '.api.url' --raw-output)
oc login "${CLUSTER_API_URL}" --username=${OSD_ADMIN_USER} --password=${OSD_ADMIN_PASS}
```

docs/development/setup-osd-cluster-idp.md
dhaus67: Seems consistent to go for `acsms-(stage|prod)-admin`.
We only need to change the BitWarden values for stage and re-create the HTPasswd IdP / user, since the script will use the username value within BitWarden.

I'll do that before merging + announce it in the group (Friday afternoon seems anyways like a good time to do this as people shouldn't be disrupted in their workflow) :-)

dhaus67: Update: update the BitWarden username + sent announcement within the group.

ebensh:
Thanks for addressing all comments, LGTM++.

@SimonBaeumer did bring up the question of whether we want account names written in public repo but I think we've addressed offline that we're okay with it for the time being.

ebensh (approved):

PR #411, opened 2022-09-15
ROX-12650: Update helm resources in central reconciler

SimonBaeumer:
## Description
Currently no helm resources were updated, thus the egress proxy remained in failed state in the staging environment.

## Test manual

Only did manual testing so far, because I'm not sure yet if the implemented behaviour is acceptable.

Set up a local cluster with `colima`, then:
- ran the operator outside the cluster (with `make install run` from `stackrox/operator`)
- ran the bootstrap script: `INSTALL_OPERATOR=NO INSTALL_OLM=NO ./dev/env/scripts/bootstrap.sh`
- started the services with `INSTALL_OPERATOR=NO ./dev/env/scripts/up.sh`
- make a central request to fleet-manager with `./scripts/create-central.sh`

Then to test that updating of the chart resources works:
- added annotation to `egress-proxy`
- simulated an update of `fleetshard-sync` with `kubectl -n acsms rollout restart deployment fleetshard-sync`
- checked that `egress-proxy` is updated (via logs, also checked that the annotation that I added was removed)

So basically the current behaviour is as follows:
- chart resources are embedded in the `fleetshard-sync` binary, so they can only change when a new binary is deployed
- chart resources are installed on the first run of `Reconcile`, or, if they exist, they are updated. After the central is ready, they are no longer updated, unless the Central objects changes. They are also updated if `fleetshard-sync` is restarted (so in case a new version of `fleetshard-sync` is installed, the updated chart resources *will* be installed - which is what I think we're trying to achieve) 
- changes (manual or made by other controllers) to the chart resources will be lost if `fleetshard-sync` is updated or restarted, but also if the Central is changed (is this acceptable for now?)
- I think it's not likely to have hot update loops right now, because if the managed Central is ready and not changed (see line 84) the `Reconcile` function just exits


misberner:
Right, indeed I wrote this in the PR description of https://github.com/stackrox/acs-fleet-manager/pull/361:
> For reconciliation, it merely checks if a resource of the given kind and name exists, but does not attempt to actually reconcile any diverging spec. This matches what is currently done with other resources, such as routes.

Of course, my thinking was that the better approach would need to be implemented once we actually need to change the resources, not merely fix a mundane bug. Meh! I think however that in the short-term, manually bouncing stuff is better. I'll comment above regarding where I see problems with the current extended-yet-still-simple approach and how it could be done better.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -5,13 +5,12 @@ import (
...
 	"context"
 	"fmt"
-	"strconv"
-	"sync/atomic"
-
 	"github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/charts"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"strconv"
```
misberner, line 12: import sorting regression?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -462,6 +461,13 @@ func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remot
 		if err != nil && !apiErrors.IsAlreadyExists(err) {
 			return fmt.Errorf("failed to create object %s/%s of type %v: %w", key.Namespace, key.Name, obj.GroupVersionKind(), err)
 		}
+
+		if apiErrors.IsAlreadyExists(err) {
```
misberner, line 465: This won't work because of the `continue` in l.455. You'll have to remove that one first.

Also, you can only update if the resource version is matching. YOu need to do something like
```
obj.SetResourceVersion(out.GetResourceVersion())
```
before calling the `Update`.

Finally, we need to make sure that we don't run into hot update loops, where some controller sets a field, and the `Update` then removes that field value. I know that `status` is specifically protected, so calling `Update` with an empty `status` field of the object doesn't matter. Same applies for OpenAPI schema defaults and maybe even admission controller defaults. So this probably should work in most cases. However, I think I would overall prefer if we determined whether an `Update` is necessary by storing some annotation containing a hash of the manifest we wanted to apply. That doesn't prevent manual modifications, but the operator isn't any different in that regard.

vladbologa:
I made another PR to continue work on this task: https://github.com/stackrox/acs-fleet-manager/pull/465

vladbologa:
Replaced by #465

PR #413, opened 2022-09-15
Deploy a non-released operator version with Helm

dhaus67:
## Description

This adds preparation to deploy an upstream version of the operator to a dataplane cluster using the `dp-terraform/helm/terraform-cluster.sh` script.

It's not set in stone whether we want to do it _exactly_ in this way, but at the very least this works in a rough state and we can remove it once we have found a better way to deploy the operator.

In detail, the following has been done based on [the existing documentation of adding an upstream operator](https://github.com/stackrox/acs-fleet-manager/blob/main/docs/development/setup-developer-osd-cluster.md#prepare-cluster-for-rhacs-operator):
- Add scripts for adding the pull secret for `quay.io/rhacs-eng` to the global pull secret of the OSD cluster (using the credentials within BitWarden).
- Expose a new helm setting `acsOperator.upstream.use`.
- If `acsOperator.upstream.use` create a new Catalog source using the operator index image within `quay.io/rhacs-eng`.

The code within the `terraform-cluster.sh` is commented out for now, since we don't plan to use it immediately. 
Once we decide to use upstream images (and don't have another way to do so), we can uncomment them and commit them, as this is currently the workflow within the deployment / terraforming of clusters.

## Test manual

```
# Create a dataplane cluster following the instructions within [this documentation](https://github.com/stackrox/acs-fleet-manager/blob/main/docs/development/setup-developer-osd-cluster.md#prepare-cluster-for-rhacs-operator)
# (Note: skip deploying the catalog source, instead after the cluster is provisioned and the IdP is created, go straight for the helm installation)

# For the helm installation, use the following to terraform the cluster
helm upgrade --install rhacs-terraform \
  --namespace "${NAMESPACE}" \
  --set fleetshardSync.authType="STATIC_TOKEN" \
  --set fleetshardSync.image="${FLEET_MANAGER_IMAGE}" \
  --set fleetshardSync.fleetManagerEndpoint="${FLEET_MANAGER_ENDPOINT}" \
  --set fleetshardSync.staticToken="${STATIC_TOKEN}" \
  --set fleetshardSync.clusterId="${CLUSTER_ID}" \
  --set acsOperator.enabled=true \
  --set acsOperator.upstream.use=true \
  --set acsOperator.source="rhacs-operators" \
  --set acsOperator.version="v3.72.0-35-gf72be438ec" \
  --set logging.enabled=false \
  --set observability.enabled=false ./dp-terraform/helm/rhacs-terraform

# The install plan is currently configured as "Manual", so you have to approve it.

# After approval, fleetshard-sync and the rhacs-operator-manager should show up in the namespace you have chosen
oc get pods -n "${NAMESPACE}"

# Create a central using the following curl script, verify that the provisioning succeeds:
curl -X POST -H "Authorization: Bearer ${STATIC_TOKEN}" "http://localhost:8000/api/rhacs/v1/centrals?async=true" -d "{\"name\":\"test-central\",\"cloud_provider\":\"aws\",\"region\":\"us-east-1\",\"multi_az\":true}"
```

### DIFF
--- a/docs/development/setup-developer-osd-cluster.md
+++ b/docs/development/setup-developer-osd-cluster.md
@@ -80,17 +80,22 @@ export OSD_ADMIN_USER="osd-admin"
 export OSD_ADMIN_PASS=$(date | md5)
 echo $OSD_ADMIN_PASS > ./tmp-osd-admin-pass.txt
 
-echo "{\"htpasswd\":{\"password\":\"${OSD_ADMIN_PASS}\",\"username\":\"${OSD_ADMIN_USER}\"},\"login\":true,\"mapping_method\":\"add\",\"name\":\"osd-htpasswd\",\"type\":\"HTPasswdIdentityProvider\"}" > ./tmp-osd-htpasswd-body.txt
-ocm post "/api/clusters_mgmt/v1/clusters/${CLUSTER_ID}/identity_providers" --body ./tmp-osd-htpasswd-body.txt
+ocm create idp \
+  --cluster "${CLUSTER_ID}" \
+  --type htpasswd \
+  --name HTPasswd \
+  --username "${OSD_ADMIN_USER}" \
+  --password "${OSD_ADMIN_PASS}"
 
-echo "{\"id\":\"${OSD_ADMIN_USER}\"}" > ./tmp-osd-cluster-admins-body.txt
-ocm post "/api/clusters_mgmt/v1/clusters/${CLUSTER_ID}/groups/cluster-admins/users" --body ./tmp-osd-cluster-admins-body.txt
+ocm create user \
+  --group cluster-admins \
+  --cluster "${CLUSTER_ID}" \
+  "${OSD_ADMIN_USER}"
 ```
 
-3. Login to OSD cluster with `oc` command
+3. Login to OSD cluster with `ocm` command. This will automatically set the correct context for the `oc` command.
 ```
-export CLUSTER_API_URL=$(ocm get "/api/clusters_mgmt/v1/clusters/${CLUSTER_ID}" | jq '.api.url' --raw-output)
-oc login "${CLUSTER_API_URL}" --username=${OSD_ADMIN_USER} --password=${OSD_ADMIN_PASS}
+ocm cluster login "${CLUSTER_ID}"
 ```
 If login step fails, it can be the case that previously created auth provider and user are not applied yet on the cluster. You can wait few seconds and try again.
 
@@ -166,7 +171,6 @@ You should be able to see `rhacs-operators` pod running.
 export STATIC_TOKEN=$(bw get item "64173bbc-d9fb-4d4a-b397-aec20171b025" | jq '.fields[] | select(.name | contains("JWT")) | .value' --raw-output)
 
 export FLEET_MANAGER_IMAGE=quay.io/app-sre/acs-fleet-manager:main
-export STARTING_CSV="rhacs-operator.v${RHACS_OPERATOR_CATALOG_VERSION}"
 ```
 
 9. Prepare namespace
@@ -201,8 +205,8 @@ helm upgrade --install rhacs-terraform \
   --set fleetshardSync.clusterId="${CLUSTER_ID}" \
   --set acsOperator.enabled=true \
   --set acsOperator.source="${RHACS_OPERATOR_CATALOG_NAME}" \
-  --set acsOperator.startingCSV="${STARTING_CSV}" \
-  --set logging.enabled=false
+  --set logging.enabled=false \
+  --set acsOperator.version="v${RHACS_OPERATOR_CATALOG_VERSION}" \
   --set observability.enabled=false ./dp-terraform/helm/rhacs-terraform
 ```
 
--- a/dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
@@ -16,5 +16,19 @@ spec:
   name: rhacs-operator
   source: {{ .Values.acsOperator.source }}
   sourceNamespace: {{ .Values.acsOperator.sourceNamespace }}
-  startingCSV: {{ .Values.acsOperator.startingCSV }}
+  startingCSV: rhacs-operator.{{ .Values.acsOperator.version }}
+---
+{{- if .Values.acsOperator.upstream.use }}
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: {{ .Values.acsOperator.source }}
+  namespace: openshift-marketplace
+spec:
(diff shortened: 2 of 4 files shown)


dhaus67:
* main
  * **PR #413** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/413" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/413?utm_source=stack-comment).

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -36,7 +36,10 @@ acsOperator:
   enabled: false
   source: redhat-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: rhacs-operator.v3.70.0
+  version: v3.70.0
+  # This setting specifies whether to use a upstream operator to test unreleased versions of ACS.
+  upstream:
```
stehessel, line 41: Why not just use `upstream: false`? Do you have any other parameters in mind?

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -99,6 +99,17 @@ case $ENVIRONMENT in
     ;;
 esac
 
+
+## Uncomment this section if you want to deploy an upstream version of the operator.
```
stehessel, line 103: Not a fan of having these commands in our deployment, I think we should do away with the scripts all together and use a purely declarative Helm setup. That said, it's better than having the option to deploy upstream releases I suppose. If we do decide to run that in production we can still improve it.

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -36,7 +36,10 @@ acsOperator:
   enabled: false
   source: redhat-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: rhacs-operator.v3.70.0
```
stehessel, line 39: 👍

stehessel (approved):

dp-terraform/helm/rhacs-terraform/values.yaml
dhaus67: I initially had, but doesn't seem like we need the nesting for now, reverted it to simply `upstream`.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
dhaus67: Fully agree, should be pure helm going forward. I would expect us, if we _have_ to deploy upstream and are certain we need it frequently we hopefully have a better solution instead of this script.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -99,6 +99,17 @@ case $ENVIRONMENT in
...
+## Update the global pull secret within the dataplane cluster to include the read-only credentials for quay.io/rhacs-eng
+#QUAY_READ_ONLY_USERNAME=$(bw get username "66de0e1f-52fd-470b-ad9b-ae0701339dda")
+#QUAY_READ_ONLY_PASSWORD=$(bw get password "66de0e1f-52fd-470b-ad9b-ae0701339dda")
+#quay_basic_auth=$(echo "${QUAY_READ_ONLY_USERNAME}:${QUAY_READ_ONLY_PASSWORD}")
+#oc get secret/pull-secret -n openshift-config --template='{{index .data ".dockerconfigjson" | base64decode}}' > ./tmp-pull-secret.json
+#oc registry login --registry="quay.io/rhacs-eng" --auth-basic="${quay_basic_auth}" --to=./tmp-pull-secret.json --skip-check
```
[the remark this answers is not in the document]
dhaus67: If you set it once more, you will receive a message such as:
```
Info: pull secret not updated
```

So, if you re-execute it, simply nothing will be changed and this will have no effect.

dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
```diff
@@ -16,5 +16,19 @@ spec:
...
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: {{ .Values.acsOperator.source }}
+  namespace: openshift-marketplace
+spec:
+  displayName: 'RHACS Development'
+  publisher: 'Red Hat ACS'
+  sourceType: grpc
+  image: quay.io/rhacs-eng/stackrox-operator-index:{{ .Values.acsOperator.version }}
```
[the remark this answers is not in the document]
dhaus67: So, if you use `quay.io/rhacs-eng`, the central deployed will also come from `quay.io/rhacs-eng`.

So, sadly, it won't use public images for that :/

I've checked w/o adding the pull secret for the whole cluster, the operator fails to start due to the catalog source not being able to pull the index.

I feel like there _should_ be some easier way to handle this (such as only linking the pull secret for specific namespaces, that being the namespace the helm chart is installed + openshift-marketplace), but I'd say once we rely more on this, we can improve this (I added a comment that this could be improved).

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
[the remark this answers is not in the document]
mtodor: Thank you!

dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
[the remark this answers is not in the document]
mtodor: The problem are dynamic namespaces created by fleet-sync when a new ACS instance is provisioned. If we would go this route, then fleet-sync would be responsible for adding pull secrets to newly created namespaces. 😞

docs/development/setup-developer-osd-cluster.md
```diff
@@ -80,17 +80,22 @@ export OSD_ADMIN_USER="osd-admin"
 export OSD_ADMIN_PASS=$(date | md5)
 echo $OSD_ADMIN_PASS > ./tmp-osd-admin-pass.txt
 
-echo "{\"htpasswd\":{\"password\":\"${OSD_ADMIN_PASS}\",\"username\":\"${OSD_ADMIN_USER}\"},\"login\":true,\"mapping_method\":\"add\",\"name\":\"osd-htpasswd\",\"type\":\"HTPasswdIdentityProvider\"}" > ./tmp-osd-htpasswd-body.txt
-ocm post "/api/clusters_mgmt/v1/clusters/${CLUSTER_ID}/identity_providers" --body ./tmp-osd-htpasswd-body.txt
+ocm create idp \
```
mtodor, line 83: Awesome! I missed these commands. 😅

mtodor (approved):

PR #414, opened 2022-09-15
ROX-12158: Extract authentication mechanics outside of redhatsso/client.go

ivan-degtiarenko:
## Description
This PR:
* removes caching mechanism we used in GetToken()
* moves authentication outside of `redhatsso/client.go` to `sso_auth_http_client.go`
* uses `oauth2` library to automatically obtain and refresh token within `httpClient`
* removes `accessToken` parameter from `client.go` methods signatures
* updates tests accordingly


ivan-degtiarenko:
* main
  * **PR #414** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/414" target="_blank">Graphite</a>  👈
    * **PR #416** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/416" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/414?utm_source=stack-comment).

pkg/client/redhatsso/client.go
```diff
@@ -102,8 +103,16 @@ func (c *rhSSOClient) getCachedToken(tokenKey string) (string, error) {
...
+func (c *rhSSOClient) GetAcsAPIToken() (string, error) {
+	return c.getToken([]string{"api.iam.acs"})
+}
+
 // GetToken ...
 func (c *rhSSOClient) GetToken() (string, error) {
+	return c.getToken([]string{})
+}
+
+func (c *rhSSOClient) getToken(additionalScopes []string) (string, error) {
```
dhaus67, line 115: Use ellipsis so we don't have to pass an empty array to the function in `GetToken()`:
```suggestion
func (c *rhSSOClient) getToken(additionalScopes ...string) (string, error) {
```

dhaus67 (approved):
Minor comment w.r.t function signature, otherwise 🚢

pkg/client/redhatsso/client.go
```diff
@@ -102,8 +103,16 @@ func (c *rhSSOClient) getCachedToken(tokenKey string) (string, error) {
...
+}
+
 // GetToken ...
 func (c *rhSSOClient) GetToken() (string, error) {
+	return c.getToken()
+}
+
+func (c *rhSSOClient) getToken(additionalScopes ...string) (string, error) {
 	cachedTokenKey := fmt.Sprintf("%s%s", c.realmConfig.Realm, c.realmConfig.ClientID)
 	cachedToken, _ := c.getCachedToken(cachedTokenKey)
```
dhaus67, line 117: Re-reading through this:
Right now, if one were to use the _same_ client struct to call service account methods and then ACS OIDC client methods, you could potentially run into the issue that the cached token returned here will not have the scopes you require (i.e. `api.iam.acs`).
I'm not sure whether this is something actually going to happen (will we always have two different clients, is it even possible to use the same service account for both the ACS API and the "other" API?).
But, its still possible currently (and could lead to some interesting debugging if it ever occurs). We could do a quick fix for this by adding the `additionalScopes` to the `cachedTokenKey`, so we can be sure it's going to be a cached token with the correct scopes.

dhaus67 (changes requested):
See the comment.

pkg/client/redhatsso/client.go
rukletsov: I don't think we should have multiple service accounts or tokens for a fleet manager instance, see https://chat.google.com/room/AAAAnbZRzRE/XWwTguzFnh8

pkg/client/redhatsso/client.go
```diff
@@ -102,9 +103,18 @@ func (c *rhSSOClient) getCachedToken(tokenKey string) (string, error) {
 	return "", errors.Errorf("failed to retrieve cached token")
 }
 
+func (c *rhSSOClient) GetAcsAPIToken() (string, error) {
```
rukletsov, line 106: I don't think we should have a separate call for this but rather extend `GetToken()` with the additional scopes we need.

pkg/client/redhatsso/client.go
```diff
@@ -102,9 +103,18 @@ func (c *rhSSOClient) getCachedToken(tokenKey string) (string, error) {
...
+	return c.getToken("api.iam.acs")
```
rukletsov, line 107: Let's add `redhat-sso-additional-scopes` config parameter. This way we not only avoid hard-coding magic values but also avoid rebuilding code one SSO config changes.

rukletsov (changes requested):

pkg/client/redhatsso/client.go
ivan-degtiarenko: That would make it necessary to ensure this config value is set up everywhere correctly, and we already suffer from too many config values.

Do you have reasons to think we would need to change the `scope` parameter value faster than via code change(that's the only benefit I see in adding this configuration parameter)?

pkg/client/redhatsso/client.go
rukletsov: This is why we need to extract this as a config parameter : ). Why do we need the "multi-purpose general RH SSO client" path here? We're not editing the fleet manager template.

pkg/client/redhatsso/client.go
rukletsov: We suffer from poor config management and not from having too many values. Having some params hard coded, some called out as defaults, and some exposed is a nightmare.

I think that hard coded parameters almost always strike back. It's much cleaner to have all config for a component in one single place, this way it's easier to verify what's going on and why. Also, the scope name for dynamic client api sounds very temporary to me, I _hope_ we will need to change it soon.

Moreover, what I get from [this discussion](https://chat.google.com/room/AAAAnbZRzRE/XWwTguzFnh8):
* we might need to add at least one more scope soon
* having multiple claims is reasonable
* scope name for dynamic client API might change

So what are the drawbacks of *not* exposing scopes as a parameter?

ivan-degtiarenko: > We suffer from poor config management and not from having too many values. Having some params hard coded, some called out as defaults, and some exposed is a nightmare.
> I think that hard-coded parameters almost always strike back. It's much cleaner to have all config for a component in one single place

I agree with that sentiment. The problem(and the only drawback) is that **at the moment** there is no component config in a single place - it's quite the opposite. For each config value in `fleet-manager` we might have:
* default hard-coded value in default config(see `NewDatabaseConfig` etc)
* hard-coded value in code for different stages(`NewStageEnvLoader` etc)
* `service-template` value of env variable(see https://github.com/stackrox/acs-fleet-manager/blob/92307d9034838e15231e2f730f106333f37e702c/templates/service-template.yml#L42-L44)
*  value assigned in Makefile(see https://github.com/stackrox/acs-fleet-manager/blob/92307d9034838e15231e2f730f106333f37e702c/Makefile#L704-L704)
* value set to env variable in [app-interface](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/acs-fleet-manager/cicd/saas.yaml)

IMHO, none of these places works(at least right now) as `one single place`. The simplest(and at the same time the most wrong) way to add  `redhat-sso-additional-scopes` config value in a single place would be to make it the default hard-coded value in the default config.  But that defeats the purpose of the exercise in the first place. Maybe you have a good solution for this in mind so we won't have to make changes across several files and repos.

> * we might need to add at least one more scope soon
> * having multiple claims is reasonable
> * scope name for dynamic client API might change

Thank you for referencing this discussion! I see where you are coming from now.
In all aforementioned cases, I don't see the rush that would make it necessary to change scopes via config value rather than in code. 
* From the discussion you mentioned, it seems that adding new scope will come with new functionality(and whether this functionality will be placed to the same client is a question for the discussion above).
* Not sure what case you mean by `having multiple claims is reasonable`
* `scope name for dynamic client API might change` - either way if done incorrectly(`api.iam.acs` is replaced with some other scope name without a grace period) this will break prod until a new value scope name value is used. 
If done correctly, switching from old scope name to new scope name will be a change in [app-interface deployment file](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/acs-fleet-manager/cicd/saas.yaml), and the difference will be that one change would be a config value change, when other would be a commit value change.


**tldr**: the drawback is exactly the one that you mentioned - configuration is split between several places in code, deployment files and environment config files. At the same time, it's not clear to me how using configuration value instead of re-deploying a new version will make our lives easier in the cases you mentioned.

pkg/client/redhatsso/client.go
ivan-degtiarenko: > Why do we need the "multi-purpose general RH SSO client" path here? 

What I meant is RH SSO client we can use for several purposes(create SA etc.) However, from the [discussion](https://chat.google.com/room/AAAAnbZRzRE/XWwTguzFnh8) you mentioned I can see there is no need to use minimal `scope` parameter as opposed to adding all scopes needed for client to function into the token call.

Let's keep the `GetToken()` call  with additional scopes instead of `GetAcsApiToken`. @dhaus67, do you have any concerns?

dhaus67: >I can see there is no need to use minimal scope parameter as opposed to adding all scopes needed for client to function into the token call.

Looking at the discussion, there's this sentence: `...it would be more ideal if the FM could scope itself so it would only have these scopes when they are necessary`

I don't really get the "no multi-purpose genereal RH SSO client" argument @rukletsov. If we already foresee that it's required in _some_ future, why not already add it? It saves us time once we _have_ to do it (i.e. once we create dynamic service accounts via RH SSO, which I'm _certain_ will come up if we have more than 1 dataplane cluster, and it's something within the scope of the dataplane CD efforts lead by Draco).

To keep you unblocked, you can remove the `GetAcsApiToken` one for the time being.

dhaus67 (approved):

pkg/client/redhatsso/client.go
```diff
@@ -104,7 +104,12 @@ func (c *rhSSOClient) getCachedToken(tokenKey string) (string, error) {
 
 // GetToken ...
 func (c *rhSSOClient) GetToken() (string, error) {
-	cachedTokenKey := fmt.Sprintf("%s%s", c.realmConfig.Realm, c.realmConfig.ClientID)
+	return c.getToken()
```
rukletsov, line 107: So now we don't need `api.iam.acs` any more? I see it is added in https://github.com/stackrox/acs-fleet-manager/pull/416 but then you need to update PR title and description.

e2e/e2e_auth_test.go
```diff
@@ -261,29 +262,27 @@ func (a *authTestClientFleetManager) doRequestAndUnmarshal(url string, v interfa
...
-		ClientID:         clientID,
-		ClientSecret:     clientSecret, // pragma: allowlist secret
-		TokenEndpointURI: "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
-		JwksEndpointURI:  "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/certs",
-		APIEndpointURI:   "/auth/realms/redhat-external",
-	})
+	oauth2Cfg := &clientcredentials.Config{
+		ClientID:     clientID,
+		ClientSecret: clientSecret, // pragma: allowlist secret
+		TokenURL:     "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
```
misberner, line 271: Just out of curiosity, is there a reason we're using manual OAuth2 configuration instead of OIDC discovery via https://sso.redhat.com/auth/realms/redhat-external/.well-known/openid-configuration ? Is this inherited from the template?

e2e/e2e_auth_test.go
```diff
@@ -261,29 +262,27 @@ func (a *authTestClientFleetManager) doRequestAndUnmarshal(url string, v interfa
...
+	}
 
 	var accessToken string
 	err := retry.WithRetry(
 		func() error {
 			var getTokenErr, retryableErr error
-			accessToken, getTokenErr = client.GetToken()
+			token, getTokenErr := oauth2Cfg.TokenSource(context.Background()).Token()
```
misberner, line 278: I would suggest to rewrite the entire function body as
```
token, err := oauth2Cfg.TokenSource(context.Background()).Token()
if err != nil {
  return retry.MakeRetryable(err)
}
accessToken = token.AccessToken
return nil
```

pkg/client/redhatsso/client.go
```diff
@@ -2,45 +2,31 @@ package redhatsso
...
-const (
-	// access token duration before expiration
-	tokenLifeDuration    = 5 * time.Minute
-	cacheCleanupInterval = 299 * time.Second
-)
-
 // SSOClient ...
 //
 //go:generate moq -out client_moq.go . SSOClient
 type SSOClient interface {
```
misberner, line 16: Just this be called `ServiceAccountsClient` (used as `redhatsso.ServiceAccountsClient`)?

pkg/client/redhatsso/client.go
```diff
@@ -2,45 +2,31 @@ package redhatsso
...
 	GetConfig() *iam.IAMConfig
 	GetRealmConfig() *iam.IAMRealmConfig
-	GetServiceAccounts(accessToken string, first int, max int) ([]serviceaccountsclient.ServiceAccountData, error)
-	GetServiceAccount(accessToken string, clientID string) (*serviceaccountsclient.ServiceAccountData, bool, error)
-	CreateServiceAccount(accessToken string, name string, description string) (serviceaccountsclient.ServiceAccountData, error)
-	DeleteServiceAccount(accessToken string, clientID string) error
-	UpdateServiceAccount(accessToken string, clientID string, name string, description string) (serviceaccountsclient.ServiceAccountData, error)
-	RegenerateClientSecret(accessToken string, id string) (serviceaccountsclient.ServiceAccountData, error)
+	GetServiceAccounts(first int, max int) ([]serviceaccountsclient.ServiceAccountData, error)
+	GetServiceAccount(clientID string) (*serviceaccountsclient.ServiceAccountData, bool, error)
```
misberner, line 20: Is `clientID` the official RHSSO parlance? For me `clientID` in the context of RHSSO refers to an OAuth2 client, so would belong to the dynamic client registration API.

pkg/client/redhatsso/client.go
```diff
@@ -2,45 +2,31 @@ package redhatsso
...
+	CreateServiceAccount(name string, description string) (serviceaccountsclient.ServiceAccountData, error)
+	DeleteServiceAccount(clientID string) error
+	UpdateServiceAccount(clientID string, name string, description string) (serviceaccountsclient.ServiceAccountData, error)
+	RegenerateClientSecret(id string) (serviceaccountsclient.ServiceAccountData, error)
```
misberner, line 24: Same question as above. Is this a "client secret" in the sense of oauth2?

pkg/client/redhatsso/sso_auth_http_client.go
```diff
@@ -0,0 +1,25 @@
...
+	"net/http"
+
+	"github.com/stackrox/acs-fleet-manager/pkg/client/iam"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+// NewSsoAuthHTTPClient returns http client which uses 2-legged OAuth2 flow to automatically get and refresh
+// access token.
+func NewSsoAuthHTTPClient(scopes []string, realmConfig *iam.IAMRealmConfig) *http.Client {
```
misberner, line 15:
```suggestion
func NewSSOAuthHTTPClient(scopes []string, realmConfig *iam.IAMRealmConfig) *http.Client {
```

pkg/client/redhatsso/client.go
```diff
@@ -2,45 +2,31 @@ package redhatsso
...
 }
 
 // NewSSOClient ...
 func NewSSOClient(config *iam.IAMConfig, realmConfig *iam.IAMRealmConfig) SSOClient {
+	httpClient := NewSsoAuthHTTPClient([]string{}, realmConfig)
```
misberner, line 29: Why empty scopes?

pkg/client/redhatsso/client.go
misberner: The API call seems to call it `ServiceAccountSecret`. That seems to be a better wording especially since `Client` is overloaded (the containing interface is called `...Client`)

pkg/client/redhatsso/integration/client_test.go
```diff
@@ -32,17 +34,17 @@ func Test_SSOClient_GetServiceAccounts(t *testing.T) {
...
 	for i := 0; i < 20; i++ {
-		_, err := client.CreateServiceAccount(accessToken, fmt.Sprintf("test_%d", i), fmt.Sprintf("test account %d", i))
+		_, err := client.CreateServiceAccount(fmt.Sprintf("test_%d", i), fmt.Sprintf("test account %d", i))
 		Expect(err).ToNot(HaveOccurred())
 	}
-	accounts, err := client.GetServiceAccounts(accessToken, 0, 100)
+	accounts, err := client.GetServiceAccounts(0, 100)
 	Expect(err).ToNot(HaveOccurred())
-	Expect(accounts).To(HaveLen(20))
+	Expect(accounts).To(HaveLen(21))
```
misberner, line 47: Maybe `require` that prior to the test, there is exactly one service account.

pkg/client/redhatsso/integration/client_test.go
```diff
@@ -53,18 +55,18 @@ func Test_SSOClient_GetServiceAccount(t *testing.T) {
...
 	// create 20 service accounts
 	for i := 0; i < 20; i++ {
-		serviceAccount, err := client.CreateServiceAccount(accessToken, fmt.Sprintf("test_%d", i), fmt.Sprintf("test account %d", i))
+		serviceAccount, err := client.CreateServiceAccount(fmt.Sprintf("test_%d", i), fmt.Sprintf("test account %d", i))
 		Expect(err).ToNot(HaveOccurred())
 		serviceAccountList = append(serviceAccountList, serviceAccount)
 	}
 
-	serviceAccount, found, err := client.GetServiceAccount(accessToken, serviceAccountList[5].GetClientId())
+	serviceAccount, found, err := client.GetServiceAccount(serviceAccountList[5].GetClientId())
```
misberner, line 69: Why does this test create 20 service accounts to test lookup of a single one? 2 (maybe 3, to ensure that the server doesn't return the first or the last unconditionally) would suffice

pkg/client/redhatsso/integration/client_test.go
```diff
@@ -79,24 +81,24 @@ func Test_SSOClient_RegenerateSecret(t *testing.T) {
...
```
misberner, line 95: Single service account suffices

misberner (approved):
Much much nicer! Comments revolve mostly around naming and some restructuring suggestions

pkg/client/redhatsso/integration/client_test.go
```diff
@@ -11,11 +11,13 @@ import (
...
 )
 
-func getClient(baseURL string) redhatsso.SSOClient {
+func getClient(baseURL string, serviceAccount serviceaccountsclient.ServiceAccountData) redhatsso.SSOClient {
 	config := iam.IAMConfig{
 		SsoBaseURL: baseURL,
 		RedhatSSORealm: &iam.IAMRealmConfig{
 			Realm:            "redhat-external",
+			ClientID:         *serviceAccount.ClientId,
+			ClientSecret:     *serviceAccount.Secret,
```
dhaus67, line 20: nit: add the pragma `pragma: allowlist secret`

pkg/client/redhatsso/integration/client_test.go
```diff
@@ -111,9 +113,10 @@ func Test_SSOClient_CreateServiceAccount(t *testing.T) {
...
 
-	client := getClient(server.BaseURL())
-	accessToken := server.GenerateNewAuthToken()
-	serviceAccount, err := client.CreateServiceAccount(accessToken, "test_1", "test account 1")
+	initialServiceAccount := server.GetInitialServiceAccount()
+	client := getClient(server.BaseURL(), initialServiceAccount)
+
+	serviceAccount, err := client.CreateServiceAccount("test_1", "test account 1")
 	Expect(err).ToNot(HaveOccurred())
 	Expect(*serviceAccount.Name).To(Equal("test_1"))
```
dhaus67, line 121: nit: using either `require` for the error to not have occurred or adding `Expect(serviceAccount).ToNot(BeNil())` with `require` would be good to here. Otherwise, in case you receive an error and the `serviceAccount` is nil, you'll get a nil-pointer panic here.

pkg/client/redhatsso/sso_auth_http_client.go
dhaus67: Additionally, this would also be better from consumer POV:
```suggestion
func NewSSOAuthHTTPClient(realmConfig *iam.IAMRealmConfig, scopes ...string) *http.Client {
```

pkg/client/redhatsso/integration/client_test.go
```diff
@@ -32,17 +34,17 @@ func Test_SSOClient_GetServiceAccounts(t *testing.T) {
 
 	defer server.Stop()
 
-	client := getClient(server.BaseURL())
-	accessToken := server.GenerateNewAuthToken()
+	initialServiceAccount := server.GetInitialServiceAccount()
```
dhaus67, line 37: why adding this `initialServiceAccount` to the mock server instead of just setting static client ID / secret in `getClient`?

It makes it hard to understand the tests and see that we expect 21 service accounts when we created 20 without reading the specifics of the mock server.

e2e/e2e_auth_test.go
ivan-degtiarenko: Correct, we inherited manual oauth2 configuration together with everything else SSO-related

pkg/client/redhatsso/sso_auth_http_client.go
```diff
@@ -0,0 +1,25 @@
...
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+// NewSSOAuthHTTPClient returns http client which uses 2-legged OAuth2 flow to automatically get and refresh
+// access token.
+func NewSSOAuthHTTPClient(realmConfig *iam.IAMRealmConfig, scopes ...string) *http.Client {
+	cfg := clientcredentials.Config{
+		ClientID:     realmConfig.ClientID,
+		ClientSecret: realmConfig.ClientSecret,
+		TokenURL:     fmt.Sprintf("%s%s", realmConfig.BaseURL, realmConfig.TokenEndpointURI),
```
dhaus67, line 19: `realmConfig.TokenEndpointURI` already contains the base url, see:
```go
	c.TokenEndpointURI = baseURL + "/auth/realms/" + c.Realm + "/protocol/openid-connect/token"
```

.secrets.baseline
```diff
@@ -271,22 +271,61 @@
         "is_secret": false
       }
     ],
+    "pkg/client/redhatsso/api/api/openapi.yaml": [
```
dhaus67, line 274: This shouldn't be there, the hook explicitly ignores it 🤔 

Can you double check?

.secrets.baseline
```diff
@@ -271,22 +271,61 @@
...
         "type": "Secret Keyword",
         "filename": "pkg/client/redhatsso/client_moq.go",
         "hashed_secret": "5394d7802ce230899b9befae3d5e1d53c5a4dbb6",
         "is_verified": false,
-        "line_number": 385,
-        "is_secret": false
+        "line_number": 321
+      }
+    ],
+    "pkg/client/redhatsso/client_test.go": [
```
dhaus67, line 299: I'd like to see those entries go away by using the pragmas.

Rebasing + merging will be a nightmare with these, so we should keep them to a minimum.

dhaus67 (changes requested):

.secrets.baseline
ivan-degtiarenko: The approach I usually took was - rebase, accept any version and then re-run `detect` command. 
Will try to use pragmas.

pkg/client/redhatsso/sso_auth_http_client.go
ivan-degtiarenko: I see. Do you think it makes sense to initialise `APIEndpointURI` in a similar manner? Or maybe rename it to `APIEndpointPath` to make it clear from the name that it needs to be appended to the base URL?

dhaus67: For consistency it would be better to check the occurrences of `APIEndpointURI` and instantiate it in a similar manner.

pkg/client/redhatsso/client.go
rukletsov: This sounds to me like the implementation detail. We divorced authorization server and resource server despite they are represented by a single entity in the only known use case, RHSSO.

pkg/client/redhatsso/sso_auth_http_client.go
ivan-degtiarenko: From what I see in the IAM config: https://github.com/stackrox/acs-fleet-manager/blob/b459ccdf304e86aaae15ec51c07dbf79022e4fdf/pkg/client/iam/config.go#L66-L89, `APIEndpointURI` will have different values depending on whether it's internal or external SSO, so adding constant prefix wouldn't work.

For now changed how `TokenEndpointURI` is used in `sso_auth_http_client.go`

pkg/client/redhatsso/client.go
ivan-degtiarenko: @misberner as per conversation with CIAM, all CRUD operations in service accounts API should use `id` and not `clientId`.  So the name should reflect that :) I applied the change in https://github.com/stackrox/acs-fleet-manager/pull/416

pkg/client/redhatsso/client.go
ivan-degtiarenko: So what it actually is is a client secret(used for "client credentials" flow) attached to the service account. So correct name would be `RegenerateServiceAccountClientSecret` 😞

.secrets.baseline
ivan-degtiarenko: I think if you run `detect-secrets` manually, this entry is added.

dhaus67 (approved):

rukletsov (approved):

PR #415, opened 2022-09-16
new cluster placement strategies

johannes94:
## Description
Implemented 2 cluster placement strategies for central:

- DefaultClusterPlacementStrategy
  - first cluster in Ready state that has SkipScheduling set to false for central
  - is enabled by default
- TargetClusterPlacementStrategy
  - allows to specify a command line flag `-dataplane-cluster-target`, if that flag is set centrals are placed 

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/internal/dinosaur/pkg/migrations/20220916000000_add_skip_scheduling_to_clusters.go
+++ b/internal/dinosaur/pkg/migrations/20220916000000_add_skip_scheduling_to_clusters.go
@@ -1,0 +1,54 @@
+package migrations
+
+// Migrations should NEVER use types from other packages. Types can change
+// and then migrations run on a _new_ database will fail or behave unexpectedly.
+// Instead of importing types, always re-create the type in the migration, as
+// is done here, even though the same type is defined in pkg/api
+
+import (
+	"fmt"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"github.com/stackrox/acs-fleet-manager/pkg/api"
+	"github.com/stackrox/acs-fleet-manager/pkg/db"
+	"gorm.io/gorm"
+)
+
+func addSkipSchedulingToClusters() *gormigrate.Migration {
+	type Cluster struct {
+		db.Model
+		CloudProvider                    string   `json:"cloud_provider"`
+		ClusterID                        string   `json:"cluster_id" gorm:"uniqueIndex:uix_clusters_cluster_id"`
+		ExternalID                       string   `json:"external_id"`
+		MultiAZ                          bool     `json:"multi_az"`
+		Region                           string   `json:"region"`
+		Status                           string   `json:"status" gorm:"index"`
+		StatusDetails                    string   `json:"status_details" gorm:"-"`
+		IdentityProviderID               string   `json:"identity_provider_id"`
+		ClusterDNS                       string   `json:"cluster_dns"`
+		ProviderType                     string   `json:"provider_type"`
+		ProviderSpec                     string   `json:"provider_spec"`
+		ClusterSpec                      string   `json:"cluster_spec"`
+		AvailableCentralOperatorVersions api.JSON `json:"available_central_operator_versions"`
+		SupportedInstanceType            string   `json:"supported_instance_type"`
+		SkipScheduling                   bool     `json:"skip_scheduling" gorm:"default:false"`
+	}
+
+	return &gormigrate.Migration{
+		ID: "20220916000000",
+		Migrate: func(tx *gorm.DB) error {
+			err := tx.AutoMigrate(&Cluster{})
+			if err != nil {
+				return fmt.Errorf("migrating 20220114114501: %w", err)
+			}
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
+			err := tx.Migrator().DropTable(&Cluster{})
+			if err != nil {
+				return fmt.Errorf("rolling back 20220114114501: %w", err)
+			}
+			return nil
+		},
+	}
+}
--- a/internal/dinosaur/pkg/migrations/20220916000000_add_skip_scheduling_to_clusters.go
+++ b/internal/dinosaur/pkg/migrations/20220916000000_add_skip_scheduling_to_clusters.go
@@ -39,14 +39,14 @@ func addSkipSchedulingToClusters() *gormigrate.Migration {
 		Migrate: func(tx *gorm.DB) error {
 			err := tx.AutoMigrate(&Cluster{})
 			if err != nil {
-				return fmt.Errorf("migrating 20220114114501: %w", err)
+				return fmt.Errorf("migrating 20220916000000: %w", err)
 			}
 			return nil
 		},
 		Rollback: func(tx *gorm.DB) error {
-			err := tx.Migrator().DropTable(&Cluster{})
+			err := tx.Migrator().DropColumn(&Cluster{}, "SkipScheduling")
 			if err != nil {
-				return fmt.Errorf("rolling back 20220114114501: %w", err)
+				return fmt.Errorf("rolling back 20220916000000: %w", err)
 			}
 			return nil
 		},
--- a/internal/dinosaur/pkg/services/cluster_placement_strategy.go
+++ b/internal/dinosaur/pkg/services/cluster_placement_strategy.go
@@ -2,6 +2,7 @@ package services
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/dbapi"
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/config"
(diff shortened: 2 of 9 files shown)


mtesseract:
do you need additional reviewers here?

johannes94:
> 

@mtesseract Yes, probably. I added you to the list :)

internal/dinosaur/pkg/services/cluster_placement_strategy.go
```diff
@@ -20,29 +21,76 @@ type ClusterPlacementStrategy interface {
...
+}
+
+// FindCluster returns the target cluster of the placement strategy if found in the cluster list
+func (f TargetClusterPlacementStrategy) FindCluster(central *dbapi.CentralRequest) (*api.Cluster, error) {
+	clusters, err := findAllClusters(f.clusterService)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, cluster := range clusters {
```
mtesseract, line 77: I've been wondering if it might be a good idea to implement this filter as`FindClusterCriteria`, WDYT?

internal/dinosaur/pkg/services/cluster_placement_strategy.go
```diff
@@ -20,29 +21,76 @@ type ClusterPlacementStrategy interface {
...
-func (d FirstDBClusterPlacementStrategy) FindCluster(dinosaur *dbapi.CentralRequest) (*api.Cluster, error) {
-	clusters, err := d.clusterService.FindAllClusters(FindClusterCriteria{})
+func (d DefaultClusterPlacementStrategy) FindCluster(dinosaur *dbapi.CentralRequest) (*api.Cluster, error) {
+	clusters, err := findAllClusters(d.clusterService)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, cluster := range clusters {
+		if cluster.Status == api.ClusterReady && !cluster.SkipScheduling {
```
mtesseract, line 54: This could be implemented by extending `FindClusterCriteria`, WDYT?

internal/dinosaur/pkg/services/cluster_placement_strategy.go
```diff
@@ -20,29 +21,76 @@ type ClusterPlacementStrategy interface {
...
+	for _, cluster := range clusters {
+		if cluster.ID == f.targetClusterID {
+			return cluster, nil
+		}
+	}
+
+	return nil, fmt.Errorf("target cluster: %v not found in cluster list", f.targetClusterID)
+}
+
+func findAllClusters(c ClusterService) ([]*api.Cluster, error) {
```
mtesseract, line 86: Given the way how this function is used, shouldn't it rather be `findCluster` and return `(*api.Cluster, error)`?

internal/dinosaur/pkg/services/clusters.go
```diff
@@ -593,6 +595,22 @@ func (c clusterService) UpdateMultiClusterStatus(clusterIds []string, status api
...
 }
 
+func (c clusterService) UpdateMultiClusterSkipScheduling(clusterIds []string, skipScheduling bool) *apiErrors.ServiceError {
+	if len(clusterIds) == 0 {
+		return apiErrors.Validation("ids is empty")
+	}
+
+	dbConn := c.connectionFactory.New().
+		Model(&api.Cluster{}).
+		Where("cluster_id in (?)", clusterIds)
```
mtesseract, line 605: Is it correct that this function returns success even if some arbitrary `clusterIds` have been provided for which not a single cluster was found? At the same time the function returns error if `clusterIds` is empty.

internal/dinosaur/pkg/workers/clusters_mgr.go
```diff
@@ -375,15 +375,16 @@ func (c *ClusterManager) reconcileDeprovisioningCluster(cluster *api.Cluster) er
 
 func (c *ClusterManager) reconcileCleanupCluster(cluster api.Cluster) error {
 	glog.Infof("Removing Dataplane cluster %s fleetshard service account", cluster.ClusterID)
-	serviceAcountRemovalErr := c.FleetshardOperatorAddon.RemoveServiceAccount(cluster)
-	if serviceAcountRemovalErr != nil {
-		return errors.Wrapf(serviceAcountRemovalErr, "Failed to removed Dataplance cluster %s fleetshard service account", cluster.ClusterID)
-	}
+	// TODO(addon): reactivate this, if required for cluster terraforming by fleet-manager
```
mtesseract, line 378: Is deactivating this code required?

internal/dinosaur/pkg/workers/clusters_mgr.go
```diff
@@ -702,15 +703,23 @@ func (c *ClusterManager) reconcileClusterWithManualConfig() []error {
...
 			idsOfClustersToDeprovision = append(idsOfClustersToDeprovision, c.Clusterid)
 		}
 	}
 
+	if len(idsOfClusterToSkipScheduling) != 0 {
+		if err := c.ClusterService.UpdateMultiClusterSkipScheduling(idsOfClusterToSkipScheduling, true); err != nil {
+			return []error{errors.Wrapf(err, "setting skip_scheduling for clusters: %v", idsOfClusterToSkipScheduling)}
+		}
+	}
+	glog.Infof("Set skip_scheduling to true for clusters: %v", idsOfClusterToSkipScheduling)
```
mtesseract, line 722:
```suggestion
	glog.Infof("Disabled scheduling for clusters %v", idsOfClusterToSkipScheduling)
```

internal/dinosaur/pkg/workers/clusters_mgr.go
```diff
@@ -702,15 +703,23 @@ func (c *ClusterManager) reconcileClusterWithManualConfig() []error {
...
+			idsOfClusterToSkipScheduling = append(idsOfClusterToSkipScheduling, c.Clusterid)
 		} else {
 			glog.Infof("Excess cluster is going to be deleted %s", c.Clusterid)
 			idsOfClustersToDeprovision = append(idsOfClustersToDeprovision, c.Clusterid)
 		}
 	}
 
+	if len(idsOfClusterToSkipScheduling) != 0 {
+		if err := c.ClusterService.UpdateMultiClusterSkipScheduling(idsOfClusterToSkipScheduling, true); err != nil {
+			return []error{errors.Wrapf(err, "setting skip_scheduling for clusters: %v", idsOfClusterToSkipScheduling)}
```
mtesseract, line 719:
```suggestion
			return []error{errors.Wrapf(err, "disabling scheduling for clusters %v", idsOfClusterToSkipScheduling)}
```

internal/dinosaur/pkg/workers/clusters_mgr.go
```diff
@@ -702,15 +703,23 @@ func (c *ClusterManager) reconcileClusterWithManualConfig() []error {
 	}
 
 	var idsOfClustersToDeprovision []string
+	var idsOfClusterToSkipScheduling []string
```
mtesseract, line 706: Question: What happens when I remove a cluster from the configuration (which has centrals allocated to it) so that the cluster will be marked as non-schedulable and later on I would like to reenable that cluster again by re-adding it to the configuration. I don't know what we are doing in this situation exactly (soft-deletion?), but would the re-added cluster would have `skip_schedulable=false`?

internal/dinosaur/pkg/migrations/20220916000000_add_skip_scheduling_to_clusters.go
```diff
@@ -0,0 +1,54 @@
...
+		ID: "20220916000000",
+		Migrate: func(tx *gorm.DB) error {
+			err := tx.AutoMigrate(&Cluster{})
+			if err != nil {
+				return fmt.Errorf("migrating 20220114114501: %w", err)
+			}
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
+			err := tx.Migrator().DropTable(&Cluster{})
```
mtesseract, line 47: I would have expected something like `DropColumn()` here?

mtesseract (changes requested):

internal/dinosaur/pkg/services/cluster_placement_strategy.go
johannes94: I've removed the entire function, because I'm now using FindClusterByID in the TargetClusterPlacementStrategy. My initial reasoning for this function was to remove duplicated code in both PlacementStrategy implementation. No that TargetClusterPlacementStrategy is not using FindAllClusters this is not necessary.

internal/dinosaur/pkg/workers/clusters_mgr.go
johannes94: It was not set to false, but I only had to change a single line in cluster_mgr.go to get it fixed.
The expected behaviour now is:

- If the cluster is in the config file and not in the DB the entry will be created with skip_scheduled `false`
- If the cluster is in the DB but not in the config file and it has centrals scheduled to it, it sets skip_scheduled to `true`
- If the cluster in in the DB and you add it to the config file after skip_scheduled was set to `true` it will update to `false` on fleet-manager restart

internal/dinosaur/pkg/services/cluster_placement_strategy.go
```diff
@@ -69,28 +67,14 @@ type TargetClusterPlacementStrategy struct {
...
-}
-
-func findAllClusters(c ClusterService) ([]*api.Cluster, error) {
-	clusters, err := c.FindAllClusters(FindClusterCriteria{})
-	if err != nil {
-		return nil, err
-	}
-	if len(clusters) == 0 {
-		return nil, errors.New("no cluster was found")
+	if cluster != nil {
```
mtesseract, line 75: nit: I would expect a function `FindClusterByID` to return some kind of not-found error in case the cluster could not be looked up. In that case you wouldn't need the additional check for `cluster != nil` here.

internal/dinosaur/pkg/services/clusters.go
```diff
@@ -271,10 +272,11 @@ func (c clusterService) FindCluster(criteria FindClusterCriteria) (*api.Cluster,
...
 	clusterDetails := &api.Cluster{
-		CloudProvider: criteria.Provider,
-		Region:        criteria.Region,
-		MultiAZ:       criteria.MultiAZ,
-		Status:        criteria.Status,
+		CloudProvider:  criteria.Provider,
+		Region:         criteria.Region,
+		MultiAZ:        criteria.MultiAZ,
+		Status:         criteria.Status,
+		SkipScheduling: criteria.SkipScheduling,
```
mtesseract, line 279: I have a concern here: If I understand this code correctly, then in L291 an SQL is assembled based on `clusterDetails`. Wouldn't the addition of `SkipScheduling` change the existing behaviour of this function such that previously any `SkipScheduling` setting would have been ignored (it did not exist) and from now on `FindCluster` would only return clusters with `skip_scheduling == false` if this flag is not set in the find criteria. In other words, clusters which are not schedulable will not be returned irregardless of wether schedulability is configured in the find criteria. Is this correct? And if so, is it expected / are all callers of `FindCluster()` happy with this?

mtesseract (commented):
Basically one question/concern.

internal/dinosaur/pkg/services/cluster_placement_strategy.go
johannes94: The implementation of FindClusterByID was like this, I don't want to change it because it could involve changing a lot of other code parts calling this method.

internal/dinosaur/pkg/services/clusters.go
johannes94: Looks to me like the way gorm works a boolean value that is false would not be included in the query: https://gorm.io/docs/query.html#Struct-amp-Map-Conditions

So the filtering would be completely ignored here, I changed it back to the loop for filtering.

PR #416, opened 2022-09-16
ROX-12157: Implement Create/Delete for dynamic clients in redhatsso client

ivan-degtiarenko:
## Description
Implementation of calls to sso.r.c ACS API. This API will be later used for dynamic OIDC client creation for Central IdP.
See details on the API here: https://sso.dev.redhat.com/auth/realms/redhat-external/apis/openapi.yaml

## Test manual

1. Executed throwaway code which creates/deletes acs client via `redhatsso/client.go`


ivan-degtiarenko:
* main
  * **PR #414** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/414" target="_blank">Graphite</a> 
    * **PR #416** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/416" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/416?utm_source=stack-comment).

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("sending acs client request: %w", err)
+	}
+
+	if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		return nil, fmt.Errorf("error creating acs client [%d]", resp.StatusCode)
+	}
+
+	defer shared.CloseResponseBody(resp)
```
dhaus67, line 115: You have to close the body directly after checking for `err != nil`, otherwise we will leak resources when i.e. `httputil.Is2xxx` will yield false.

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
...
+	responseBody, err := ioutil.ReadAll(resp.Body)
```
dhaus67, line 116: ioutils has been deprecated in favor of io / os package:
```suggestion
	responseBody, err := io.ReadAll(resp.Body)
```

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
...
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", accessToken))
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("sending acs client request: %w", err)
+	}
+
+	if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		return nil, fmt.Errorf("error creating acs client [%d]", resp.StatusCode)
```
dhaus67, line 112: Since we ran into this with the central client in fleetshard sync, wouldn't it be worthwhile to also parse the error message here and log it? This way we know whats going on when the call fails, now we would only get i.e. `error creating acs client [400]` which is suboptimal.

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("unmarshalling acs client response data: %w", err)
+	}
+
+	return &responseData, nil
+}
+
+func (c *rhSSOClient) DeleteAcsClient(accessToken string, clientID string) error {
+	client := &http.Client{}
+	req, err := http.NewRequest("DELETE", fmt.Sprintf("%s/apis/beta/acs/v1/%s", c.realmConfig.BaseURL, clientID), nil)
```
dhaus67, line 132:
```suggestion
	req, err := http.NewRequest(http.MethodDelete, fmt.Sprintf("%s/apis/beta/acs/v1/%s", c.realmConfig.BaseURL, clientID), nil)
```

dhaus67, line 132: Might be worthwhile extracting `%s/apis/beta/acs/v1` into a separate function?

dhaus67 (changes requested):

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
 	cache         *cache.Cache
 }
 
+// AcsClientResponseData ...
+type AcsClientResponseData struct {
+	Name         string `json:"name,omitempty"`
+	ClientID     string `json:"clientId,omitempty"`
+	ClientSecret string `json:"secret,omitempty"`
```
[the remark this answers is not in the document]
ivan-degtiarenko: Context for future readers: Keycloak response returns clientId as `clientId`, but clientSecret as `secret`. 

`sso.r.c` endpoint uses the same name convention :(

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
...
+	requestData := &acsClientRequestData{
+		Name:         name,
+		OrgID:        orgID,
+		RedirectUris: redirectUris,
+	}
+	bodyBytes, err := json.Marshal(requestData)
+	if err != nil {
+		return nil, fmt.Errorf("marshalling acs client request data: %w", err)
+	}
+	req, err := http.NewRequest("POST", fmt.Sprintf("%s/apis/beta/acs/v1", c.realmConfig.BaseURL), bytes.NewBuffer(bodyBytes))
```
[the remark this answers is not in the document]
ivan-degtiarenko: What difference would make using `NewReader` vs `NewBuffer` here?

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
 	cache         *cache.Cache
 }
 
+// AcsClientResponseData ...
+type AcsClientResponseData struct {
```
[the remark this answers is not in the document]
ivan-degtiarenko: I don't want to lose `Response` and `Request` part, so let's converge on `AcsOidcClientResponseData`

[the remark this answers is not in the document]
ivan-degtiarenko: At the moment there are no plans to extend API with new structures and endpoints. 

Looking at how it's done with service accounts in the same `redhatsso/client.go` file, RHSSO client returns service account response using library type(e.g original response type). Then conversion to `api.ServiceAccount` happens inside `redhatsso_service.go`: https://github.com/stackrox/acs-fleet-manager/blob/ivan%2Facs-client-create-delete/pkg/services/sso/redhatsso_service.go#L50-L50

I think it would more align with the current structure of the code if we either keep those types in `client.go` or moved them to a separate file within the same package. WDYT?

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,80 @@ type rhSSOClient struct {
...
+	ClientSecret string `json:"secret,omitempty"`
+}
+
+type acsClientRequestData struct {
+	Name         string   `json:"name,omitempty"`
+	OrgID        string   `json:"orgId,omitempty"`
+	RedirectUris []string `json:"redirectUris,omitempty"`
+}
+
+func (c *rhSSOClient) CreateAcsClient(accessToken string, name string, orgID string, redirectUris []string) (*AcsClientResponseData, error) {
```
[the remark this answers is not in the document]
ivan-degtiarenko: I don't think it would be possible to reuse `redhatsso_service.go` as the `redhatsso/client.go` to call ACS API endpoint would require a separate service account to use. 

Considering that, let's remove `GetAcsApiToken` from the interface and remove `accessToken` from create/delete methods signature

pkg/client/redhatsso/client.go
ivan-degtiarenko: My sentiment here is that it's easier to understand the end result of formatting if it is not divided between different methods, WDYT?

pkg/client/redhatsso/client.go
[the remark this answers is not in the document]
dhaus67: If we don't foresee much extension, I'd say it's fine for me to keep it within the client.

pkg/client/redhatsso/client.go
dhaus67: fine by me, but let's define `apis/beta/acs/v1` as const, so its easier to change paths if we require to do so.

pkg/client/redhatsso/client.go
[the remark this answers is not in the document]
dhaus67: It's a nit, but I like to make it clear that I'm only interested in the `io.Reader` interface.
`NewBuffer` returns a `bytes.Buffer`, which implements `io.Reader` and `io.Writer`. We technically only need the `io.Reader`, so for me feels cleaner to make that visible by using `NewReader`.

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,110 @@ type rhSSOClient struct {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("sending acs client request: %w", err)
+	}
+	defer shared.CloseResponseBody(resp)
+
+	if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		errResponse, err := extractAcsAPIError(resp)
+		if err != nil {
+			return nil, fmt.Errorf("error creating acs client [%d], error parsing %w", resp.StatusCode, err)
```
dhaus67, line 123: nit:
```suggestion
			return nil, fmt.Errorf("error creating acs client [%d], error parsing response body: %w", resp.StatusCode, err)
```

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +72,110 @@ type rhSSOClient struct {
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("sending acs client DELETE request: %w", err)
+	}
+
+	if !httputil.Is2xxStatusCode(resp.StatusCode) {
+		errResponse, err := extractAcsAPIError(resp)
+		if err != nil {
+			return fmt.Errorf("error deleting acs client [%d], error parsing: %w", resp.StatusCode, err)
```
dhaus67, line 161:
```suggestion
			return fmt.Errorf("error deleting acs client [%d], error parsing response body: %w", resp.StatusCode, err)
```

dhaus67 (approved):
Only minor nits, otherwise LGTM

rukletsov (changes requested):

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +74,110 @@ type rhSSOClient struct {
...
+	RedirectUris []string `json:"redirectUris,omitempty"`
+}
+
+type errorResponse struct {
+	Error       string `json:"error,omitempty"`
+	Description string `json:"error_description,omitempty"`
+}
+
+func (c *rhSSOClient) CreateAcsClient(name string, orgID string, redirectUris []string) (*AcsOidcClientResponseData, error) {
+	client := &http.Client{}
```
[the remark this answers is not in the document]
misberner: Not caching the client isn't harmful. The harm is done by creating new `http.Transport` objects, because the connection pool is maintained at that level.

Because this `client` doesn't set a `Transport`, it uses `http.DefaultTransport` and thus a shared instance with a shared connection pool.

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +74,110 @@ type rhSSOClient struct {
 	cache         *cache.Cache
 }
 
+// AcsOidcClientResponseData is the response from sso.redhat.com ACS API endpoint.
```
misberner, line 77: Are there no OpenAPI specs we could generate client bindings from?

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +74,110 @@ type rhSSOClient struct {
...
+		return nil, fmt.Errorf("error creating acs client [%d], error=%q, error_description=%q", resp.StatusCode, errResponse.Error, errResponse.Description)
+	}
+
+	responseBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("reading response body: %w", err)
+	}
+
+	var responseData AcsOidcClientResponseData
+	err = json.Unmarshal(responseBody, &responseData)
```
misberner, line 136: There's a lot of repeated manual json marshalling/unmarshalling going on here. Please have a look at `fleetshard/pkg/central/client/client.go:84` on a more generic way to implement a REST-based client (in case there's no OpenAPI spec, that is). You might need to adapt this for error parsing, but not by much.

pkg/client/redhatsso/client.go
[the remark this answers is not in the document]
misberner: However, I think it would be beneficial to maintain the `client` as a field. What you can then do is to write a transport wrapper (you can use an `oauth2.Transport` if you comply to the `TokenSource` interface, which probably would be a beneficial change) that automatically injects the token, and keep that bit out of the actual REST client logic.

pkg/client/redhatsso/client.go
```diff
@@ -67,6 +74,110 @@ type rhSSOClient struct {
...
+	requestData := &acsOidcClientRequestData{
+		Name:         name,
+		OrgID:        orgID,
+		RedirectUris: redirectUris,
+	}
+	bodyBytes, err := json.Marshal(requestData)
+	if err != nil {
+		return nil, fmt.Errorf("marshalling acs client request data: %w", err)
+	}
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s/%s", acsAPIPath, c.realmConfig.BaseURL), bytes.NewReader(bodyBytes))
```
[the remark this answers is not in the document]
ivan-degtiarenko: Not sure what happened here -  something broke during refactoring. Thanks for noticing, will fix

pkg/client/redhatsso/client.go
rukletsov: There are. Ivan has discussed this option with me and Daniel offline and we agreed that it might not worth it to generate a client from the provided openapi spec and later embed it into redhatsso client.

ivan-degtiarenko: Basically, we decided that:
* there's no particular benefit to using generated client(we'd still need to add authorisation, and we don't save much code-wise by using generated client vs http client)
* API is small and generating two structs instead of writing them seems unnecessary(especially considering the fact that the same way we need to keep these structs up-to-date, we would need to keep openapi spec up-to-date)

pkg/client/redhatsso/client.go
ivan-degtiarenko: Not sure I understand what you suggest here. I see two methods:
* `Create...` method has 1 marshalling and 2 unmarshalling within it. You can't avoid marshalling the data and unmarshalling the data, as well as handling error response separately. `fleetshard/pkg/central/client/client.go` does exactly that(I'm pretty sure since I wrote it). The difference between them is that one talks to proto-based API(and thus can make use of `jsonpb`), while other doesn't. 
* `Delete...` unmarshalls only error response.

pkg/client/redhatsso/client.go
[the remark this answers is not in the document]
ivan-degtiarenko: @misberner not sure I get you. If we use the simplest `StaticTokenSource` , we would still need to call `getToken()` on every client call. The only thing it would change is instead of writing to `Authorisation` header we would update value in `StaticTokenSource` instance. Or do you suggest writing auto-refreshing `TokenSource` that would call token endpoint for us, cache and track token?

pkg/client/redhatsso/client.go
misberner: I said "Please have a look at `fleetshard/pkg/central/client/client.go`" for a reason, and it is honestly a bit disappointing to see such a request apparently being outright ignored.

That's because https://github.com/stackrox/acs-fleet-manager/blob/main/fleetshard/pkg/central/client/client.go#L84 is most certainly not what you implemented, but what Daniel, who took over your PR, implemented in response to my comments. It introduces a generic REST call method that optionally takes in a protobuf message that gets jsonpb-serialized as the request body, and optionally serializes the response body to a protobuf message via jsonpb, while also parsing JSON errors and handling auth injection.

While 2 different methods isn't an awful lot, writing a single generic implementation helps with readability because it focuses the reader's attention on the parts that are actually method-specific. It also reduces the likelihood of introducing small errors due to manual repetition of boilerplate-ish code, such as correctly handling HTTP responses (bodies must always be closed, which you don't do in `Delete...`).

ivan-degtiarenko: > I said "Please have a look at fleetshard/pkg/central/client/client.go" for a reason, and it is honestly a bit disappointing to see such a request apparently being outright ignored.

Please don't take me not noticing something in the code as maliciously ignoring your requests :) 

Judging from your explanation your idea is to unify two calls with one generic implementation which will:
* conditionally marshal request and unmarshal response
* acquire token + add auth to the request
* handle error response

Is that correct?

misberner: Yes, that is correct. I am referring to the equivalent of a method like the one in the explicitly referenced line 84 in that package.

pkg/client/redhatsso/dynamicclients/client.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return &dynamicClient, nil
+}
+
+func (c *rhSSOClient) DeleteDynamicClient(clientID string) error {
+	resp, err := api.NewAPIClient(c.configuration).
+		AcsTenantsApi.
+		DeleteAcsClient(context.Background(), clientID)
+
+	defer shared.CloseResponseBody(resp)
```
dhaus67, line 85: no need to close the response body, it's done within the generated API client. It's the whole point of the generated client to not deal with this kind of stuff but rather focus on the pure functionality you want to provide.

Besides that, why even use the `http.Response` here? You could just ignore it as return parameter, we're only interested whether an error is returned.

Same applies for `CreateDynamicClient`

dhaus67 (changes requested):
I don't really understand why we create this "manual" thin wrapper around the generated API client's method (specifically `Create / Delete`) and duplicate the behavior (even taking care of closing response bodies!) when this is already dealt with in the generated client.

I would expect the thin wrapper to be responsible for creating the "correct" instance of the API client with the proper authentication method + scopes and correct configuration for the generated API client, and a way of accessing the API client. I wouldn't expect us to re-write the whole functions within the generated client, what if the API grows to include 20 API calls, are you going to write a thin wrapper around each call within this client, just to mimic the behavior of the generated client?

The only reason I see for the `Create` wrapper is due to the fact that the current version of the openAPI generator we are using is not validating the request model based on the constraints within the API spec (name must be between 1 and 50 characters). But, ideally, this should also be a part of the generated client.

pkg/services/sso/iam_service.go
```diff
@@ -24,15 +24,13 @@ type CompleteServiceAccountRequest struct {
 //
 //go:generate moq -out iam_service_moq.go . IAMService
 type IAMService interface {
-	GetConfig() *iam.IAMConfig
-	GetRealmConfig() *iam.IAMRealmConfig
```
dhaus67, line 28: Why this change, paired with the change in the route loader.

This currently leads to the following error when starting fleet manager:
```
panic: fleet-manager flag redefined: sso-base-url

...
github.com/stackrox/acs-fleet-manager/pkg/environments.(*Env).AddFlags(0xc000433380, 0xb?)
        /Users/dhaus/go/src/github.com/stackrox/acs-fleet-manager/pkg/environments/environment.go:89 +0x27f

```

That's due to the `*iam.Config` being required _twice_ in the dependency graph (due to us using it within. the `route_loader`. This means `AddFlags` will be called twice, and thus it fails.

We have to somehow fix this, either _just_ injecting the `IAMService` which has the `GetConfig` interface _or_ by rewriting the dependency injection to dedupe the required modules and not mistakenly call `AddFlags` twice.

pkg/client/redhatsso/dynamicclients/client.go
ivan-degtiarenko: Cool. I followed the example of how we handle service account API. Looks like we can remove that `defer` line from `serviceaccounts/client.go` as well

pkg/services/sso/iam_service.go
ivan-degtiarenko: I'll try to figure out if there is an easy way to use the same config twice - would be good if the dependency framework would allow us to do that :)

The reason I removed `GetConfig` in the first place is that from our discussion I assumed it shouldn't be returned from either `serviceaccounts/client.go` or `dynamicclients/client.go`. SSO config should exist independently of the service/client which is using it.

dhaus67: Resolved - culprit was the duplicated line https://github.com/stackrox/acs-fleet-manager/pull/416/files#diff-4d566b3cb726b93ff3892b83fe02f829f876d376c350f6d72a65e840eb380380R53

misberner (approved):
Looks good in general, but please add contexts to new message and change the truncation logic.

pkg/client/redhatsso/dynamicclients/client.go
```diff
@@ -0,0 +1,85 @@
...
+func (c *rhSSOClient) CreateDynamicClient(name string, orgID string, redirectUris []string) (*api.AcsClientResponseData, error) {
+	truncatedName := stringutils.Truncate(name, dynamicClientAPIMaxNameSize)
+	requestData := api.AcsClientRequestData{
+		Name:         truncatedName,
+		OrgId:        orgID,
+		RedirectUris: redirectUris,
+	}
+	dynamicClient, _, err := api.NewAPIClient(c.configuration).
+		AcsTenantsApi.
+		CreateAcsClient(context.Background(), requestData)
```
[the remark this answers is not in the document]
ivan-degtiarenko: Looking at how the `fleet-manager` code currently is, there is no `context` in any workers(`accepted_dinosaurs_mgr.go` etc). The only place in code we are at the moment calling these methods from are workers.

If we will have a good reason, we can always extract context to the interface. Right now by pulling it out to the interface, we will expose implementation detail callers don't need to know, but are forced to correctly set.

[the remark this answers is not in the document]
misberner: > The only place in code we are at the moment calling these methods from are workers.

Even if the caller doesn't have an external context readily available, exposing contexts in the interface still allows for better control regarding timeouts.

> we can always extract context to the interface.

This is often substantially more work because of the way Go's interfaces work.

> Right now by pulling it out to the interface, we will expose implementation detail callers don't need to know, but are forced to correctly set.

I don't understand this. Why is this implementation detail, and what do we force on callers to correctly set? A context parameter primarily indicates that a method may perform a blocking operation. This is not implementation detail, but essential for callers to be aware of.

pkg/client/redhatsso/serviceaccounts/client.go
```diff
@@ -1,34 +1,33 @@
...
+type Client interface {
 	GetServiceAccounts(first int, max int) ([]serviceaccountsclient.ServiceAccountData, error)
-	GetServiceAccount(clientID string) (*serviceaccountsclient.ServiceAccountData, bool, error)
+	GetServiceAccount(id string) (*serviceaccountsclient.ServiceAccountData, bool, error)
 	CreateServiceAccount(name string, description string) (serviceaccountsclient.ServiceAccountData, error)
-	DeleteServiceAccount(clientID string) error
-	UpdateServiceAccount(clientID string, name string, description string) (serviceaccountsclient.ServiceAccountData, error)
+	DeleteServiceAccount(id string) error
+	UpdateServiceAccount(id string, name string, description string) (serviceaccountsclient.ServiceAccountData, error)
 	RegenerateClientSecret(id string) (serviceaccountsclient.ServiceAccountData, error)
```
rukletsov, line 24: Didn't [we agree](https://github.com/stackrox/acs-fleet-manager/pull/414#discussion_r985951548) that the method name should be `RegenerateServiceAccountClientSecret`?

ivan-degtiarenko: To be honest, I'm no big fan of `RegenerateServiceAccountClientSecret`. My point in this PR was that if we were to name the method correctly, it would be an enormous and hard-to-read name. Saying that, it's not principal to me. Changed to `RegenerateServiceAccountClientSecret`

ivan-degtiarenko:
I removed the thin wrapper according to @dhaus67 suggestion.  That also resolves questions around whether to pass context or not(see https://github.com/stackrox/acs-fleet-manager/pull/416#discussion_r986671568) and how to name methods(see https://github.com/stackrox/acs-fleet-manager/pull/416#discussion_r986907361).

I also removed `redhatsso_service_test.go` for next reasons:
1) `redhatsso_service.go` does not use service accounts API correctly and needs rework, see https://issues.redhat.com/browse/ROX-12961. I would avoid adding this work to the scope of this PR.
2) `redhatsso_service_test.go` tested logic between service and client layer, and the client layer no longer exist.

pkg/client/redhatsso/dynamicclients/client.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/stackrox/acs-fleet-manager/pkg/client/redhatsso"
+	"github.com/stackrox/acs-fleet-manager/pkg/client/redhatsso/api"
+)
+
+// NewDynamicClientsAPI returns new instance of dynamic clients sso.redhat.com API client.
+func NewDynamicClientsAPI(realmConfig *iam.IAMRealmConfig) *api.AcsTenantsApiService {
+	httpClient := redhatsso.NewSSOAuthHTTPClient(realmConfig, "api.iam.acs")
+	configuration := &api.Configuration{
+		BasePath:  realmConfig.BaseURL + realmConfig.APIEndpointURI,
+		UserAgent: "OpenAPI-Generator/1.0.0/go",
```
dhaus67, line 14: nit: should we use a proper user-agent header for this indicating that its ACS fleet manager (in case the people from the sso.r.c want to do some metrics of our usage of the API)? I don't see at the moment that we pass version information to fleet manager via env and sadly we are not using go 1.18 to get the info from [runtime/debug.BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo) so we could just use `RHACS-Fleet-Manager/1.0`.

pkg/client/redhatsso/serviceaccounts/client.go
```diff
@@ -0,0 +1,24 @@
...
+
+	serviceaccountsclient "github.com/redhat-developer/app-services-sdk-go/serviceaccounts/apiv1internal/client"
+	"github.com/stackrox/acs-fleet-manager/pkg/client/iam"
+)
+
+// NewServiceAccountsAPI returns new instance of service accounts sso.redhat.com API client.
+func NewServiceAccountsAPI(realmConfig *iam.IAMRealmConfig) serviceaccountsclient.ServiceAccountsApi {
+	httpClient := redhatsso.NewSSOAuthHTTPClient(realmConfig, "api.iam.service_accounts")
+	configuration := &serviceaccountsclient.Configuration{
+		UserAgent: "OpenAPI-Generator/1.0.0/go",
```
dhaus67, line 14: nit: same notion as above regarding the user agent.

test/mocks/redhatsso.go
```diff
@@ -143,10 +147,12 @@ func (mockServer *redhatSSOMock) init() {
...
-	bearerTokenAuthRouter.HandleFunc("/auth/realms/redhat-external/apis/service_accounts/v1/{clientId}/resetSecret", mockServer.regenerateSecretHandler).Methods("POST")
+	bearerTokenAuthRouter.HandleFunc("/auth/realms/redhat-external/apis/service_accounts/v1/{id}", mockServer.getServiceAccountHandler).Methods("GET")
+	bearerTokenAuthRouter.HandleFunc("/auth/realms/redhat-external/apis/service_accounts/v1/{id}", mockServer.deleteServiceAccountHandler).Methods("DELETE")
+	bearerTokenAuthRouter.HandleFunc("/auth/realms/redhat-external/apis/service_accounts/v1/{id}", mockServer.updateServiceAccountHandler).Methods("PATCH")
+	bearerTokenAuthRouter.HandleFunc("/auth/realms/redhat-external/apis/service_accounts/v1/{id}/resetSecret", mockServer.regenerateSecretHandler).Methods("POST")
+	bearerTokenAuthRouter.HandleFunc("/auth/realms/redhat-external/apis/beta/acs/v1", mockServer.createDynamicClientHandler).Methods("POST")
+	bearerTokenAuthRouter.HandleFunc("/auth/realms/redhat-external/apis/beta/acs/v1/{clientId}", mockServer.deleteDynamicClientHandler).Methods("DELETE")
```
dhaus67, line 155: Might be an obvious one but all the naming regarding `id` vs `clientID` 🥴 
Is it `clientId` or `id` here? Spec has `id`.

dhaus67 (approved):
Just minor things + a `.secrets.baseline` clean-up. 
Otherwise, LGTM 🚢

test/mocks/redhatsso.go
ivan-degtiarenko: That's a good question which actually shows inconsistency in spec. It's `id` in delete request, but `clientId` in `AcsClientResponseData`.

ivan-degtiarenko: Asked in our chat with CIAM team.

PR #430, opened 2022-09-21
ROX-12747: Refactor the fleet manager client

dhaus67:
## Description

Currently, we have a single, manually written client within `fleetshard/pkg/fleetmanager` that is used to interact with fleet manager.

Initially, it was only used within fleeshard, hence also the location. Over time, the usages have grown to include the E2E tests.

Within the context of writing the probe service, we'd like to re-use this client as well. However, the current location as well as the client per-se seem unfortunate to use off-the-shelve.

This PR aims to improve this situation by doing the following:
- Move the client and auth to a generally-purpose-fit location, specifically `pkg/client/fleetmanager`.
- Create a wrapper around the three generated OpenAPI clients for the admin, public, private API of fleet manager.
- Create a transport that injects the bearer token based on the auth type, re-use the transport in a custom HTTP client within the client wrapper / OpenAPI clients.
- Improve the configuration of the different auth types by adding an options pattern, but keeping the possibility to configure them via environment variables.
- Rename the handler within the routes to properly be `admin/centrals` instead of `admin/dinosaurs`.

What's left and will be done within a follow-up PR:
The usage of this client in combination with the RH SSO auth type currently implies the following: the usage of the token-refresher. Originally thought of as an egress proxy for fleetshard-sync that will inject the auth information within requests to central, but modified to write to a file for local development. This is something completely undesired and we should change. This will be done in a follow-up, removing the need for the token-refresher and doing the client-credentials grant within the auth type. This way, the existing fleetshard-sync deployment will lose another dependency and component to fail (token-refresher) and all new services that may depend on this client (probe-service) will not suffer by having to use the token-refresher sidecar.

## Test manual

- see the e2e tests, they should be working as expected :-)


dhaus67:
* main
  * **PR #430** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/430" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/430?utm_source=stack-comment).

pkg/client/fleetmanager/client.go
```diff
@@ -0,0 +1,130 @@
...
+	auth      Auth
+}
+
+func (c *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if err := c.auth.AddAuth(req); err != nil {
+		return nil, errors.Wrapf(err, "setting auth on req %+v", req)
+	}
+	resp, err := c.transport.RoundTrip(req)
+	if err != nil {
+		return resp, errors.Wrap(err, "round trip error")
```
misberner, line 24:
```suggestion
		return resp, err
```
The wrap does not provide any value whatsoever. If the linter insists on it, it needs to be suppressed at this point

misberner: I guess you can change the entire thing to `return c.transport.RoundTrip(req)`

pkg/client/fleetmanager/auth.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// NewAuth will return Auth that can be used to add authentication of a specific AuthType to be added to HTTP requests.
+func NewAuth(t string, opts ...AuthOption) (Auth, error) {
+	authOption := &option{}
+	for _, opt := range opts {
+		opt(authOption)
+	}
+
+	factory, exists := authFactoryRegistry[t]
```
misberner, line 42: Move this line and the `exists` check to the very top of the function

pkg/client/fleetmanager/auth.go
```diff
@@ -0,0 +1,143 @@
...
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+}
+
+// getAllAuthTypes is a helper used within logging to list the possible values for auth types.
+func getAllAuthTypes() []string {
+	authTypes := make([]string, 0, len(authFactoryRegistry))
+	for authType := range authFactoryRegistry {
+		authTypes = append(authTypes, authType)
+	}
+	return authTypes
```
misberner, line 130:
```suggestion
	sort.Strings(authTypes)
	return authTypes
```
Better to have a consistent order

pkg/client/fleetmanager/auth.go
```diff
@@ -0,0 +1,143 @@
...
+//   - RHSSO_TOKEN_FILE for the path to the file containing the RH SSO access token.
+func WithOptionFromEnv() AuthOption {
+	return func(o *option) {
+		optFromEnv := &option{}
+		utils.Must(env.Parse(optFromEnv))
+		overrideValues(o, optFromEnv)
+	}
+}
+
+// setBearer is a helper to set a bearer token as authorization header on the http.Request.
```
misberner, line 119: This looks out of place in this file at this point. Maybe move it to a file `auth_token_common.go`?

fleetshard/pkg/runtime/runtime.go
```diff
@@ -37,19 +37,21 @@ var backoff = wait.Backoff{
...
+	clusterID        string
 	reconcilers      reconcilerRegistry // TODO(create-ticket): possible leak. consider reconcilerRegistry cleanup
 	k8sClient        ctrlClient.Client
 	statusResponseCh chan private.DataPlaneCentralStatus
 }
 
 // NewRuntime creates a new runtime
 func NewRuntime(config *config.Config, k8sClient ctrlClient.Client) (*Runtime, error) {
-	auth, err := fleetmanager.NewAuth(config.AuthType)
+	auth, err := fleetmanager.NewAuth(config.AuthType, fleetmanager.WithOptionFromEnv())
```
misberner, line 48: Why this change? You are breaking the functional property of `NewRuntime` of depending only on the input by suddenly and surprisingly taking _some_ fraction of the config options from the environment, with no way for the user to inject something here, requiring `Setenv` calls in tests where previously merely setting a struct field had sufficed. This is particularly confusing as, as far as I understand, `config.Config` is already intended to be populated from env vars.

I think it would be much nicer to allow setting this via `config.Config` fields.

fleetshard/pkg/runtime/runtime.go
```diff
@@ -58,6 +60,7 @@ func NewRuntime(config *config.Config, k8sClient ctrlClient.Client) (*Runtime, e
 		config:      config,
 		k8sClient:   k8sClient,
 		client:      client,
+		clusterID:   config.ClusterID,
```
misberner, line 63: Why does the `Runtime` now have a single specific `clusterID`?

misberner (changes requested):
Thanks for making this change!

fleetshard/pkg/runtime/runtime.go
dhaus67: Previously, the manually written client had the `clusterID` from the `config.Config` injected as a struct member.

Now, since we use the generated one, we don't have this as a member anymore. Since the client is used throughout different places (e.g. `runtime.Start, runtime.handleReconcileResult`) I've added this as a field in the struct.

Not sure there's a better place / way to inject this, felt most natural to me.

openapi/fleet-manager-private-admin.yaml
```diff
@@ -259,7 +259,7 @@ paths:
             application/json:
               schema:
                 $ref: 'fleet-manager.yaml#/components/schemas/Error'
-  '/api/rhacs/v1/admin/db/centrals/{id}':
+  '/api/rhacs/v1/admin/centrals/db/{id}':
```
misberner, line 262: Again, how is backwards-compatibility ensured here?

pkg/client/fleetmanager/auth.go
```diff
@@ -0,0 +1,141 @@
...
+func init() {
+	authFactoryRegistry = map[string]authFactory{
+		ocmFactory.GetName():         ocmFactory,
+		rhSSOFactory.GetName():       rhSSOFactory,
+		staticTokenFactory.GetName(): staticTokenFactory,
+	}
+}
+
+// NewAuth will return Auth that can be used to add authentication of a specific AuthType to be added to HTTP requests.
+func NewAuth(t string, opts ...AuthOption) (Auth, error) {
```
misberner, line 37: I find this factory function to be a bit weird. Allowing the type `t` to be dynamic is somewhat undermined by requiring the option for the concrete provider to be spelled out at compile time. Indeed, it seems like this is only invoked with either _all_ the provider-specific `With...` options, or with `WithOptionFromEnv()`. However, even `WithOptionFromEnv` in my view conceals a little bit too much the nature of this function: the environment is required to contain _all_ the options for the various auth provider types, or at least the one the user had in mind when specifying `t` (which may or may not come from the environment).

To emphasize the fact that options and `t` are decoupled, I would propose to export the `option` struct (call it `Options`), make it the second argument (no varargs), add a function `func OptionsFromEnv() Options` or a method `func (o *Options) PopulateFromEnv()`, and get rid of the `With...` option setters.

If you think about it, the individual `With...` options are useless. Either I am specifying all of them anyway, or I am specifying a single one because I statically know `t` when calling `NewAuth`. But if I statically know, e.g., that I want to instantiate an `RHSSO` authenticator, I would better use a `NewRHSSOAuth(opts RhSsoOption) (Auth, error)`, which would be less error-prone.

misberner (changes requested):

openapi/fleet-manager-private-admin.yaml
rukletsov: I agree with the thinking in general. Here I think we can cut that corner still:
* we have exactly 0 users (so let's ship this quick 😛)
* SP does not promise API stability
* nor SP promises flawless communication between fleet manager and fleetshard sync (not to mention again there is exactly 0 requests per day)

pkg/client/fleetmanager/auth.go
dhaus67: Yeah, I agree. I wasn't initially sure whether I wanted to keep the factory as-is or split it up to be more "static".

I've chosen now to do the following:
- Create `NewAuth` with which you can specify the type and a _single_ option as parameter.
- Create additional functions that allow to create a _specific_ auth type (e.g. `NewRHSSOAuth`), which take in the _specific_ option as parameter (e.g. `RHSSOOption`).
- Create a helper `OptionFromEnv` that will re-use env settings to populate an option struct.

openapi/fleet-manager-private-admin.yaml
dhaus67: The only thing that changed here was aligning the spec to the actual implementation.

The spec specified the endpoint to be `api/rhacs/v1/admin/db/centrals/..`, but the actual implementation is `api/rhacs/v1/admin/centrals/db/..`, which was not correct and never working.

I skipped anything related to backwards compatibility due to:
- This is fixing a broken spec, wich specified a wrong endpoint.
- The admin API is not used by neither fleetshard-sync nor UI (only referenced in the runbook).

internal/dinosaur/pkg/routes/route_loader.go
```diff
@@ -209,36 +209,34 @@ func (s *options) buildAPIBaseRouter(mainRouter *mux.Router, basePath string, op
...
+	adminCentralHandler := handlers.NewAdminDinosaurHandler(s.Dinosaur, s.AccountService, s.ProviderConfig)
 	adminRouter := apiV1Router.PathPrefix("/admin").Subrouter()
 
-	// TODO(ROX-11683): For now using RH SSO issuer for the admin API, but needs to be re-visited within this ticket.
 	adminRouter.Use(auth.NewRequireIssuerMiddleware().RequireIssuer(
 		[]string{s.IAM.GetConfig().InternalSSORealm.ValidIssuerURI}, errors.ErrorNotFound))
 	adminRouter.Use(auth.NewRolesAuhzMiddleware(s.AdminRoleAuthZConfig).RequireRolesForMethods(errors.ErrorNotFound))
 	adminRouter.Use(auth.NewAuditLogMiddleware().AuditLog(errors.ErrorNotFound))
-	adminDinosaursRouter := adminRouter.PathPrefix("/dinosaurs").Subrouter()
+	adminCentralsRouter := adminRouter.PathPrefix("/centrals").Subrouter()
```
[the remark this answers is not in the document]
dhaus67: So, admin API is not used by either UI nor fleetshard sync.

The only reference we currently have is [within this runbook](https://github.com/stackrox/acs-managed-service-runbooks#changing-resources-configuration-of-a-production-central-tenant) where I'd create a PR once this is approved and merge both simultaneously.

I can still do the exercise of creating a backwards-compatible API path, but feel like the actual value is quite small since the API is not used in any component other the runbook.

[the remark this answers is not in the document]
misberner: Sounds good, I thought that this API was in use

openapi/fleet-manager-private-admin.yaml
misberner: SG

misberner (approved):

PR #431, opened 2022-09-21
ROX-12747: Remove token-refresher

dhaus67:
## Description

This is a follow-up from the previous PR. See the parent PR for additional context.

Since we don't want to force any consumer to require a token-refresher when using the RH SSO auth type, this PR will remove the need for it by doing the following:
- Remove the config `RHSSO_TOKEN_FILE`.
- Refactor the RH SSO auth type to re-use the `redhatsso.Client` that already exists.
- Include new configuration parameters to set the client ID / client Secret as well as the realm / endpoint of the SSO server.
- Remove the token-refresher from the Helm chart.


dhaus67:
* main
  * **PR #430** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/430" target="_blank">Graphite</a> 
    * **PR #431** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/431" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/431?utm_source=stack-comment).

stehessel (approved):

pkg/client/fleetmanager/auth_rhsso.go
```diff
@@ -19,35 +21,35 @@ var (
...
-	// The file is populated by the token-refresher, which will ensure the token is not expired.
-	token, err := shared.ReadFile(r.tokenFilePath)
+	token, err := r.tokenSource.Token()
 	if err != nil {
-		return errors.Wrapf(err, "reading token file %q within rhsso auth", r.tokenFilePath)
+		return errors.Wrap(err, "retrieving token from token source")
 	}
-
-	setBearer(req, token)
+	setBearer(req, token.AccessToken)
```
misberner, line 53: It's a bit sad that we manually inject this into a request given that the oauth2 lib already has helpers for injecting this at the client level, but I guess we have to do this in order to remain consistent with other auth methods.

pkg/client/fleetmanager/auth_rhsso.go
```diff
@@ -19,35 +21,35 @@ var (
...
-// CreateAuth ...
+// CreateAuth creates an Auth using RH SSO.
 func (f *rhSSOAuthFactory) CreateAuth(o Option) (Auth, error) {
-	tokenFilePath := o.Sso.TokenFile
-	if _, err := shared.ReadFile(tokenFilePath); err != nil {
-		return nil, fmt.Errorf("reading token file: %w", err)
+	cfg := clientcredentials.Config{
+		ClientID:     o.Sso.ClientID,
+		ClientSecret: o.Sso.ClientSecret, //pragma: allowlist secret
+		TokenURL:     fmt.Sprintf("%s/auth/realms/%s/protocol/openid-connect/token", o.Sso.Endpoint, o.Sso.Realm),
```
misberner, line 39: Would it make sense to use the OIDC library here, or is that overkill? There's a discovery document at https://sso.redhat.com/auth/realms/redhat-external/.well-known/openid-configuration , so using `https://sso.redhat.com/auth/realms/redhat-external` as the OIDC issuer should work.

pkg/client/fleetmanager/auth_rhsso.go
```diff
@@ -19,35 +21,35 @@ var (
...
+		Scopes:       []string{"openid"},
```
misberner, line 40: Given that the intention is to remove token refresher, I would expect that the above token URL gives us a refresh token, that the oauth2 `TokenSource` uses to transparently issue a new token, if necessary. Is that true? And do we indeed get a refresh token _without_ specifying `offline_access` here? That would be a bit unexpected.

misberner (changes requested):

pkg/client/fleetmanager/auth_rhsso.go
dhaus67: I think I gotta clear up the initial naming. The `token-refresher` used the same oauth2 libary `oauth2/clientcredentialsgrant`.

The speciality about it is that [the new property of the reuseTokenSource](https://github.com/golang/oauth2/blob/master/oauth2.go#L289) isn't using [a tokenRefresher](https://github.com/golang/oauth2/blob/master/oauth2.go#L255), which would perform a `refresh_token` grant, but actually using [the clientcredentials.tokenSource](https://github.com/golang/oauth2/blob/master/clientcredentials/clientcredentials.go#L83) (which does not perform the `refresh_token` grant but obtains a new access token using the configured client credentials).

It's been the same behavior before compared to [the "original" token-refresher](https://github.com/observatorium/token-refresher/blob/master/main.go#L208). The name implies that it will perform a `refresh_token` grant, but actually does not.

Does that clear things up?

pkg/client/fleetmanager/auth_rhsso.go
dhaus67: Don't think it's overkill at all, it actually makes it more robust compared to what we currently do with manually constructing URLs.

pkg/client/fleetmanager/auth_rhsso.go
misberner: Yeah, makes sense. Thanks!

misberner (approved):

PR #434, opened 2022-09-22
Rename dinosaur to central

dhaus67:
## Description

Rename _most_ of the structs, variables, comments etc. from Dinosaur/dinosaur to Central/central.

Rename was performed by going through each directory, searching for `dinosaur` **case-insensitive**, and replacing the variables after reviewing their usage (i.e. avoid renaming database specific fields). For this, the Goland `Refactor` functionality was used.

The files / folders that were skipped for now are:
- Anything under `config`.
- Metrics defined under `internal/central/constants/metrics.go` and `internal/central/pkg/metrics`.
- Anything under `docs/legacy`.
- Mocks for observatorium under `pkg/client/observatorium`.
- Some tests, specifically:
  - `pkg/api/cluster_types_test.go`
  - `pkg/shared/secrets/secrets_test.go`
  -  Tests under `internal/central/test` (partially commented out ones as well).

## Test manual

```
# Run the following command to search for the names:
grep -Hrn 'dinosaur' .

# You will see the list of files that still reference dinosaur. Parts of the files are commented out, other parts not.

```

PR #436, opened 2022-09-22
ROX-12742: rename "dinosaur" to "central" in error messages and logs

vladbologa:
## Description
I went through all the string literals in the codebase that contain "dinosaur", and changed them to "central" (or "Central", or "centrals") in the following cases:
- they are part of error messages
- they are part of log messages
- are outputted by CLI tools

Left everything else untouched.

~~- [ ] CI and all relevant tests are passing~~

### DIFF
--- a/internal/dinosaur/pkg/services/data_plane_dinosaur.go
+++ b/internal/dinosaur/pkg/services/data_plane_dinosaur.go
@@ -21,14 +21,16 @@ import (
 type dinosaurStatus string
 
 const (
-	statusInstalling         dinosaurStatus = "installing"
-	statusReady              dinosaurStatus = "ready"
-	statusError              dinosaurStatus = "error"
-	statusRejected           dinosaurStatus = "rejected"
-	statusDeleted            dinosaurStatus = "deleted"
-	statusUnknown            dinosaurStatus = "unknown"
-	dinosaurOperatorUpdating string         = "DinosaurOperatorUpdating"
-	dinosaurUpdating         string         = "DinosaurUpdating"
+	statusInstalling dinosaurStatus = "installing"
+	statusReady      dinosaurStatus = "ready"
+	statusError      dinosaurStatus = "error"
+	statusRejected   dinosaurStatus = "rejected"
+	statusDeleted    dinosaurStatus = "deleted"
+	statusUnknown    dinosaurStatus = "unknown"
+
+	// TODO: Renaming these dinosaurs will require a DB migration step
+	dinosaurOperatorUpdating string = "DinosaurOperatorUpdating"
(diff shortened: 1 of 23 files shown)


internal/dinosaur/pkg/services/data_plane_dinosaur.go
```diff
@@ -27,8 +27,8 @@ const (
 	statusRejected           dinosaurStatus = "rejected"
 	statusDeleted            dinosaurStatus = "deleted"
 	statusUnknown            dinosaurStatus = "unknown"
-	dinosaurOperatorUpdating string         = "DinosaurOperatorUpdating"
-	dinosaurUpdating         string         = "DinosaurUpdating"
+	dinosaurOperatorUpdating string         = "CentralOperatorUpdating"
+	dinosaurUpdating         string         = "CentralUpdating"
```
dhaus67, line 31: I think those will be a bit more involving for refactor.

If you look at their reference (e.g. [here](https://github.com/stackrox/acs-fleet-manager/blob/main/internal/dinosaur/pkg/services/data_plane_dinosaur.go#L158)), they are compared against the [dbapi.DataPlaneCentralStatus.Conditions[].Reason](https://github.com/stackrox/acs-fleet-manager/blob/main/internal/dinosaur/pkg/api/dbapi/data_plane_central_status.go#L20), which as the name suggest is stored within the DB.

We have two options for this:
- Skip refactoring this value altogether and keep the original value for now (incl. a TODO marker to refactor in the future)
- Add a migration similar to [this one](https://github.com/stackrox/acs-fleet-manager/blob/main/internal/dinosaur/pkg/migrations/202209141000_add_central_auth_lease_type.go). Within the migration, we should rename all the previous values `DinosaurOperatorUpdating/DinosaurUpdating` to their new values `CentralOperatorUpdating/CentralUpdating`.

I'd prefer the migration + renaming, but it's up to you. Either way it's probably best suited to do this in a separate PR so its more explicit.

dhaus67 (changes requested):
Blocking on one comment that highlights a non-backward compatible change.

Otherwise, LGTM! I'd say we prefer to merge this PR over [pull/434](https://github.com/stackrox/acs-fleet-manager/pull/434) since the latter will probably take more review cycles and is up for additional splitting.

internal/dinosaur/pkg/services/data_plane_dinosaur.go
vladbologa: Thanks for reviewing! I will do the migration in a separate PR, I agree that it makes more sense that way.

dhaus67 (approved):

PR #439, opened 2022-09-26
ROX-11831: Test Env: Simplify by hard-coding kubernetes.default.svc as data-plane cluster 

mtesseract:
## Description

This is a long-overdue simplification of the test environment code.

I finally decided to implement this because alpine-image related delays popped up in my workflow during testing.

https://issues.redhat.com/browse/ROX-11831

## Test manual

```
# To run tests locally run:
```


dev/env/scripts/lib.sh
```diff
@@ -252,25 +249,17 @@ assemble_kubeconfig() {
...
-    fi
-
     config=$(
         cat <<EOF
 apiVersion: v1
 clusters:
-    - $CLUSTER
+    - cluster:
+        server: kubernetes.default.svc
+      name: \"$CLUSTER_NAME\"
```
johannes94, line 262: nit: are you sure you need the escapes here? the other variables are also not escaped e.g $NEW_CONTEXT_NAME

johannes94 (approved):
LGTM

dev/env/scripts/lib.sh
mtesseract: In a situation like this I would highly recommend to add quotation marks to enforce this value being interpreted as a string. I don't know if it is possible to have a cluster name which could also be interpreted as a number, it should certainly be possible (I assume) to have a cluster name "true" which could then be interpreted as a boolean?

PR #445, opened 2022-09-27
ROX-12829: Fix Central Data URL

mtesseract:
## Description

See https://issues.redhat.com/browse/ROX-12829

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/internal/dinosaur/pkg/presenters/dinosaur.go
+++ b/internal/dinosaur/pkg/presenters/dinosaur.go
@@ -7,6 +7,10 @@ import (
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/public"
 )
 
+const (
+	sensorDataPort = 443 // The port for connecting sensor to the data URL.
+)
+
 // ConvertDinosaurRequest from payload to DinosaurRequest
 func ConvertDinosaurRequest(dinosaurRequestPayload public.CentralRequestPayload, dbDinosaurrequest ...*dbapi.CentralRequest) *dbapi.CentralRequest {
 	// TODO implement converter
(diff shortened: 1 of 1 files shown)


johannes94 (approved):
LGTM

internal/dinosaur/pkg/presenters/dinosaur.go
```diff
@@ -33,7 +37,7 @@ func PresentCentralRequest(request *dbapi.CentralRequest) public.CentralRequest
 		Owner:          request.Owner,
 		Name:           request.Name,
 		CentralUIURL:   fmt.Sprintf("https://%s", request.GetUIHost()),
-		CentralDataURL: fmt.Sprintf("https://%s", request.GetDataHost()),
+		CentralDataURL: fmt.Sprintf("%s:%d", request.GetDataHost(), sensorDataPort),
```
kovayur, line 40: This is not URL anymore, need to think about renaming the field. Raised ROX-12830 for this.

kovayur (approved):
LGTM

internal/dinosaur/pkg/presenters/dinosaur.go
mtesseract: agreed, thanks for creating the ticket.

mtesseract: On the other hand, when we fix sensor to also understand an https URL, then the above change could be reverted and the field wouldn't need to be renamed?

kovayur: Both options will work, anyway let's document it first and then decide how to deal with it properly.

johannes94: I'd not change this fields name right now, because this would probably also require a change to the console UI

rukletsov:
For the record, this PR has been reduced to the actual fix. Update to tests will follow.

PR #419, opened 2022-09-19
Reduce fleetshard-sync log verbosity

SimonBaeumer:
## Description
This pull request reduces the log verbosity. As it's helpful to debug the system then terminals and not being flooded by log messages with no debugging value.

## Checklist (Definition of Done)

- [ ] Unit and integration tests added
- [ ] Documentation added if necessary
- [ ] CI and all relevant tests are passing
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.

## Test manual

 - CI

PR #450, opened 2022-09-27
Ensure log entries point to relevant code lines

rukletsov:
## Description
https://github.com/stackrox/acs-fleet-manager/pull/419 introduced wrappers around glog calls. This is a bit unfortunate because:
* the codebase now uses `glog.*` and `log.*` variants
* log entries created via wrappers will point wrapper code and not the actual caller, e.g.: `E0927 11:29:48.933799       1 log.go:36] Unexpected error occurred ...`, as observed in [ROX-12836](https://issues.redhat.com/browse/ROX-12836)

This PR:
* removes the wrappers in favor of direct `glog.*` calls.

## Test manual

TODO(alexr): Carry out a manual test


fleetshard/pkg/log/log.go
```diff
@@ -1,42 +1,23 @@
...
+
+	"github.com/golang/glog"
+)
 
 const (
 	debugLogLevel = 10
 )
 
-// Info write info messages.
-func Info(args ...interface{}) {
```
dhaus67, line 10: Why not keep the wrapper and instead use `glog.Info/Error/WarningDepth` within the wrappers function?

Right now it feels weird to _know_ when to use what: `glog` for "normal" logging and `log` for debug logging - I'd rather have it consistent: either using `glog` all the time or the wrapper (in this case `log`).

dhaus67 (commented):
Need to run good old `go mod tidy` to add the missing go.sum entry.

fleetshard/pkg/log/log.go
```diff
@@ -1,42 +1,23 @@
...
-// Warning logs to the WARNING and INFO logs.
-func Warning(args ...interface{}) {
-	glog.Warning(args...)
+	// To avoid log entry point to this function, instruct glog to skip one
+	// extra stack frame.
+	//
+	// Note that `Verbose.InfoDepth()` is unavailable in "golang/glog" at the
+	// moment of writing. It is added to our private "glog" fork.
+
+	msg := fmt.Sprintf(format, args)
```
misberner, line 21: Careful, this passes a *single* argument of type `[]interface{}` to `Sprintf`. So `Debugf("%s %s", "foo", "bar")` will print as `"[foo bar] %!s(MISSING)"`. What you want is `args...`.

fleetshard/pkg/log/log.go
```diff
@@ -1,42 +1,23 @@
...
-	glog.Errorf(format, args...)
-}
-
-// Warning logs to the WARNING and INFO logs.
-func Warning(args ...interface{}) {
-	glog.Warning(args...)
+	// To avoid log entry point to this function, instruct glog to skip one
+	// extra stack frame.
+	//
+	// Note that `Verbose.InfoDepth()` is unavailable in "golang/glog" at the
```
misberner, line 18: This hardly seems to justify the fork. What you want is
```
if glog.V(debugLogLevel) {
  msg := fmt.Sprintf(format, args)
  glog.InfoDepth(1, msg)
}
```

That's because `fmt.Sprintf` potentially has non-trivial performance implications (e.g., rendering of a larger struct).

fleetshard/pkg/log/log.go
misberner: I agree with Daniel that it's weird to have `glog.Info..` for most stuff, and then `log.Debug` for debug levels. Is it really that much effort to write the extra `V(...)`? Is the worry a lack of consistency in verbosity levels?

misberner (changes requested):

fleetshard/pkg/log/log.go
rukletsov: This is embarrassing.

fleetshard/pkg/log/log.go
rukletsov: I went down a bunch of rabbit holes so I actually think wrappers are not that necessary. Updated the PR.

misberner (approved):

SimonBaeumer:
Thanks for fixing this @rukletsov. Did not recognized that the reported source file changed.

SimonBaeumer (approved):

PR #456, opened 2022-09-29
ROX-12640: Prefer schedule replicas on different nodes

kurlov:
## Description

- `topologyKey: "kubernetes.io/hostname"` will suggest the orchestrator to avoid deploying pods in the same host name
- drop `podAffinityTerm` because there is only one condition so specifying `weight` is redundant

Details https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

## Test manual

Will test stage deployment

### DIFF
--- a/templates/service-template.yml
+++ b/templates/service-template.yml
@@ -895,15 +895,13 @@ objects:
           affinity:
             podAntiAffinity:
               preferredDuringSchedulingIgnoredDuringExecution:
-              - podAffinityTerm:
-                  labelSelector:
-                    matchExpressions:
-                    - key: app
-                      operator: In
-                      values:
-                      - fleet-manager
-                  topologyKey: failure-domain.beta.kubernetes.io/zone
-                weight: 100
+                labelSelector:
+                  matchExpressions:
+                  - key: app
+                    operator: In
+                    values:
+                    - fleet-manager
+                topologyKey: "kubernetes.io/hostname"
           serviceAccount: fleet-manager
           serviceAccountName: fleet-manager
           volumes:


templates/service-template.yml
```diff
@@ -895,15 +895,13 @@ objects:
...
-                      - fleet-manager
-                  topologyKey: failure-domain.beta.kubernetes.io/zone
-                weight: 100
+                labelSelector:
+                  matchExpressions:
+                  - key: app
+                    operator: In
+                    values:
+                    - fleet-manager
+                topologyKey: "kubernetes.io/hostname"
```
ebensh, line 904: For my own information, how does this work with Multi-AZ setups? You could put 3 replicas onto 3 different nodes, but all those nodes could be in the same AZ?

Or is each "Node" replicated across each of 3 AZs?

templates/service-template.yml
```diff
@@ -895,15 +895,13 @@ objects:
           affinity:
             podAntiAffinity:
               preferredDuringSchedulingIgnoredDuringExecution:
-              - podAffinityTerm:
```
ebensh, line 898: Is dropping this (and the weight) intentional?

kurlov: Yes, it's intentional, we have just one condition and `podAffinityTerm` is only for `weight` there so it looks redundant

templates/service-template.yml
kurlov: Yes, all nodes could be in the same AZ. Originally KubeLinker complained about non `hostname` anti affinity. See rule [here](https://docs.kubelinter.io/#/generated/checks?id=no-anti-affinity).
We can introduce additional antiaffinity condition with [topology.kubernetes.io/region](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone) with lower `weight`.
In any case we should drop `failure-domain.beta.kubernetes.io/zone` because it's [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone)

kurlov: Should I add another condition in scope of this PR?

SimonBaeumer (approved):

PR #425, opened 2022-09-20
ROX-12157: Implement augmentWithDynamicAuthConfig

ivan-degtiarenko:
## Description
This PR augments central requests with Central IdP OIDC client data by using dynamic clients API provided to us by CIAM team.

Changes within this PR:
* call to RHSSO dynamic clients API to get OIDC client data
* GetIssuer() method in IAMService to avoid us depending on a config value
* Truncate name according to dynamic clients API limitation
* Limit client creation requests only to centrals with already set `host`

## Checklist (Definition of Done)

- [ ] Unit and integration tests added
- [ ] Documentation added if necessary
- [ ] CI and all relevant tests are passing
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.

## Test manual
0. **Remove orgId check:** https://github.com/stackrox/acs-fleet-manager/blob/667ae5a81952f3508742e905a6f57b502e1fbcd9/fleetshard/pkg/central/reconciler/init_auth.go#L156-L161 
1. Deploy `fleetshard-sync` to OSD cluster with `sso.stage.redhat.com` service account auth:
```
helm -n $NAMESPACE install rhacs-terraform dp-terraform/helm/rhacs-terraform/ \                                                                                                                    
  --set fleetshardSync.authType="RHSSO" \
  --set fleetshardSync.fleetManagerEndpoint=${FLEET_MANAGER_ENDPOINT} \
  --set fleetshardSync.clusterId=${CLUSTER_ID} \
  --set fleetshardSync.image="${FLEET_MANAGER_IMAGE}" \
  --set fleetshardSync.gitCommitSHA="wow" \
  --set fleetshardSync.gitDescribeTag="bowaoo" \
  --set acsOperator.enabled=true \
  --set acsOperator.source="${RHACS_OPERATOR_CATALOG_NAME}" \
  --set logging.enabled=false \
  --set acsOperator.version="v${RHACS_OPERATOR_CATALOG_VERSION}" \
  --set fleetshardSync.redHatSSO.clientId="${CLIENT_ID}" \
  --set fleetshardSync.redHatSSO.clientSecret="${CLIENT_SECRET}" \
  --set fleetshardSync.redHatSSO.endpoint=https://sso.stage.redhat.com \
--set observability.enabled=false
```
In general follow steps here:
https://github.com/stackrox/acs-fleet-manager/blob/main/docs/development/setup-developer-osd-cluster.md
2. Add `orgId` of fleetshardSync token to the list here:
[...]

PR #460, opened 2022-09-30
Delete dynamic OIDC client within deprovisioning

dhaus67:
## Description

Adds the ability to delete previously created OIDC clients using the dynamic client API.

The deletion will only be attempted if no static OIDC client is used.

## Test manual

1. Run fleet manager, fleetshard, operator. Ensure the correct settings are applied for dynamic OIDC client generation.
2. Create a central using i.e. `scripts/create-central.sh`.
3. Verify that an auth provider is created (ensure no log similar to "static config found" is present) and usable.
4. Delete the central instance and verify the OIDC client got deleted successfully.


dhaus67:
* main
  * **PR #414** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/414" target="_blank">Graphite</a> 
    * **PR #416** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/416" target="_blank">Graphite</a> 
      * **PR #460** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/460" target="_blank">Graphite</a>  👈
      * **PR #425** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/425" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/460?utm_source=stack-comment).

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_centrals_mgr.go
```diff
@@ -0,0 +1,129 @@
...
+	quotaService, factoryErr := k.quotaServiceFactory.GetQuotaService(api.QuotaType(central.QuotaType))
+	if factoryErr != nil {
+		return factoryErr
+	}
+	err := quotaService.DeleteQuota(central.SubscriptionID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete subscription id %s for central %s", central.SubscriptionID, central.ID)
+	}
+
+	if k.centralConfig.HasStaticAuth() {
```
ivan-degtiarenko, line 116: Maybe it's not that big of a deal, but I'm concerned that in case we re-deploy fleet-manager and switch from dynamic to static auth(for example, if dynamic clients do not work properly), we will leave behind orphaned clients which will not be removed.

Should we somehow identify from central request itself whether it was created with static or dynamic client?

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_centrals_mgr.go
```diff
@@ -0,0 +1,129 @@
...
+		glog.V(7).Infoln("static config found; no dynamic client will be deleted")
+	} else {
+		if err := k.dynamicAPI.DeleteDynamicClient(central.ClientID); err != nil {
+			return errors.Wrapf(err, "failed to delete dynamic OIDC client id %s for central %s",
+				central.ClientID, central.ID)
```
ivan-degtiarenko, line 121: Is there a reason why we prefer `central.ID` over `{central.Namespace}/{central.Name}`? Trying to figure out if I should fix this line in my PR: https://github.com/stackrox/acs-fleet-manager/pull/425/files#diff-789dbe395b2c0c640a766ed0ce12c89bf34a195dc04da3364710192030dfe2dfR140

ivan-degtiarenko (commented):
Overall looks good, one concern regarding static auth check

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_centrals_mgr.go
dhaus67: I aligned it with the other log statements within the workers.

I wouldn't choose `central.Namespace/central.Name` due to it being unambiguous, we would have to use `central.ClusterID/central.Namespace/central.Name` for complete uniqueness.
We can check whether this is something we want to change in the future and align everything towards the `central.ClusterID/central.Namespace/central.Name` tuple, but for now I'd like to be consistent with the existing log statements and keep `central.ID`.

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_centrals_mgr.go
dhaus67: Discussed this offline a bit, so posting it here for posterity:

This is definitely valid and something which _could_ happen.

IMO, the best solution for this would be to have an API with which one can determine whether a client ID belongs to a client which was dynamically created using the dynamic client API. This currently isn't possible.

To workaround this, I've added a bool property to the central request as proposed by @ivan-degtiarenko, which specifies whether a static client was used during creation or not.

The newly added migration will, by default, assume that every existing central request uses the static client (since this is something we have not yet deployed on any clusters).

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_centrals_mgr.go
```diff
@@ -0,0 +1,132 @@
...
+	err := quotaService.DeleteQuota(central.SubscriptionID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete subscription id %s for central %s", central.SubscriptionID, central.ID)
+	}
+
+	if central.UsesStaticClient {
+		glog.V(7).Infoln("central uses static client; no dynamic client will be attempted to be deleted")
+	} else {
+		if _, err := k.dynamicAPI.DeleteAcsClient(context.TODO(), central.ClientID); err != nil {
+			return errors.Wrapf(err, "failed to delete dynamic OIDC client id %s for central %s",
```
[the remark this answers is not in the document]
dhaus67: The way it currently works is that, yes, if any error occurred during a reconciliation, the reconciliation will be retried until the issue is fixed. The errors and number of failures will be added to the metrics ([see this code path](https://github.com/stackrox/acs-fleet-manager/blob/bad3fbe9c031d222b5fb1e775a7ec8a2862e2692/pkg/workers/reconciler.go#L76-L89)).

The way reconcilers are used within fleetshard-sync, ACS operator and the kubernetes controller-runtime is exactly as such, retry with potential back-offs until the reconciliation could be performed.

In case we have instances being stuck in the reconciliation, we have the ability to perform a "force delete" via the admin API introduced here: https://github.com/stackrox/acs-fleet-manager/blob/main/openapi/fleet-manager-private-admin.yaml#L262
However, there will be a reason the deletion didn't work in the first place, and orphaned resources are not dealt with in this API. There's a reason the reconciliation couldn't finish to begin with.

I'm not sure what to change in this behavior based on your comment, I don't agree with the sentiment that it "doesn't seem right".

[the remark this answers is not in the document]
rukletsov: Ok, so if we plan to infinitely retry, we should probably add an alert for reconciler failure count growing. Currently we don't have [anything like this](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/resources/observability/prometheusrules/acs-fleet-manager-production.prometheusrules.yaml).

> I don't agree with the sentiment that it "doesn't seem right".
What I mean is that there should be some recovery path from this error. Options that seem reasonable to me:
* retry finite number of times, give up, proceed with the cleanup => should not be that bad because RHSSO must detect and clean up orphans
* as above but with a clear signal that dynamic client has not been removed (a record in fleet DB, etc) so a human can follow up
* abort deletion and notify a human to intervene.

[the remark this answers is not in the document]
dhaus67: Created [ROX-12997](https://issues.redhat.com/browse/ROX-12997) as a follow-up on the alert creation.

For me, the recovery path from this error is the SOP for resolving the alert.
I have a hard time understanding why we would decide to go for such a complicated, custom solution for this specific worker and reconciliation.

Specifically:
> should not be that bad because RHSSO must detect and clean up orphans

Is this a given? Will the OIDC clients be deleted after some time if they weren't used for login? How would RH SSO know if those clients are "orphaned" from their point of view?

>(a record in fleet DB, etc) so a human can follow up

Could you explain why this is favorable over acting on an alert with sufficient information as to what the specific error is during deprovisioning?

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_dinosaurs_mgr.go
```diff
@@ -104,6 +114,17 @@ func (k *DeletingDinosaurManager) reconcileDeletingDinosaurs(dinosaur *dbapi.Cen
 		return errors.Wrapf(err, "failed to delete subscription id %s for central %s", dinosaur.SubscriptionID, dinosaur.ID)
 	}
 
+	if dinosaur.ClientOrigin == dbapi.AuthConfigStaticClientOrigin {
+		glog.V(7).Infof("central %s uses static client; no dynamic client will be attempted to be deleted",
+			dinosaur.ID)
+	} else {
+		if resp, err := k.dynamicAPI.DeleteAcsClient(context.Background(), dinosaur.ClientID); err != nil &&
+			resp.StatusCode != http.StatusNotFound {
```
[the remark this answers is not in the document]
dhaus67: Ah, this one shouldn't have been pushed, I was unsure about whether to add this or not since I saw some hand-rolled dynamic service accounts client have this behavior, but I don't think it makes sense for delete.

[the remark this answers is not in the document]
dhaus67: So, I initially was unsure about adding this. After talking more about the reconciliation [with Alex in this thread](https://github.com/stackrox/acs-fleet-manager/pull/460#discussion_r988952045) I realized that in case we have a reconciliation error, and the reconciliation will be retried:
If we e.g. fail not deleting the client but deleting the central from the database due to i.e. database being unresponsive, we would get a different error on the second reconciliation, since the client will already have been deleted.
I'm not sure we want this behavior when debugging issues, so hence not returning an error when the client cannot be found (from the workers POV, the goal has been achieved, the client does not exist anymore).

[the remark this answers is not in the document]
ivan-degtiarenko: It's a good reason not to fail on this. However, let's still log `not found` as it shouldn't be a regular situation.

ivan-degtiarenko (approved):
LGTM modulo rebase + one comment regarding "not found"

internal/dinosaur/pkg/migrations/202205101700_add_client_origin_to_centrals.go
```diff
@@ -0,0 +1,71 @@
...
+		Routes                        api.JSON
+		RoutesCreated                 bool
+		Namespace                     string
+		RoutesCreationID              string
+		DeletionTimestamp             *time.Time
+		AuthConfig
+	}
+
+	return &gormigrate.Migration{
+		ID: "202205101700",
```
rukletsov, line 57:
```suggestion
		ID: "202210051700",
```
also the file name : )

internal/dinosaur/pkg/migrations/202205101700_add_client_origin_to_centrals.go
```diff
@@ -0,0 +1,71 @@
...
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.AutoMigrate(&CentralRequest{}); err != nil {
```
rukletsov, line 59: I see you trust "auto migrations"...

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_dinosaurs_mgr.go
```diff
@@ -104,6 +114,21 @@ func (k *DeletingDinosaurManager) reconcileDeletingDinosaurs(dinosaur *dbapi.Cen
 		return errors.Wrapf(err, "failed to delete subscription id %s for central %s", dinosaur.SubscriptionID, dinosaur.ID)
 	}
 
+	if dinosaur.ClientOrigin == dbapi.AuthConfigStaticClientOrigin {
+		glog.V(7).Infof("central %s uses static client; no dynamic client will be attempted to be deleted",
+			dinosaur.ID)
+	} else {
```
rukletsov, line 120: This at least should be `else if`, I'd go with `switch {}` with the `default` clause returning an error.

rukletsov (changes requested):

internal/dinosaur/pkg/migrations/202205101700_add_client_origin_to_centrals.go
dhaus67: I don't really have a preference, we use both in the existing migrations. I've changed it to an explicit `AddColumn` though.

internal/dinosaur/pkg/workers/dinosaurmgrs/deleting_centrals_mgr.go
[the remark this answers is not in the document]
rukletsov: I haven't said it's favourable, but is an option. I think alert is just fine.

rukletsov (changes requested):

internal/dinosaur/pkg/migrations/202210051700_add_client_origin_to_centrals.go
```diff
@@ -0,0 +1,71 @@
...
+		Namespace                     string
+		RoutesCreationID              string
+		DeletionTimestamp             *time.Time
+		AuthConfig
+	}
+
+	return &gormigrate.Migration{
+		ID: "202210051700",
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.Migrator().AddColumn(&CentralRequest{}, "ClientOrigin"); err != nil {
```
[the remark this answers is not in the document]
dhaus67: Removed the `default:shared_static_rhsso` part since it shouldn't really be part of the table schema.

Instead did the following:
- Migrate the table schema (with `AddColumn`)
- Afterwards, update all the `CentralRequests` whose `client_origin` value is `NULL` and set it to `shared_static_rhsso`.

internal/dinosaur/pkg/migrations/202210051700_add_client_origin_to_centrals.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+
+			// Update only the central requests which do not already specify client_origin after migrating the table
+			// schema. For central requests which do not specify it, we can assume shared_static_sso, since the field
+			// was not available before adding the possibility to create centrals with a dynamic OIDC client.
+			// **Note**: This update is _not_ done within a transaction since we pass `UseTransaction=false` as
+			// gormigrate option, meaning this will be on a best-effort basis, but the risk of not doing it within a
+			// transaction is minor.
+			err := tx.Model(&CentralRequest{}).
+				Where("client_origin IS NULL").
```
ivan-degtiarenko, line 69: In which cases `client_origin` will not be null?

dhaus67: We were initially talking about whether we mistakenly execute migrations twice.
Theoretically, it shouldn't happen that `client_origin` will be anything other than NULL, but I'd like to be a bit more expressive here.
I've added a comment which highlights that we shouldn't expect it to be.

internal/dinosaur/pkg/migrations/202210051700_add_client_origin_to_centrals.go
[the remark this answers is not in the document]
rukletsov: Thank you, Daniel!

internal/dinosaur/pkg/migrations/202210051700_add_client_origin_to_centrals.go
rukletsov: To make it clear: having records with set `client_origin` seems practically possible but highly unlikely. Because we have multiple shards of fleet manager running and the db migration does not happen in a transaction (I don't understand why, maybe we should ask kafkanians for reasons), a new record with `client_origin` may be inserted after the schema update and before the data update.

rukletsov (approved):

PR #465, opened 2022-10-04
ROX-12650: Update Helm resources in central reconciler

vladbologa:
## Description
Currently Helm resources are not updated in the fleetshard-sync reconciliation loop, they are only created if they don't exist.

This change would update Helm resources if a new fleetshard-sync version is deployed (the resources are currently embedded in the binary,
so they can only change with a binary update).

This PR is replacing https://github.com/stackrox/acs-fleet-manager/pull/411

~~- [x] Documentation added if necessary~~
~~- [x] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.~~

## Test manual

Only did manual testing so far, because I'm not sure yet if the implemented behaviour is acceptable.

Set up a local cluster with `colima`, then:
- ran the operator outside the cluster (with `make install run` from `stackrox/operator`)
- ran the bootstrap script: `INSTALL_OPERATOR=NO INSTALL_OLM=NO ./dev/env/scripts/bootstrap.sh`
- started the services with `INSTALL_OPERATOR=NO ./dev/env/scripts/up.sh`
- make a central request to fleet-manager with `./scripts/create-central.sh`

Then to test that updating of the chart resources works:
- added annotation to `egress-proxy`
- simulated an update of `fleetshard-sync` with `kubectl -n acsms rollout restart deployment fleetshard-sync`
- checked that `egress-proxy` is updated (via logs, also checked that the annotation that I added was removed)

The current behaviour is as follows:
- chart resources are embedded in the `fleetshard-sync` binary, so they can only change when a new binary is deployed
- chart resources are installed on the first run of `Reconcile`, or, if they exist, they are updated. After the central is ready, they are no longer updated, unless the Central objects changes. 
They are also updated if `fleetshard-sync` is restarted (so in case a new version of `fleetshard-sync` is installed, the updated chart resources *will* be installed - which is what I think 
we're trying to achieve) 
- changes (manual or made by other controllers) to the chart resources will be lost if `fleetshard-sync` is updated or restarted, but also if the Central is changed (is this acceptable for 
now?)
- I think it's not likely to have hot update loops right now, because if the managed Central is ready and not changed (see line 84) the `Reconcile` function just exits 

TBD: add some automated tests, if this approach is acceptable


SimonBaeumer:
**Testing notes:**
 - Does a second reconcile upgrade the resources? Helm should not upgrade resources which are in their final state.

**Behaviour:**
> chart resources are installed on the first run of Reconcile

This not tied to the first Reconcile run but for the chart installation. If the chart is installed it can only be upgraded or deleted.
Additionally if that reconcile run failed according to the description for installation would not happen.

> They are also updated if fleetshard-sync is restarted
why should the resources be upgraded on every fleetshard restart?

> changes (manual or made by other controllers) to the chart resources will be lost if fleetshard-sync is updated or restarted, but also if the Central is changed (is this acceptable for
now?)
Yes, the resources are managed by fleetshard. If fleetshard updates these resources changes made manually are reverted.
The same applies if a Pod or ReplicaSet is manually edited but managed by a Deployment.

> I think it's not likely to have hot update loops right now, because if the managed Central is ready and not changed (see line 84) the Reconcile function just exits
What is a "hot update loop"?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -453,12 +452,20 @@ func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remot
...
+				return fmt.Errorf("failed to update object %s/%s of type %v: %w", key.Namespace, key.Namespace, obj.GroupVersionKind(), err)
+			}
+
 			continue
 		}
 		if !apiErrors.IsNotFound(err) {
 			return fmt.Errorf("failed to retrieve object %s/%s of type %v: %w", key.Namespace, key.Name, obj.GroupVersionKind(), err)
 		}
 		err = r.client.Create(ctx, obj)
+		glog.Infof("Creating object %s/%s", obj.GetNamespace(), obj.GetName())
```
SimonBaeumer, line 468: Reducing verbosity in production systems by setting `V(10)`:

```suggestion
		glog.v(10).Infof("Creating object %s/%s", obj.GetNamespace(), obj.GetName())
```

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -453,12 +452,20 @@ func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remot
 		out.SetGroupVersionKind(obj.GroupVersionKind())
 		err := r.client.Get(ctx, key, &out)
 		if err == nil {
+			glog.Infof("Updating object %s/%s", obj.GetNamespace(), obj.GetName())
```
SimonBaeumer, line 455:
```suggestion
			glog.V(10).Infof("Updating object %s/%s", obj.GetNamespace(), obj.GetName())
```

SimonBaeumer (changes requested):
Thanks for taking on this @vladbologa.
Added some comments in the code.

vladbologa:
@SimonBaeumer To make things clearer, in this current implementation I mostly made the installation / upgrading of chart resources to follow the behaviour of Central CRs. But probably also that one is not truly correct.

Let's follow the happy path for now. If you create a central, on the first run of Reconcile you get something like this:

```
I1004 14:17:55.305674       1 runtime.go:103] Received 1 centrals
I1004 14:17:55.306263       1 runtime.go:113] Start reconcile central rhacs-ccu40kof57v001t0og50/test-central-1
I1004 14:17:55.755537       1 reconciler.go:468] Creating object rhacs-ccu40kof57v001t0og50/egress-proxy-config
I1004 14:17:55.847913       1 reconciler.go:468] Creating object rhacs-ccu40kof57v001t0og50/egress-proxy
I1004 14:17:55.862879       1 reconciler.go:468] Creating object rhacs-ccu40kof57v001t0og50/egress-proxy
I1004 14:17:55.872082       1 reconciler.go:468] Creating object rhacs-ccu40kof57v001t0og50/egress-proxy
I1004 14:17:55.885786       1 reconciler.go:201] Creating central rhacs-ccu40kof57v001t0og50/test-central-1
I1004 14:17:55.898309       1 reconciler.go:205] Central rhacs-ccu40kof57v001t0og50/test-central-1 created
```

Then every 10 seconds, until the Central is ready, there's this:
```
I1004 14:18:55.455700       1 runtime.go:103] Received 1 centrals
I1004 14:18:55.455874       1 runtime.go:113] Start reconcile central rhacs-ccu40kof57v001t0og50/test-central-1
I1004 14:18:55.650730       1 reconciler.go:455] Updating object rhacs-ccu40kof57v001t0og50/egress-proxy-config
I1004 14:18:55.750160       1 reconciler.go:455] Updating object rhacs-ccu40kof57v001t0og50/egress-proxy
I1004 14:18:55.760350       1 reconciler.go:455] Updating object rhacs-ccu40kof57v001t0og50/egress-proxy
I1004 14:18:55.849530       1 reconciler.go:455] Updating object rhacs-ccu40kof57v001t0og50/egress-proxy
I1004 14:18:55.859002       1 reconciler.go:208] Update central rhacs-ccu40kof57v001t0og50/test-central-1
```

So the chart resources will be updated several times. I don't think this has side effects, because we're installing the same objects. The same happens (and happened before) to Centrals.

After the central is ready, there will be no more updates, due to this check at the beginning of `Reconcile`:
```
	if !changed && r.wantsAuthProvider == r.hasAuthProvider && isRemoteCentralReady(remoteCentral) {
		return nil, ErrCentralNotChanged
	}
```

The hash of the central object is stored in memory. Which is why if `fleetshard-sync` is restarted, `changed` will be false and it'll do another Update of both Central and the chart resources.

I could store the hashes of the rendered chart resources in memory, like we do for centrals, and then I think I could avoid doing needless updates.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -453,12 +452,20 @@ func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remot
...
+			obj.SetResourceVersion(out.GetResourceVersion())
+			err := r.client.Update(ctx, obj)
```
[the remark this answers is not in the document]
vladbologa: > Can you add a protection to the update call that not every time the update is called? I.e. with a hash in an annotation like @misberner described before.

I'm considering 2 possibilities now:
1. We save the hash of the applied resources in-memory, like we do for Centrals. This will prevent manual modifications, and could cause hot update loops, and will also cause one extra update when `fleetshard-sync` is restarted.
2. If I understood correctly what @misberner was suggesting, I would apply the resource and then store the hash as an annotation. Then on reconcile, I would compare the hash of the resources that come with the `fleetshard-sync` binary and update only if they are different. This can happen IMO only when a new version of `fleetshard-sync` is deployed.

So option (1) would not allow manual or external modifications to the resources. Option (2) would allow manual or external modifications, and would not perform an upgrade unless there's a new version of the chart resources.

Does (2) sound like indeed the one we want? There could also be a variation of (2) where `fleetshard-sync` could update the resources even if the hash annotation says we shouldn't, if the object is in a failed state. I'm open to suggestions here if I'm missing something.

> Also you described in you PR description that the `egress-proxy` now has an annotation. I could not find it. Can you point me to it?

That's just a manual annotation I added during manual testing, to verify that the object was updated. (i.e. by checking that the annotation disappears)

SimonBaeumer:
> I could store the hashes of the rendered chart resources in memory, like we do for centrals, and then I think I could avoid doing needless updates.

Why in memory if we can apply on the resource instead?
The reason for the Central hash to be in memory is that we can avoid the Reconcile there, here we can skip resources on a resource level.

fleetshard/pkg/central/reconciler/reconciler.go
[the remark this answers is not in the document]
SimonBaeumer: As discussed offline, no need to implement an update protection here. If an upgrade is necessary we will fix it.
Hot reconcile loops don't affect here as the Reconcile is triggered based on changes on the Central object received from the fleet-manager.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -435,6 +435,12 @@ func (r *CentralReconciler) ensureCentralCRDeleted(ctx context.Context, central
 }
 
 func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remoteCentral private.ManagedCentral) error {
+	if r.hasChartResources {
```
SimonBaeumer, line 438: I don't understand this here?
Also it introduces state in the reconciler which because it introduces dependencies between independent Reconcile runs.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -337,6 +337,46 @@ func TestChartResourcesAreAddedAndRemoved(t *testing.T) {
...
+	err = fakeClient.Update(context.TODO(), &dummySvc)
+	assert.NoError(t, err)
+
+	_, err = r.Reconcile(context.TODO(), simpleManagedCentral)
+	require.NoError(t, err)
+	err = fakeClient.Get(context.TODO(), dummySvcKey, &dummySvc)
+	assert.NoError(t, err)
+
+	// check that the chart resource was not modified after Reconcile, because chart
+	// resources are normally not updated once installed
```
SimonBaeumer, line 366: But the resources should be updated no?
This seems like a misunderstanding between us, my understanding was that resources are always updated if a Helm update is necessary.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -337,6 +337,46 @@ func TestChartResourcesAreAddedAndRemoved(t *testing.T) {
 	assert.True(t, k8sErrors.IsNotFound(err))
 }
 
+func TestChartResourcesAreAddedAndUpdated(t *testing.T) {
+	chart, err := charts.LoadChart(testdata, "testdata/tenant-resources")
```
SimonBaeumer, line 341: Testing looks very good, thanks for implementing it!

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
vladbologa: I was trying to avoid updating the helm chart several times. But perhaps you are right and this just complicates things. I removed it.

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: I don't think they should be updated, but also nothing bad happens if they are.
The resources are installed the first time `Reconcile` is called for this Central. Then whenever `Reconcile` is called again, if the central has been changed or is not ready yet, they will be updated.

Imo they should only be updated when fleetshard-sync restarts (because it might be a newer version). So that's what I was trying to achieve here. But I removed it, if it just complicates the logic by adding state.

SimonBaeumer (approved):

PR #472, opened 2022-10-07
ROX-12990: Configure default Central resources

mtesseract:
## Description

This is part of https://issues.redhat.com/browse/ROX-12990.

Configure default central resources in service template. Can be easily adjusted in the future.
This is for *new* tenants. Currently effective defaults are in `internal/dinosaur/pkg/defaults`.

Essentially: new configuration reflects that current CPU usage is <= old defaults and current memory usage is >= old defaults.

Especially watch out for copy and paste bugs.

The new settings would be automatically deployed on staging, I believe.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/templates/service-template.yml
+++ b/templates/service-template.yml
@@ -396,7 +396,7 @@ parameters:
 
 - name: CENTRAL_CPU_LIMIT
   displayName: Default CPU limit for central deployments for newly created tenants
-  description: Default CPU request for central deployments for newly created tenants
+  description: Default CPU limit for central deployments for newly created tenants
   value: "250m"
 
 - name: CENTRAL_MEMORY_LIMIT
@@ -435,8 +435,8 @@ parameters:
   value: "1G"
 
 - name: SCANNER_DB_CPU_LIMIT
-  displayName: Default CPU request for scanner-db deployments for newly created tenants
-  description: Default CPU request for scanner-db deployments for newly created tenants
+  displayName: Default CPU limit for scanner-db deployments for newly created tenants
+  description: Default CPU limit for scanner-db deployments for newly created tenants
   value: "500m"
 
 - name: SCANNER_DB_MEMORY_LIMIT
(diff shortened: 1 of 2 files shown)


templates/service-template.yml
```diff
@@ -384,6 +384,66 @@ parameters:
   description: SupportedProviders configuration file, that it's passed to the fleet manager executable in the flag --providers-config-file
   value: /config/provider-configuration.yaml
 
+- name: CENTRAL_CPU_REQUEST
+  displayName: Default CPU request for central deployments for newly created tenants
```
kurlov, line 388: Nitpick: I think `displayName` should be brief. it could be just `Central CPU request` and `description` will contain all the details. This way `displayName` doesn't duplicate `description`. We follow this pattern for other envs. Another option would be just drop `displayName` if it's too redundant. What do you think?
The same nitpick applies to other new envs

kurlov (commented):
Looks great! Just one small comment

templates/service-template.yml
SimonBaeumer: +1

templates/service-template.yml
```diff
@@ -1050,6 +1110,33 @@ objects:
             env:
               - name: "OCM_ENV"
                 value: "${ENVIRONMENT}"
+
+              - name: CENTRAL_CPU_REQUEST
```
SimonBaeumer, line 1114: Are the env variables already applied? The diff does not contain how they are applied.

mtesseract: This is my understanding.

templates/service-template.yml
mtesseract: ok, changed.

templates/service-template.yml
SimonBaeumer: Is it applied or not? 🤔

mtesseract: Yes, it is my understanding that this YAML *defines* how the deployments should look like on our clusters.
I guess if something would be missing here @kurlov would have added a comment about this.

SimonBaeumer (approved):

mtesseract:
Don't merge as of now, there's a typo, will fix later.

PR #473, opened 2022-10-07
ROX-12157: remove central-idp-client-id from stage.go

ivan-degtiarenko:
## Description
As part of integrating with dynamic client registration endpoint, we need to properly configure `qaprod` environment. To do that, we need to disable static authentication in `qaprod` environment(see https://github.com/stackrox/acs-fleet-manager/blob/b4cf67d02c72c6c144e39a744aa0467a5d252ed1/internal/dinosaur/pkg/config/central.go#L107-L115). `qaprod` environment uses `stage` label, so we need to clean `central-idp-client-id` from `stage.go`

## Test manual

1. Executed `OCM_ENV=stage ./fleet-manager serve` and verified it started successfully

### DIFF
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -108,7 +108,7 @@ func (c *CentralConfig) ReadFiles() error {
 // specified and false otherwise.
 func (c *CentralConfig) HasStaticAuth() bool {
 	// We don't look at other integral parts of the auth config like
-	// RhSsoIssuer or RhSsoClientSecret. Failure to provide a working auth
+	// CentralIDPIssuer or CentralIDPClientSecret. Failure to provide a working auth
 	// configuration should not mask an intent to use static configuration.
 	return c.CentralIDPClientID != ""
 }
--- a/internal/dinosaur/pkg/environments/stage.go
+++ b/internal/dinosaur/pkg/environments/stage.go
@@ -17,8 +17,6 @@ func NewStageEnvLoader() environments.EnvLoader {
 		"additional-sso-issuers-file":         "config/additional-sso-issuers.yaml",
 		"jwks-file":                           "config/jwks-file-static.json",
 		"fleetshard-authz-config-file":        "config/fleetshard-authz-org-ids-development.yaml",
-		"central-idp-client-id":               "rhacs-ms-dev",
-		"central-idp-issuer":                  "https://sso.stage.redhat.com/auth/realms/redhat-external",
 		"admin-authz-config-file":             "config/admin-authz-roles-dev.yaml",
 	}
 }


rukletsov (approved):

internal/dinosaur/pkg/config/central.go
```diff
@@ -108,7 +108,7 @@ func (c *CentralConfig) ReadFiles() error {
 // specified and false otherwise.
 func (c *CentralConfig) HasStaticAuth() bool {
```
dhaus67, line 109: With the dynamic client config options now being "active", we should try and have a verification similar to this in the `pkg/client/iam/config.go`.
Currently, the `redhat-sso-client-id/secret-file` is [treated as optional](https://github.com/stackrox/acs-fleet-manager/blob/e2d1c7e5431049427012fab186df91755c82b06d/pkg/shared/config.go#L58-L61), but the challenge here is that we have to distinguish that _either_ of the two is set and used. Currently, I don't see this being handled, so we probably should use this PR to polish it.

internal/dinosaur/pkg/environments/stage.go
```diff
@@ -17,8 +17,6 @@ func NewStageEnvLoader() environments.EnvLoader {
 		"additional-sso-issuers-file":         "config/additional-sso-issuers.yaml",
 		"jwks-file":                           "config/jwks-file-static.json",
 		"fleetshard-authz-config-file":        "config/fleetshard-authz-org-ids-development.yaml",
-		"central-idp-client-id":               "rhacs-ms-dev",
```
dhaus67, line 20: I suppose we agree on keeping static within dev environments by default, instead of switching this to dynamic clients, right?

rukletsov: I'd say so.

internal/dinosaur/pkg/config/central.go
rukletsov: > we have to distinguish that either of the two is set and used

Why so? Imagine a case when dynamic clients stop working, so we temporarily add static config without removing redhatsso service account.

dhaus67: I think that case is fine and correctly handled.
I don't see the case being handled where static config isn't set _and_ redhatsso service account isn't set.
Since `HasStaticAuth` will yield false, we currently assume that dynamic config is used (e.g. [within the provisioning](https://github.com/stackrox/acs-fleet-manager/blob/e2d1c7e5431049427012fab186df91755c82b06d/internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go#L98-L104) as well as within the deletion). I'd like us to figure that out when fleet manager is starting, instead of waiting for the first central creation request being reconciled.

ivan-degtiarenko: I think it would be nice to check that at least one of auth config augmentation options is configured properly. However, the problem with that is that currently they are located in separate config structs, and are read independently. I also think it would be pretty hard to change that.

What we can do is fail **worker creation** if none of the augmentation methods is set correctly. I'll check if it makes `fleet-manager` exit.

ivan-degtiarenko: It does.

cmd/fleet-manager/main_test.go
```diff
@@ -37,3 +52,26 @@ func TestInjections(t *testing.T) {
 	env.MustResolve(&workerList)
 	Expect(workerList).To(HaveLen(9))
 }
+
+func createMockSecretFile() *os.File {
+	file, err := ioutil.TempFile("", "idp-client-secret-")
```
dhaus67, line 57: You could use a combination of `t.TempDir()` and `os.CreateTemp(t.TempDir(), "file")` here instead.

1. [t.TempDir()](https://pkg.go.dev/testing#T.TempDir) will create a temporary test directory bound to the current test execution. It will also be automatically cleaned after the test execution.
2. `os.CreateTemp` is the non-deprecated version of `ioutil.TempFile`. I know we have a lot of references in the template for `ioutil` but we shouldn't introduce more of those IMO.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
```diff
@@ -37,8 +39,14 @@ type CentralAuthConfigManager struct {
 var _ workers.Worker = (*CentralAuthConfigManager)(nil)
 
 // NewCentralAuthConfigManager creates an instance of this worker.
-func NewCentralAuthConfigManager(centralService services.DinosaurService, iamConfig *iam.IAMConfig, centralConfig *config.CentralConfig) *CentralAuthConfigManager {
-	dynamicClientsAPI := dynamicclients.NewDynamicClientsAPI(iamConfig.RedhatSSORealm)
+func NewCentralAuthConfigManager(centralService services.DinosaurService, iamConfig *iam.IAMConfig, centralConfig *config.CentralConfig) (*CentralAuthConfigManager, error) {
```
dhaus67, line 42: Since I struggle a bit with the whole dependency injection part sometimes and what it exactly does, would be good to have a clarifying comment that mentions it will lead to an exit of the process due to the way the function is used.

cmd/fleet-manager/main_test.go
```diff
@@ -37,3 +54,26 @@ func TestInjections(t *testing.T) {
 	env.MustResolve(&workerList)
 	Expect(workerList).To(HaveLen(9))
 }
+
+func createMockSecretFile() *os.File {
```
dhaus67, line 58: Should be removed, no? Don't see any references of it.

dhaus67 (approved):
LGTM!
However, let's hold off on merging this before we merged #460 and removed the client-idp-secret in stage vault as said in the description.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
```diff
@@ -47,9 +56,34 @@ func NewCentralAuthConfigManager(centralService services.DinosaurService, iamCon
...
+		realmConfig:             realmConfig,
 		dynamicClientsAPIClient: dynamicClientsAPI,
+	}, nil
+}
+
+func validateAugmentationIsConfigured(realmConfig *iam.IAMRealmConfig, centralConfig *config.CentralConfig) error {
+	if centralConfig.HasStaticAuth() {
+		return nil
 	}
+	return validateDynamicAugmentationIsConfigured(realmConfig)
```
rukletsov, line 68: I don't think that auth config mgr should perform such validation. Instead, it should call `IAMRealmConfig.IsConfigured()` which would simply report whether there was an _intent_ to configure RH SSO service account.

`IAMConfig.ReadFiles()` should do the validation similar to what `validateDynamicAugmentationIsConfigured()` does to ensure that iff the user wants to configure RH SSO service acc, they do it properly.

rukletsov (changes requested):

pkg/client/iam/config.go
```diff
@@ -134,6 +163,12 @@ func (ic *IAMConfig) ReadFiles() error {
 
 	ic.RedhatSSORealm.setDefaultURIs(ic.SsoBaseURL)
 	ic.InternalSSORealm.setDefaultURIs(ic.InternalSsoBaseURL)
+	if err := ic.RedhatSSORealm.validateConfiguration(); err != nil {
+		return fmt.Errorf("validating external RH SSO realm config: %w", err)
+	}
+	if err := ic.InternalSSORealm.validateConfiguration(); err != nil {
+		return fmt.Errorf("validating internal RH SSO realm config: %w", err)
+	}
```
dhaus67, line 171: Maybe should add a notion that the internal SSO realm will not be configured with client credentials at the moment, as it will only serve as a configuration of the endpoints + realm. In the future it _may_ be extended with client secrets, but this depends.

dhaus67 (approved):

PR #477, opened 2022-10-10
ROX-13018: Add an ability to load secrets from Parameter Store for dev / local deployments

kovayur:
## Description
This PR adds loading from the Parameter Store on dev / local environment. This iterations covers **deployment** scripts only, CI and README docs will be covered in next PRs.

As part of this PR I also made changes in the `AUTH_TYPE` resolution logic:
1. For the fleetshard deployment I switched from `STATIC_TOKEN` to `RHSSO` to make the test environment closer to real ones.
2. For E2E tests now always the `STATIC_TOKEN` is used.
    - Alternative auth types require the same owner as in STATIC_ADMIN_TOKEN, making the choice of authentication type error-prone
    - Both tokens are to be placed in the Secured storage for dev env
    - For now the `AUTH_TYPE` only affects the deployment, not tests. I think it is incorrect to use the same variable for both tests and deployment because they are used for different API consumers.

Naming for client id and secret environment variables was also unified, in other words - `RHSSO_CLIENT_ID` and `RHSSO_CLIENT_SECRET` were renamed to the `RHSSO_SERVICE_ACCOUNT_CLIENT_ID` and `RHSSO_SERVICE_ACCOUNT_CLIENT_SECRET` respectively. This matches the environment variables names declared in the SSO service.

## Test manual
1. Install [chamber](https://github.com/segmentio/chamber) by `go install github.com/segmentio/chamber/v2@latest`
4. Configure access to AWS Parameter Store as described in the chamber repo documentation. It is recommended to use `aws-vault`.
5. Run `./dev/env/scripts/up.sh`


kovayur:
* main
  * **PR #477** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/477" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/477?utm_source=stack-comment).

e2e/README.md
```diff
@@ -23,7 +23,7 @@ $ go clean -testcache && CLUSTER_ID=1234567890abcdef1234567890abcdef \
   CLUSTER_ID=1234567890abcdef1234567890abcdef \
   STATIC_TOKEN=$(bw get item "64173bbc-d9fb-4d4a-b397-aec20171b025" | jq '.fields[] | select(.name | contains("JWT")) | .value' --raw-output) \
   OCM_TOKEN=$(ocm token --refresh) \
-  RHSSO_CLIENT_ID=$(bw get username 028ce1a9-f751-4056-9c72-aea70052728b) RHSSO_CLIENT_SECRET=$(bw get password 028ce1a9-f751-4056-9c72-aea70052728b) \
+  RHSSO_SERVICE_ACCOUNT_CLIENT_ID=$(bw get username 028ce1a9-f751-4056-9c72-aea70052728b) RHSSO_SERVICE_ACCOUNT_CLIENT_SECRET=$(bw get password 028ce1a9-f751-4056-9c72-aea70052728b) \
```
dhaus67, line 26: Might be worthwhile to have a link at the top of this README to `docs/development/howto-run-e2e-tests.md`, this way people know this doc exists in the first place.

dhaus67 (approved):
Just as a note: the manual test step you describe in the PR currently does not apply, right? This PR doesn't have the documentation, so not really sure how the whole setup and configuration was done but I suppose we find out in the other PRs 😜

dev/env/defaults/00-defaults.env
```diff
@@ -59,3 +59,6 @@ export FLEET_MANAGER_RESOURCES_DEFAULT='{"requests":{"cpu":"200m","memory":"300M
 export FLEETSHARD_SYNC_RESOURCES_DEFAULT='{"requests":{"cpu":"200m","memory":"300Mi"},"limits":{"cpu":"200m","memory":"300Mi"}}'
 export DB_RESOURCES_DEFAULT='{"requests":{"cpu":"100m","memory":"300Mi"},"limits":{"cpu":"100m","memory":"300Mi"}}'
 export RHACS_OPERATOR_RESOURCES_DEFAULTS='{"requests":{"cpu":"200m","memory":"300Mi"},"limits":{"cpu":"200m","memory":"300Mi"}}'
+
+export ENABLE_EXTERNAL_CONFIG_DEFAULT=false
+export CHAMBER_DEFAULT=chamber
```
SimonBaeumer, line 64: What does this env variable do?

dev/env/scripts/lib.sh
```diff
@@ -308,3 +313,9 @@ docker_logged_in() {
         return 1
     fi
 }
+
+load_external_config() {
+  local chamber="${CHAMBER:-$CHAMBER_DEFAULT}"
```
SimonBaeumer, line 318: Why not expecting "chamber" in the path directly? I don't understand the env var here

SimonBaeumer (changes requested):
Please add documentation how chamber is used here. If the code is merged would mean that an undocumented change is in main and may break development environments.

Please also document that `chamber` must be installed in the README.md getting started guide.

kovayur:
Documentation added in #485

dev/env/scripts/lib.sh
kovayur: This allows to override the executable.
Might be helpful in case when we want to override executable or add extra parameters, for example:
```
# .envrc
export CHAMBER="aws-vault exec dev -- chamber" 
```
Such style is used pretty often in our scripts, i.e. `$KUBECTL` or  `$DOCKER`.

dev/env/defaults/00-defaults.env
kovayur: This is the default path to an executable.

dev/env/scripts/lib.sh
mtesseract: @SimonBaeumer
This style serves two purposes:
1. As @kovayur already pointed out, the user has full control over overwriting stuff, for example `export DOCKER=$PWD/.rancher-desktop/bin/docker`.
2. On the other hand this allows for smarter defaulting logic, which is not constant, but could be influenced by the cluster type for instance. So, for example, when it is an openshift cluster, then `OPENSHIFT_MARKETPLACE_DEFAULT=true` is automatically set. For example, when the current kubernetes cluster is named "colima", then `DOCKER_DEFAULT="colima nerdctl -- -n k8s.io"`, so that using `$DOCKER ...` you can directly build images for the kubernetes cluster. This is for example picked up in the Makefile.

mtesseract: You can find all sorts of cluster-specific defaulting logic in `dev/env/defaults/cluster-type-*`.

.openshift-ci/tests/e2e.sh
```diff
@@ -61,30 +61,21 @@ if [[ "$RUN_AUTH_E2E" == "true" ]]; then
...
 
-case "$AUTH_TYPE" in
-OCM)
-
-    log "Logging in with client credentials + Refreshing OCM Service Token"
-    ocm login --token "${OCM_OFFLINE_TOKEN}"
-    OCM_TOKEN=$(ocm token --refresh)
-    export OCM_TOKEN
-    ;;
+if [[ -z "$STATIC_TOKEN" ]]; then
```
mtesseract, line 72: The variable `$STATIC_TOKEN` is referenced in a manifest (`manifests/fleet-manager/01-fleet-manager-secrets.yaml`). Therefore I would suggest to move these kind of checks somewhere here https://github.com/stackrox/acs-fleet-manager/blob/main/dev/env/scripts/lib.sh#L91.

Because, it is completely supported here to, for example, use the `up.sh` or even `apply` script directly to apply something in `dev/env/manifests`. When you move the check into `lib.sh` the user using `up.sh` or `apply` would also see this error message, not only the user using `e2e.sh` (or `up.sh`).

mtesseract (changes requested):

.openshift-ci/tests/e2e.sh
kovayur: As stated in the description, `STATIC_TOKEN` now used only in e2e tests. For deployment, the RHSSO auth type is always used.

e2e/README.md
kovayur: Thanks, done in #485

.openshift-ci/tests/e2e.sh
kovayur: With that in mind, I think it's the right place.

dev/env/scripts/lib.sh
SimonBaeumer: Thanks for the clarification @mtesseract and @kovayur!

.openshift-ci/tests/e2e.sh
mtesseract: I see, thanks. I've still been wondering if that should be in `e2e.sh` or in the `Makefile`, in the `test/e2e` target.
I mean, would be good to also see a friendly error message about this var being undefined when using `make test/e2e` or even `ginkgo ... ./e2e/...` directly.
Related: could we define **the** supported/primary way of executing the e2e test suite (or is it just me who is unsure about this, as I see several different ways of executing it)?

mtesseract (approved):

PR #485, opened 2022-10-13
ROX-13018: Improve and unify the documentation on using the development environment

kovayur:
## Checklist (Definition of Done)

- [ ] Unit and integration tests added
- [ ] Documentation added if necessary
- [ ] CI and all relevant tests are passing
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.

## Test manual

TODO: Add manual testing efforts

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup OCM_OFFLINE_TOKEN=<ocm-offline-token> OCM_ENV=development
make verify lint binary test test/integration
```

PR #492, opened 2022-10-14
ROX-13018: Automate Parameter Store toolset installation

kovayur:
## Description
This is a follow-up PR on #485. Our goal is to automate as much deployment/installation steps as possible. 
When `make deploy/dev` is called for the first time, it will install the necessary tools (`chamber`, `aws-vault`) and will prompt for the AWS credentials. For macOS, credentials will be stored in a dedicated Keychain.

## Test manual
1. Clean your local bin path (`${PROJECT_PATH}/bin`)
2. Run `make deploy/dev`
3. Follow the suggested steps
4. Application is successfully deployed

```
# To run tests locally run:
```


kovayur:
* main
  * **PR #477** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/477" target="_blank">Graphite</a> 
    * **PR #485** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/485" target="_blank">Graphite</a> 
      * **PR #492** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/492" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/492?utm_source=stack-comment).

Makefile
```diff
@@ -128,6 +128,12 @@ ifeq (, $(shell which ${LOCAL_BIN_PATH}/go-bindata 2> /dev/null))
 	}
 endif
 
+${LOCAL_BIN_PATH}/chamber:
+	GOBIN=${LOCAL_BIN_PATH} $(GO) install github.com/segmentio/chamber/v2@v2.10.12
```
mtesseract, line 132: If I understand this right then you are planning to make the `dev/env/scripts/*` dependant on `chamber`, is this correct? Maybe it would make sense to add this `LOCAL_BIN_PATH` to `PATH` in `dev/env/scripts/lib.sh::init` so that the scripts these tools in path automatically? (Would allow, e.g. to use these scripts directly, without make)

dev/env/scripts/lib.sh
```diff
@@ -315,7 +316,21 @@ docker_logged_in() {
...
 
 load_external_config() {
-  local chamber="${CHAMBER:-$CHAMBER_DEFAULT}"
+  local chamber
+  local use_aws_vault="${USE_AWS_VAULT:-$USE_AWS_VAULT_DEFAULT}"
+  if [ "$use_aws_vault" = true ]; then
+    local profile="${AWS_VAULT_PROFILE:-$AWS_VAULT_PROFILE_DEFAULT}"
+    local credentials_exist
+    credentials_exist=$(aws-vault list --credentials | grep "${profile}" || true)
+    if [[ ! $credentials_exist ]]; then
```
mtesseract, line 325: why not

```suggestion
    if ! aws-vault list --credentials | grep "${profile}"; then
```

mtesseract: In any case, I guess something like
```
grep "^${profile}$"
```

would be more reliable (depending on the precise output of the tool).

dev/env/scripts/lib.sh
```diff
@@ -315,7 +316,21 @@ docker_logged_in() {
...
+      log "Creating AWS Vault profile '$profile'"
+      aws-vault add "$profile"
+    fi
+    chamber="aws-vault exec ${profile} -- chamber"
```
mtesseract, line 329: nit: I guess, if the user would accidentally store a value containing a whitespace in `AWS_VAULT_PROFILE`, then this would probably fail with some usage error.
```suggestion
    chamber="aws-vault exec '${profile}' -- chamber"
```

mtesseract (changes requested):

dev/env/scripts/lib.sh
kovayur: I've just checked, with single quotes it doesn't work, but works with a whitespace

dev/env/scripts/lib.sh
kovayur: > why not

If I inline, I should suppress the output, with variable it looks cleaner IMHO.

> In any case, I guess something like

Thanks, added

Makefile
kovayur: Reworked. Now the executable should always be in `PATH`. The scripts can now be used directly, without make

.openshift-ci/tests/e2e-test.sh
```diff
@@ -6,14 +6,14 @@ export GITROOT
 source "${GITROOT}/dev/env/scripts/lib.sh"
 init
 
-bootstrap.sh
+make prepare/dev
```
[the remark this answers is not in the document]
kovayur: Reverted

dev/env/defaults/cluster-type-colima/env
```diff
@@ -9,3 +9,6 @@ if grep -q "runtime: docker" <(colima status 2>&1); then
 else
   export DOCKER_DEFAULT="colima nerdctl -- -n k8s.io" # When nerdctl and docker can be used
 fi
+
+export ENABLE_EXTERNAL_CONFIG_DEFAULT="true"
```
mtesseract, line 13: Q: Do I understand this correctly that we want these two settings for basically all local installations but not for remote installations (infra, OpenShift CI)?

Makefile
```diff
@@ -129,6 +130,21 @@ ifeq (, $(shell which ${LOCAL_BIN_PATH}/go-bindata 2> /dev/null))
...
 
+# This goal is intended for installing the tools required for running (and sourcing) the dev scripts (dev/env/scripts)
+# This way scripts may be called outside of makefile
+script-tools: $(CHAMBER_BIN) $(AWS_VAULT_BIN)
+.PHONY: script-tools
+
+CHAMBER_BIN := $(GOBIN)/chamber
+$(CHAMBER_BIN): $(TOOLS_DIR)/go.mod $(TOOLS_DIR)/go.sum
+	@echo "+ $@"
+	@cd $(TOOLS_DIR) && $(GO) install github.com/segmentio/chamber/v2
```
SimonBaeumer, line 141: Why is it necessary to change the directory before installing the tool?

From `go help install` it seems to be installed into `$GOROOT/bin`, `$GOBIN` or `$GOTOOLDIR`:

```
Executables are installed in the directory named by the GOBIN environment
variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH
environment variable is not set. Executables in $GOROOT
are installed in $GOROOT/bin or $GOTOOLDIR instead of $GOBIN.
```

Makefile
```diff
@@ -129,6 +130,21 @@ ifeq (, $(shell which ${LOCAL_BIN_PATH}/go-bindata 2> /dev/null))
 	}
 endif
 
+# This goal is intended for installing the tools required for running (and sourcing) the dev scripts (dev/env/scripts)
+# This way scripts may be called outside of makefile
+script-tools: $(CHAMBER_BIN) $(AWS_VAULT_BIN)
```
SimonBaeumer, line 135:
```suggestion
install-dev-tools: $(CHAMBER_BIN) $(AWS_VAULT_BIN)
```

dev/env/scripts/lib.sh
```diff
@@ -59,6 +59,8 @@ init() {
     GITROOT_DEFAULT=$(git rev-parse --show-toplevel)
     export GITROOT=${GITROOT:-$GITROOT_DEFAULT}
     set -eu -o pipefail
+    # Ensure that the necessary tools are installed
+    make -s -C "$GITROOT" script-tools
```
mtesseract, line 63: nit, just a possible performance improvement:
We could check for something like:
```
if [[ ! ( -x "${GITROOT}/tools/chamber" && -x "${GITROOT}/tools/other-tool" ... ) ]]; then
   make ...
fi
```

dev/env/scripts/lib.sh
```diff
@@ -315,7 +318,21 @@ docker_logged_in() {
 }
 
 load_external_config() {
-  local chamber="${CHAMBER:-$CHAMBER_DEFAULT}"
+  local chamber
+  local use_aws_vault="${USE_AWS_VAULT:-$USE_AWS_VAULT_DEFAULT}"
+  if [ "$use_aws_vault" = true ]; then
```
SimonBaeumer, line 323: nit: consistency. I prefer `[[ ]]` because it has more features for string matching.

```suggestion
  if [[ "$use_aws_vault" = true ]]; then
```

dev/env/scripts/lib.sh
```diff
@@ -315,7 +318,21 @@ docker_logged_in() {
 }
 
 load_external_config() {
-  local chamber="${CHAMBER:-$CHAMBER_DEFAULT}"
+  local chamber
+  local use_aws_vault="${USE_AWS_VAULT:-$USE_AWS_VAULT_DEFAULT}"
```
mtesseract, line 322: I think it would be more in line with the current structure to do
```
export USER_AWS_VAULT=${USE_AWS_VAULT:-$USE_AWS_VAULT_DFAULT}
```
in the init functions where all the other such variables are created and defaulting is applied. And then just refer to `$USE_AWS_VAULT` here instead of introducing a different indirection `$use_aws_vault` -- or am I missing something here?

dev/env/scripts/lib.sh
```diff
@@ -315,7 +318,21 @@ docker_logged_in() {
...
+  if [ "$use_aws_vault" = true ]; then
+    local profile="${AWS_VAULT_PROFILE:-$AWS_VAULT_PROFILE_DEFAULT}"
```
mtesseract, line 324: Same for this.

tools/go.mod
```diff
@@ -0,0 +1,54 @@
+module github.com/stackrox/acs-fleet-manager/tools
```
SimonBaeumer, line 1: Why adding a separate `go.mod` file here?

SimonBaeumer (changes requested):

dev/env/scripts/lib.sh
```diff
@@ -315,7 +318,21 @@ docker_logged_in() {
...
+    local credentials_exist
+    credentials_exist=$(aws-vault list --credentials | grep "^${profile}$" || true)
+    if [[ ! $credentials_exist ]]; then
```
mtesseract, line 327: Why not something like
```suggestion
    if [[ ! aws-vault list --credentials | grep -q "^${profile}$" ]]; then
```
?
Otherwise you are kind of throwing away the exit code (`|| true`) which exactly encodes this control boolean, no?

dev/env/scripts/lib.sh
```diff
@@ -315,7 +318,21 @@ docker_logged_in() {
...
+      log "Creating AWS Vault profile '$profile'"
+      aws-vault add "$profile"
+    fi
+    chamber="aws-vault exec ${profile} -- chamber"
```
mtesseract, line 331:
```suggestion
    chamber="aws-vault exec '${profile}' -- chamber"
```

dev/env/scripts/lib.sh
```diff
@@ -315,7 +318,21 @@ docker_logged_in() {
...
+  else
+    chamber="chamber"
```
mtesseract, line 333: Wouldn't you need to make sure that this `tools/` directory is in `PATH` somewhere?

docs/development/setup-test-environment.md
```diff
@@ -43,30 +41,12 @@ The scripts can be configured using environment variables, the most important op
...
-    Key                             Version         LastModified            User
-    ...
-
-    $ chamber list fleetshard-sync
-    Key                             Version         LastModified            User
-    ...
-    ```
-4. Set up a test cluster using [one of the supported](#Cluster setup) types
-5. Ensure the `kubectl` context is pointing to the desired cluster:
+2. Set up a test cluster using [one of the supported](#Cluster setup) types
```
mtesseract, line 44: nit, in order to keep future PRs small(er), I usually like to use
```
1. ...
1. ...
1. ...
```
instead if enumerating by hand.

tools/go.mod
```diff
@@ -0,0 +1,54 @@
...
+
+go 1.17
```
mtesseract, line 3: Didn't we move to go 1.18?

mtesseract (changes requested):

Makefile
kovayur: This is for using a separate `go.mod` for tools

Makefile
kovayur: Here I wanted to point out that this target is intended for **scripts**. I wouldn't call `chamber` as a dev tool because it should be used in all the environments in the target state.

dev/env/scripts/lib.sh
kovayur: My idea was to keep the tools versions in the `tools/go.mod`. Tools themselves are installed in `$GOPATH/bin`.

`make script-tools` will install them only if:
1. They doesn't exist
2. Version update

In other cases it will skipped: `make: Nothing to be done for 'script-tools'.` (`-s` suppresses this output)
Your suggestion duplicates (1) and breaks (2).

dev/env/scripts/lib.sh
kovayur: `!` changed to `-z`. 
Here I'm checking the command output.
`|| true` is required so that no failure occurs when the pipe command (`grep`) returns a non-zero exit code.

tools/go.mod
kovayur: Not in acs-fleet-manager yet, eager to do it :)

dev/env/scripts/lib.sh
kovayur: Done.

dev/env/scripts/lib.sh
kovayur: Done.

dev/env/scripts/lib.sh
kovayur: Fixed

tools/go.mod
kovayur: I think it makes sense to minimise the number of dependencies in the root go.mod file which become available from the  source code. Same approach exists in stackrox (`tools/linters/go.mod`).
Also, by separating tools, `tools/go.sum` wouldn’t be updated as frequently as root `go.sum`. It means that re-installation of tools will occur less frequently when we call tool installation targets.

dev/env/scripts/lib.sh
mtesseract: ah, could have seen that, thanks.

dev/env/scripts/lib.sh
kovayur: You are right. Fixed. Thanks for the detailed explanation.

mtesseract (approved):

SimonBaeumer (approved):

scripts/external_config.sh
```diff
@@ -15,12 +16,18 @@ init_chamber() {
         export PATH="$GITROOT/bin:$PATH"
     fi
     ensure_tool_installed chamber
+
+    if [[ "$ENABLE_EXTERNAL_CONFIG" != "true" ]]; then
+        return
```
mtesseract, line 21: Not really abou this line.

Am I right in that nowhere in this file is a log emitted informing the user which chamber/aws-vault mode is being active? There are three different scenarios, as I understand it:

* external config is disabled
* external config is enabled and aws vault is enabled
* external config is enabled and aws vault is disabled

WDYT about printing this out somewhere? Could be that I just haven't found it...

mtesseract (approved):
lgtm.
Just have two smaller comments, feel free to discard.

scripts/external_config.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
kovayur: Fixed

scripts/external_config.sh
kovayur: Added diagnostic log in the `up.sh`

scripts/external_config.sh
```diff
@@ -19,8 +20,9 @@ init_chamber() {
         ensure_tool_installed aws-vault
         ensure_aws_profile_exists
     elif [[ -z "${AWS_SESSION_TOKEN:-}" ]] || [[ -z "${AWS_ACCESS_KEY_ID:-}" ]] || [[ -z "${AWS_SECRET_ACCESS_KEY:-}" ]]; then
-        die "Error: Unable to resolve one of the following environment variables: AWS_SESSION_TOKEN, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY.
-            Please set them or use aws-vault by setting USE_AWS_VAULT=true."
+        echo "Error: Unable to resolve one of the following environment variables: AWS_SESSION_TOKEN, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY.
```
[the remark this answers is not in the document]
kovayur: Because it does not source `lib.sh`. I've extracted `log` and `die` to the `scripts/lib/log.sh` for reusing it in `external_config` and therefore, the terraforming script

dev/env/scripts/lib.sh
```diff
@@ -56,8 +46,6 @@ dump_env() {
 }
 
 init() {
-    GITROOT_DEFAULT=$(git rev-parse --show-toplevel)
```
mtesseract, line 59: Why did you remove this?
This is exported, so I don't know from the top of my head if this has unintended side-effects.

dev/env/scripts/lib.sh
```diff
@@ -1,19 +1,9 @@
...
-}
+GITROOT_DEFAULT=$(git rev-parse --show-toplevel)
+export GITROOT=${GITROOT:-$GITROOT_DEFAULT}
 
-log() {
-    # shellcheck disable=SC2059
-    printf "$*"
-    echo
-}
+source "$GITROOT/scripts/lib/log.sh"
```
mtesseract, line 6:
```suggestion
# shellcheck source=/dev/null
source "$GITROOT/scripts/lib/log.sh"
```

dev/env/scripts/up.sh
```diff
@@ -6,7 +6,10 @@ GITROOT="$(git rev-parse --show-toplevel)"
 export GITROOT
 # shellcheck source=/dev/null
 source "${GITROOT}/dev/env/scripts/lib.sh"
+source "${GITROOT}/scripts/lib/external_config.sh"
```
mtesseract, line 9:
```suggestion
# shellcheck source=/dev/null
source "${GITROOT}/scripts/lib/external_config.sh"
```

PR #501, opened 2022-10-17
ROX-13144: Improve e2e diagnostics and ergonomics

mtesseract:
## Description

While debugging an e2e test failure I noticed several shortcomings. This PR does a lot of polishing related to diagnostics and ergonomics, feedback cycle improvements, etc.

https://issues.redhat.com/browse/ROX-13144

- [x] ~~Unit and integration tests added~~ (n/a, the whole PR is about improving the test suite itself)
- [x] ~~Documentation added if necessary~~ (n/a, see above)
- [x] ~~Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.~~ (n/a)

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,8 @@ endif
 endif
 image_tag = $(TAG)
 
+GINKGO_FLAGS ?= -v
+
 # The version needs to be different for each deployment because otherwise the
 # cluster will not pull the new image from the internal registry:
 version:=$(shell date +%s)
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,11 @@ binary:=fleet-manager
 # Otherwise image tag is generated based on git tags.
 ifeq ($(TAG),)
 ifeq (,$(wildcard CI_TAG))
+ifeq ($(IGNORE_REPOSITORY_DIRTINESS),true)
+TAG=$(shell git describe --tags --abbrev=10 --long)
+else
 TAG=$(shell git describe --tags --abbrev=10 --dirty --long)
+endif
 else
 TAG=$(shell cat CI_TAG)
 endif
@@ -353,7 +357,7 @@ test/cluster/cleanup:
 	./scripts/cleanup_test_cluster.sh
 .PHONY: test/cluster/cleanup
 
-test/e2e: gotestsum
+test/e2e:
 	CLUSTER_ID=1234567890abcdef1234567890abcdef \
 	RUN_E2E=true \
 	ENABLE_CENTRAL_EXTERNAL_CERTIFICATE=$(ENABLE_CENTRAL_EXTERNAL_CERTIFICATE) \
@@ -364,8 +368,17 @@ test/e2e: gotestsum
 		--race --trace \
 		--json-report=e2e-report.json \
 		--timeout=$(TEST_TIMEOUT) \
+		--slow-spec-threshold=5m \
 		 ./e2e/...
 .PHONY: test/e2e
+
+test/e2e/reset:
+	@./dev/env/scripts/reset
+.PHONY: test/e2e/reset
+
+test/e2e/cleanup:
+	@./dev/env/scripts/down.sh
+.PHONY: test/e2e/cleanup
 
 # generate files
 generate: moq openapi/generate
--- a/dev/env/defaults/00-defaults.env
+++ b/dev/env/defaults/00-defaults.env
@@ -10,12 +10,13 @@ export CLUSTER_ID_DEFAULT="1234567890abcdef1234567890abcdef"
 export CLUSTER_DNS_DEFAULT="cluster.local"
 
 export IMAGE_REGISTRY_DEFAULT="quay.io/rhacs-eng"
-STACKROX_VERSION_TAG="3.72.0"
+STACKROX_VERSION_TAG="3.72.0" # Note: SCANNER_VERSION_DEFAULT needs to be in sync with this.
 export STACKROX_OPERATOR_VERSION_DEFAULT="${STACKROX_VERSION_TAG}"
 export CENTRAL_VERSION_DEFAULT=$(echo "$STACKROX_VERSION_TAG" | sed -e 's/0-nightly/x-nightly/;')
-export SCANNER_VERSION_DEFAULT="2.25.1" # This one matches the above operator version tag.
+export SCANNER_VERSION_DEFAULT="2.26.0" # This one matches the above operator version tag.
 export STACKROX_OPERATOR_NAMESPACE_DEFAULT="stackrox-operator"
-export FLEET_MANAGER_IMAGE_DEFAULT="fleet-manager:$(make -s -C "$GITROOT" tag)"
+export FLEET_MANAGER_IMAGE_DEFAULT=""
+export IGNORE_REPOSITORY_DIRTINESS_DEFAULT="false"
 export ENABLE_DB_PORT_FORWARDING_DEFAULT="false"
 export ENABLE_FM_PORT_FORWARDING_DEFAULT="false"
 export OPENSHIFT_MARKETPLACE_DEFAULT="false"
--- a/dev/env/scripts/down.sh
+++ b/dev/env/scripts/down.sh
@@ -15,12 +15,4 @@ delete "${MANIFESTS_DIR}/db" || true
 delete "${MANIFESTS_DIR}/fleet-manager" || true
 delete "${MANIFESTS_DIR}/fleetshard-sync" || true
 
-central_namespaces=$($KUBECTL get namespace -o jsonpath='{range .items[?(@.status.phase == "Active")]}{.metadata.name}{"\n"}{end}' | grep '^rhacs-.*$' || true)
-
-for namespace in $central_namespaces; do
-    $KUBECTL delete namespace "$namespace" &
-done
-log "Waiting for leftover RHACS namespaces to be deleted... "
-for p in $(jobs -pr); do
-    wait "$p"
-done
+delete_tenant_namespaces
--- a/dev/env/scripts/lib.sh
+++ b/dev/env/scripts/lib.sh
@@ -153,9 +153,15 @@ init() {
     export SKIP_TESTS=${SKIP_TESTS:-$SKIP_TESTS_DEFAULT}
     export ENABLE_CENTRAL_EXTERNAL_CERTIFICATE=${ENABLE_CENTRAL_EXTERNAL_CERTIFICATE:-$ENABLE_CENTRAL_EXTERNAL_CERTIFICATE_DEFAULT}
     export CENTRAL_DOMAIN_NAME=${CENTRAL_DOMAIN_NAME:-$CENTRAL_DOMAIN_NAME_DEFAULT}
-    export FLEET_MANAGER_IMAGE="${FLEET_MANAGER_IMAGE:-$FLEET_MANAGER_IMAGE_DEFAULT}"
+    export FLEET_MANAGER_IMAGE=${FLEET_MANAGER_IMAGE:-$FLEET_MANAGER_IMAGE_DEFAULT}
+    export IGNORE_REPOSITORY_DIRTINESS=${IGNORE_REPOSITORY_DIRTINESS:-$IGNORE_REPOSITORY_DIRTINESS_DEFAULT}
 
-    if [[ "$ROUTE53_ACCESS_KEY" == "" || "$ROUTE53_SECRET_ACCESS_KEY" == "" ]]; then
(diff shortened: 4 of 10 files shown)


Makefile
```diff
@@ -353,7 +357,7 @@ test/cluster/cleanup:
 	./scripts/cleanup_test_cluster.sh
 .PHONY: test/cluster/cleanup
 
-test/e2e: gotestsum
+test/e2e:
```
mtesseract, line 360: bogus dep removed here.

Makefile
```diff
@@ -13,7 +13,11 @@ binary:=fleet-manager
 # Otherwise image tag is generated based on git tags.
 ifeq ($(TAG),)
 ifeq (,$(wildcard CI_TAG))
+ifeq ($(IGNORE_REPOSITORY_DIRTINESS),true)
```
mtesseract, line 16: This allows to do `IGNORE_REPOSITORY_DIRTINESS=true make tag` and the tag would be constructed without any `-dirty` suffix. This is useful in combination with the deployment scripts when one actively works on the e2e tests but doesn't require new images being build all the time.

Makefile
```diff
@@ -364,9 +368,18 @@ test/e2e: gotestsum
 		--race --trace \
 		--json-report=e2e-report.json \
 		--timeout=$(TEST_TIMEOUT) \
+		--slow-spec-threshold=5m \
 		 ./e2e/...
 .PHONY: test/e2e
 
+test/e2e/reset:
```
mtesseract, line 375: This is useful for quick feedback cycles when developing e2e tests. Instead of using something like `down.sh && up.sh` which always does clean fresh deployments, it is often sufficient to just erase all tenants, from control-plane (DB) and data-plane (Kubernetes resources).  Can be used as in:
```
make test/e2e/reset test/e2e
```

dev/env/defaults/00-defaults.env
```diff
@@ -10,12 +10,13 @@ export CLUSTER_ID_DEFAULT="1234567890abcdef1234567890abcdef"
 export CLUSTER_DNS_DEFAULT="cluster.local"
 
 export IMAGE_REGISTRY_DEFAULT="quay.io/rhacs-eng"
-STACKROX_VERSION_TAG="3.72.0"
+STACKROX_VERSION_TAG="3.72.0" # Note: SCANNER_VERSION_DEFAULT needs to be in sync with this.
 export STACKROX_OPERATOR_VERSION_DEFAULT="${STACKROX_VERSION_TAG}"
 export CENTRAL_VERSION_DEFAULT=$(echo "$STACKROX_VERSION_TAG" | sed -e 's/0-nightly/x-nightly/;')
-export SCANNER_VERSION_DEFAULT="2.25.1" # This one matches the above operator version tag.
+export SCANNER_VERSION_DEFAULT="2.26.0" # This one matches the above operator version tag.
```
mtesseract, line 16: The previous scanner version was not in sync with the above stackrox (operator) version. Consequently the `bootstrap.sh` script did not take care of pulling the correct scanner image and the therefore the scanner deployments failed due to the image not being present in some situations.

dev/env/scripts/down.sh
```diff
@@ -15,12 +15,4 @@ delete "${MANIFESTS_DIR}/db" || true
...
-central_namespaces=$($KUBECTL get namespace -o jsonpath='{range .items[?(@.status.phase == "Active")]}{.metadata.name}{"\n"}{end}' | grep '^rhacs-.*$' || true)
-
-for namespace in $central_namespaces; do
-    $KUBECTL delete namespace "$namespace" &
-done
-log "Waiting for leftover RHACS namespaces to be deleted... "
-for p in $(jobs -pr); do
-    wait "$p"
-done
+delete_tenant_namespaces
```
mtesseract, line 18: I have factored this out into a function in `lib.sh` so that it can be called on the command line directly. As in:
```
bash -c "source dev/env/scripts/lib.sh; init; delete_tenant_namespaces"
```
I needed something like this from time to time and was to lazy to always write long shell loops on the command line.

dev/env/scripts/lib.sh
```diff
@@ -315,7 +321,31 @@ docker_logged_in() {
 }
 
 load_external_config() {
-  local chamber="${CHAMBER:-$CHAMBER_DEFAULT}"
-  eval "$($chamber env fleet-manager)"
-  eval "$($chamber env fleetshard-sync)"
+    local chamber="${CHAMBER:-$CHAMBER_DEFAULT}"
```
mtesseract, line 324: reindented

dev/env/scripts/lib.sh
```diff
@@ -315,7 +321,31 @@ docker_logged_in() {
...
+    eval "$($chamber env fleetshard-sync)"
+}
+
+delete_tenant_namespaces() {
+    central_namespaces=$($KUBECTL get namespace -o jsonpath='{range .items[?(@.status.phase == "Active")]}{.metadata.name}{"\n"}{end}' | grep '^rhacs-.*$' || true)
+    if [[ ! "$central_namespaces" ]]; then
+        log "No left-over RHACS tenant namespaces to be deleted."
+        return
+    fi
+    for namespace in $central_namespaces; do
```
mtesseract, line 335: delete central CRs first to make sure that we don't end up in a race where a running operator initiates a reconcile while we are trying to empty the namesapce.

e2e/e2e_auth_test.go
```diff
@@ -21,9 +21,8 @@ const (
 var _ = Describe("AuthN/Z Fleet* components", func() {
 
 	BeforeEach(func() {
-		if env := getEnvDefault("RUN_AUTH_E2E", "false"); env == "false" {
-			Skip("The RUN_AUTH_E2E variable was not set, skipping the tests. If you want to run the auth tests, " +
-				"set RUN_AUTH_E2E=true")
+		if !runningAuthTests {
+			Skip("Skipping auth test")
```
mtesseract, line 25: I have tried to make the e2e output a bit more concise. This long messages doesn't need to be printed again and again for every test spec. Instead it is now printed once at the beginning of the test suite.

e2e/e2e_suite_test.go
```diff
@@ -18,15 +18,18 @@ import (
...
+	routeService         *k8s.RouteService
+	dnsEnabled           bool
+	routesEnabled        bool
+	route53Client        *route53.Route53
+	waitTimeout          = getWaitTimeout()
+	dpCloudProvider      = getEnvDefault("DP_CLOUD_PROVIDER", "standalone")
+	dpRegion             = getEnvDefault("DP_REGION", "standalone")
+	authType             = "OCM"
+	fleetManagerEndpoint = "http://localhost:8000"
+	runningAuthTests     bool
```
mtesseract, line 32: Added these because I moved some of the related code into the suite initialization.

e2e/e2e_suite_test.go
```diff
@@ -80,6 +83,23 @@ var _ = BeforeSuite(func() {
...
+		fleetManagerEndpoint = val
+	}
+	GinkgoWriter.Printf("FLEET_MANAGER_ENDPOINT: %q\n", fleetManagerEndpoint)
+
+	if val := getEnvDefault("RUN_AUTH_E2E", "false"); val == "true" {
+		runningAuthTests = true
+		GinkgoWriter.Printf("Executing auth tests")
+	} else {
+		GinkgoWriter.Printf("Skipping auth tests. If you want to run the auth tests, set RUN_AUTH_E2E=true")
+	}
```
mtesseract, line 102: This is now part of suite initialization (`BeforeSuite`) instead of spec initialization (`BeforeSpec`).

e2e/e2e_test.go
```diff
@@ -51,13 +51,9 @@ const (
 var _ = Describe("Central", func() {
 	var client *fleetmanager.Client
 	var adminAPI *private.DefaultApiService
-	BeforeEach(func() {
```
mtesseract, line 54: has been moved to suite initialization.

e2e/e2e_test.go
```diff
@@ -69,28 +65,28 @@ var _ = Describe("Central", func() {
...
 		adminClient, err := fleetmanager.NewClient(fleetManagerEndpoint, adminAuth)
 		adminAPI = adminClient.AdminAPI()
-
 		Expect(err).ToNot(HaveOccurred())
-
 	})
 
 	Describe("should be created and deployed to k8s", func() {
 		var err error
-
```
mtesseract, line 79: Moved down because I added a `GinkgoWriter.Printf` call, which seems only to be allowed within test leaf nodes (e.g. `It`) as opposed to container nodes.

dev/env/defaults/00-defaults.env
```diff
@@ -10,12 +10,13 @@ export CLUSTER_ID_DEFAULT="1234567890abcdef1234567890abcdef"
...
 export STACKROX_OPERATOR_NAMESPACE_DEFAULT="stackrox-operator"
-export FLEET_MANAGER_IMAGE_DEFAULT="fleet-manager:$(make -s -C "$GITROOT" tag)"
```
mtesseract, line 18: We now set this in `init()`, because it depends on the defaulting logic for `IGNORE_REPOSITORY_DIRTINESS`.

dev/env/scripts/lib.sh
```diff
@@ -153,9 +153,15 @@ init() {
...
+    export FLEET_MANAGER_IMAGE=${FLEET_MANAGER_IMAGE:-$FLEET_MANAGER_IMAGE_DEFAULT}
+    export IGNORE_REPOSITORY_DIRTINESS=${IGNORE_REPOSITORY_DIRTINESS:-$IGNORE_REPOSITORY_DIRTINESS_DEFAULT}
 
-    if [[ "$ROUTE53_ACCESS_KEY" == "" || "$ROUTE53_SECRET_ACCESS_KEY" == "" ]]; then
+    if [[ "$FLEET_MANAGER_IMAGE" == "" ]]; then
+        tag=$(make -s -C "$GITROOT" tag)
+        FLEET_MANAGER_IMAGE="fleet-manager:${tag}"
+    fi
+
+    if [[ "$ENABLE_CENTRAL_EXTERNAL_CERTIFICATE" != "false" && ("$ROUTE53_ACCESS_KEY" == "" || "$ROUTE53_SECRET_ACCESS_KEY" == "") ]]; then
```
mtesseract, line 164: This reduces noise

e2e/e2e_test.go
```diff
@@ -69,54 +65,71 @@ var _ = Describe("Central", func() {
...
+				fmt.Sprintf("Central ID: %s", createdCentral.Id),
+			}
+			printNotes(notes)
 			namespaceName, err = services.FormatNamespace(createdCentral.Id)
 			Expect(err).To(BeNil())
 			Expect(constants.CentralRequestStatusAccepted.String()).To(Equal(createdCentral.Status))
 		})
 
 		It("should transition central's state to provisioning", func() {
+			if createdCentral == nil {
```
mtesseract, line 101: Fail fast, when the central could not be created.

dev/env/scripts/lib.sh
kovayur: Do we have any editor config for shell scripts which I need to import?

dev/env/scripts/reset
```diff
@@ -0,0 +1,19 @@
...
+# shellcheck source=/dev/null
+source "${GITROOT}/dev/env/scripts/lib.sh"
+init
+
+log "Resetting state of control-plane and data-plane..."
+
+port-forwarding start fleet-manager 8000 8000
+port-forwarding start db 5432 5432
+sleep 1
+psql -h localhost -U fleet_manager -d rhacsms -c 'DELETE FROM central_requests;'
```
kovayur, line 14: This introduces the new tool, isn't it?
Wouldn't it be better if we call a command inside docker using the postgres image instead?

kovayur (changes requested):
Looks good already, just a small remark about using psql

dev/env/scripts/reset
mtesseract: It does, yes. I am a bit concerned about startup times when introducing a dockerized postgresq here just for executing a query. How about using the local psql of it exists and spawning a docker container otherwise?

mtesseract: Regarding the editor config, I fear we don't have it right now, but we should integrate https://editorconfig.org/ for this, I guess.

kovayur: > How about using the local psql of it exists and spawning a docker container otherwise
I'm fine with this, thanks!

kovayur (approved):

PR #505, opened 2022-10-18
ROX-12689: oc login automatically, drop need for ocm login

porridge:
## Description

This approach relieves us from having to automate `ocm` login (unclear if this is even possible at all), at the cost of maintaining some (non-secret) per-cluster information in this script.

- [x] Documentation added if necessary (i.e. changes to dev setup, test execution, ...) :point_right: will update the [environments confluence page](https://docs.engineering.redhat.com/pages/viewpage.action?pageId=288990393) before merging once this is approved, to mention the need to update this script after creating the cluster.

## Test manual

Tested by prepending the `helm` command with an `echo`, running the script, and checking for the expected post-login output from `oc` was there.

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
+++ b/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
@@ -10,9 +10,6 @@ if [[ $# -ne 2 ]]; then
     echo "Usage: $0 [environment] [cluster]" >&2
     echo "Known environments: stage prod"
     echo "Cluster typically looks like: acs-{environment}-dp-01"
-    echo ""
-    echo "Note: you need to be logged into OCM for your environment's administrator"
-    echo "Note: you need to be logged into OC for your cluster's administrator"
     exit 2
 fi
 
@@ -29,9 +26,6 @@ load_external_config observability OBSERVABILITY_
 
 case $ENVIRONMENT in
   stage)
-    # TODO: Fetch OCM token and log in as appropriate user as part of script.
-    EXPECT_OCM_ID="2ECw6PIE06TzjScQXe6QxMMt3Sa"
-
     FM_ENDPOINT="https://xtr6hh3mg6zc80v.api.stage.openshift.com"
 
     FLEETSHARD_SYNC_IMAGE="quay.io/app-sre/acs-fleet-manager:f6ad53e"
@@ -41,9 +35,6 @@ case $ENVIRONMENT in
     ;;
 
   prod)
-    # TODO: Fetch OCM token and log in as appropriate user as part of script.
-    EXPECT_OCM_ID="2BBslbGSQs5PS2HCfJKqOPcCN4r"
-
     FM_ENDPOINT="https://api.openshift.com"
 
     FLEETSHARD_SYNC_IMAGE="quay.io/app-sre/acs-fleet-manager:6da74ac"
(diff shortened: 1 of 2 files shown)


dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -99,20 +97,38 @@ case $ENVIRONMENT in
...
+  CLUSTER_URL_INFIX="pnz3.p1"
+  OC_TOKEN_BITWARDEN_ID="..."
+  ;;
+*)
+  echo "Unknown cluster ${CLUSTER_NAME}" >&2
+  exit 2
+  ;;
+esac
+
+OC_TOKEN=$(bw get password "${OC_TOKEN_BITWARDEN_ID}")
```
ebensh, line 127: How often will the token expire and need to be updated in BitWarden?

ebensh: More generally, depending on BitWarden at all means the process will be manual, not automatic - there's no service account credentials for BitWarden.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -99,20 +97,38 @@ case $ENVIRONMENT in
...
+  if [[ $EXPECTED_ENVIRONMENT != $ENVIRONMENT ]]; then
+    echo "Cluster ${CLUSTER_NAME} is expected to be in environment ${EXPECTED_ENVIRONMENT}, not ${ENVIRONMENT}" >2
+    exit 2
+  fi
+}
+
+case $CLUSTER_NAME in
+acs-stage-dp-01)
+  assert_environment stage
+  CLUSTER_ID="d3ccd7a0-9b33-41e3-9024-6677f42259ed"
```
ebensh, line 111: Please add a comment here on how the ID is fetched (one liner using ocm would be nice) so we can update them when adding or changing the data plane

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -99,20 +97,38 @@ case $ENVIRONMENT in
...
+  CLUSTER_URL_INFIX="blbk.p1"
```
ebensh, line 112: Same here

ebensh (approved):

porridge:
FYI @kovayur since this is somewhat related to secret management in the sense that it will require addition of another (per-cluster) secret.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: This is only using bitwarden for consistency with all the other things that are fetched from there. When we migrate to another secret management solution this will obviously move as well.

porridge: Regarding the expiry, when writing this I was under the impression that credentials in one's `kubeconfig` generally don't require a refresh, so I thought the TTL was pretty high. But now that you point this out I realize that it's slightly different, it's a token *which is used to obtain* credentials for a kubeconfig. And the token could have a (potentially much lower TTL). I'll investigate closer.

kovayur: As an alternative we can also take a look at the command:
`ocm cluster login $CLUSTER_ID`

mtesseract: If I recall correctly the expiry time of these tokens was so short (2 days or something like that?) that I stopped persisting these tokens on my system.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: Sadly the cluster ID emitted by `ocm cluster list` is something different than the ID shown in the UI. :facepalm:

porridge:
PTAL @kovayur @ebensh

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: FTR, `ocm cluster login` [does not](https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/cluster/login/login.go) actually help w.r.t. authentication. [Even](https://github.com/openshift-online/ocm-cli/issues/412) with `--token`.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -58,20 +49,41 @@ case $ENVIRONMENT in
...
+}
+
+# The following values can be retrieved from the Red Hat Hybrid Cloud Console.
+# - Cluster ID the first piece of information on the "Details" pane of the Overview tab of the given cluster.
+# - The URL infix is the part of the Control Plane API endpoint between cluster name and "openshiftapps.com",
+#   in the "Cluster ingress" pane of the Networking tab for the given cluster.
+case $CLUSTER_NAME in
+acs-stage-dp-01)
+  assert_environment stage
+  CLUSTER_ID="d3ccd7a0-9b33-41e3-9024-6677f42259ed"
```
kovayur, line 67: WDYT about moving these variables (`CLUSTER_ID` and `CLUSTER_URL_INFIX `) to the Parameter Store?
No need to do it now, just interested in your opinion on the external configuration.

kovayur (approved):

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: I think this is a good idea:
- this script cannot do much without parameter store being up anyway
- with some luck, this will help us avoid a manual step of modifying this script at the time of cluster creation in the  future

PR #511, opened 2022-10-21
Support deploying Operator from Quay on remote OpenShift Cluster

mtesseract:
## Description

Until now it was only possible to deploy non-marketplace (i.e. Quay) operator on local clusters (e.g. Colima).
With this change it is possible to also deploy such non-release operators to remote OpenShift clusters.

In case we would need to use a non-release operator in the future it is good to have this capability also during testing.

- [ ] ~~Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~~ (n/a, since this fixes an undocumented limitation).

## Test manual

1. Spin up remote OpenShift cluster
2. Make sure `QUAY_USER` and `QUAY_TOKEN` is set in your environment.
3. Make sure that you have your `AUTH_TYPE` set up correctly (note: there is currently a bug regarding `STATIC_TOKEN`...).
4. Use for example `export FLEET_MANAGER_IMAGE=quay.io/mclasmei/fleet-manager:v0.0.0-357-gcb53739d8a` (when deploying to remote cluster we need to use fleet-manager images pullable from Quay.)
5. Exec `OPENSHIFT_MARKETPLACE=false OPERATOR_SOURCE=quay dev/env/scripts/bootstrap.sh`
6. Verify that the operator is running in the namespace `stackrox-operator`.


.openshift-ci/tests/e2e.sh
```diff
@@ -24,8 +24,6 @@ if [[ "${OPENSHIFT_CI:-}" == "true" ]]; then
     done
     export STATIC_TOKEN="${FLEET_STATIC_TOKEN:-}"
     export STATIC_TOKEN_ADMIN="${FLEET_STATIC_TOKEN_ADMIN:-}"
-    export QUAY_USER="${IMAGE_PUSH_USERNAME:-}"
```
mtesseract, line 27: This was just plain wrong. `QUAY_USER` and `QUAY_TOKEN` are expected to contain pull secrets, not push secrets.

If this functionality shall be enabled we would need to add proper pull secrets as `QUAY_USER` & `QUAY_TOKEN in the OpenShift CI Vault.

But since we don't actually need to deploy operators from Quay right now in OpenShift CI, I have not added such credentials in the Vault.

dev/env/scripts/bootstrap.sh
```diff
@@ -92,24 +92,32 @@ if [[ "$INSTALL_OPERATOR" == "true" ]]; then
             # operatorhubio catalog is ready, which is why an additional delay has been added.
             echo "Waiting for CatalogSource to include rhacs-operator..."
             while true; do
-                $KUBECTL -n "$STACKROX_OPERATOR_NAMESPACE" get packagemanifests.packages.operators.coreos.com -o json |
-                    jq -r '.items[].metadata.name' | grep -q '^rhacs-operator$' && break
+                if $KUBECTL -n "$STACKROX_OPERATOR_NAMESPACE" get packagemanifests.packages.operators.coreos.com -o json |
```
mtesseract, line 95: The `if ... then` logic protects agains terminating here early.

The `jq` part has been changed to check for the correct catalog.

dev/env/scripts/bootstrap.sh
```diff
@@ -92,24 +92,32 @@ if [[ "$INSTALL_OPERATOR" == "true" ]]; then
...
-            #   Message:               all available catalogsources are healthy
-            #   Reason:                AllCatalogSourcesHealthy
-            #   Status:                False
-            #   Type:                  CatalogSourcesUnhealthy
-            #   Message:               error using catalog operatorhubio-catalog (in namespace olm): failed to list bundles: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 10.43.96.123:50051: i/o timeout"
-            #   Status:                True
-            #   Type:                  ResolutionFailed
-            #
-            # Therefore we wait for the operatorhubio-catalog/registry-server container to become ready.
-            wait_for_container_to_become_ready "olm" "olm.catalogSource=operatorhubio-catalog" "registry-server"
```
mtesseract, line 112: This shall only be done when deploying OLM. But on OpenShift we don't deploy OLM.

dev/env/defaults/00-defaults.env
```diff
@@ -48,7 +48,7 @@ export OBSERVABILITY_CONFIG_ACCESS_TOKEN_DEFAULT=""
 export IMAGE_PULL_DOCKER_CONFIG_DEFAULT=""
 export SPAWN_LOGGER_DEFAULT="false"
 export DUMP_LOGS_DEFAULT="false"
-export OPERATOR_SOURCE_DEFAULT="quay"
+export OPERATOR_SOURCE_DEFAULT=""
```
mtesseract, line 51: It doesn't make much sense here to specify a default, in my opinion.
In fact, this defaulting here was problematic and the cause for a CI failure (because it caused defaulting logic in a follow-up stage to deduce `INHERIT_IMAGE_PULL_SECRETS=true`, even though the cluster type was `openshift-ci`).

Instead this is now simply initialized as empty. And only later it is is set to either `quay` or `marketplace` depending on the `is_openshift_cluster` predicate.

kovayur (approved):

PR #514, opened 2022-10-24
ROX-12076: Add `make code/check` to pre-commit hook

stehessel:
This also executes the formatter during CI.

These are the formatting changes extracted from https://github.com/stackrox/acs-fleet-manager/pull/408, with the addition that the formatter is now executed as part of the pre-commit hook and CI.

## Test manual

```
# To run tests locally run:
```


misberner (approved):

janisz (approved):

stehessel:
Sorry for re-ping, but I noticed that `gofmt` behaves differently than I expected. In particular, it does not fail if it formats a file. So the CI didn't work as intended (formatting was done, but didn't fail the pipeline). Could you take another look?

Makefile
```diff
@@ -432,6 +433,14 @@ openapi/generate/rhsso: go-bindata openapi-generator
 	$(GOFMT) -w pkg/client/redhatsso/api
 .PHONY: openapi/generate/rhsso
 
+# fail if formatting is required
+code/check:
+	@if ! [ -z $$($(GOFMT) -l `find . -type f -name '*.go' -not -path "./vendor/*"`) ]; then \
```
misberner, line 438: I guess there is some value in remaining consistent with `code/fix` below, but generally speaking, this would better be rewritten as:
```suggestion
	@if ! [ -z "$$(find . -path './vendor' -prune -o -type f -name '*.go' -print0 | xargs -0 $(GOFMT) -l)" ]; then \
```
1. `""` prevents issues with spaces in expansion. `[ .. ]` does not do anything special with `$(...)`. The following, for example, will return a zero exit status: `[ -z $(echo foo -o -n bar) ]`, because the expansion string is tokenized due to the missing quotes (if you use `[[ ... ]]`, which is a bash builtin, bash does recognize that `$(...)` should be treated as a single argument to `-z`).
2. `find . -path './vendor' -prune -o ...` makes sure find does not traverse the `vendor` directory, if any (your version does post-filtering, which requires first finding all `.go` files within vendor as well)
3. `find ... -print0 | xargs -0` prevents any tokenization issues in case file paths contain spaces.

misberner (approved):
I think this is fine for consistency, but please see the recommendations on better bash scripting.

On a more general note: the approach we take in stackrox is that CI does formatting (`make code/fix`), and it that is not a no-op (i.e., modifies the working tree), CI fails. The nice thing about this is that the same pattern can be used for formatting, checking that generated files are up to date, etc. I don't know if this repo _has_ a CI stage for ensuring generated files are up to date, but if it does, I would recommend to implement the fix there. Reason being that `code/fix` also runs `go mod tidy`, but you don't seem to be checking that the `go.mod` file actually is tidy, so there's some unexpected discrepancy.

The downside of this approach of course is that you cannot implement this pattern as part of a precommit, I think

stehessel:
There is a pre-commit hook that checks for generated files, see https://github.com/stackrox/acs-fleet-manager/blob/main/.pre-commit-config.yaml#L4.

Anyway, I think I'll leave the formatting as it is for now and just include the improved shell scripting.

PR #517, opened 2022-10-25
ROX-12151: cloud accounts endpoint

ivan-degtiarenko:
## Description
Introduces the call to AMS client as well as new endpoint. The organization ID is taken from the caller's claims.

## Test manual

1. `./fleet-manager serve --ocm-base-url=https://api.openshift.com --ocm-client-id-file=secrets/ocm-service.clientId --ocm-client-secret-file=secrets/ocm-service.clientSecret `
2. `GET http://localhost:8000/api/rhacs/v1/cloud_accounts` with `rhacs-test-customer` Bearer token
3. Returned result:
```
{
    "cloudAccounts": [
        {
            "cloudAccountId": "<redacted>",
            "cloudProviderId": "aws"
        }
    ]
}
```


pkg/client/ocm/client.go
```diff
@@ -600,3 +602,26 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
...
+	}
+
+	quotaCostClient := organizationClient.Organization(organizationID).QuotaCost()
+
+	quotaCostList, err := quotaCostClient.List().Parameter("fetchCloudAccounts", true).Send()
+	if err != nil {
+		return nil, fmt.Errorf("error getting cloud accounts: %w", err)
+	}
+
+	quotaCostList.Items().Each(func(qc *amsv1.QuotaCost) bool {
```
stehessel, line 621: This will list all cloud accounts, right? We should filter here on the cloud accounts that are attached to our marketplace SKU, i.e. `"quota_id": "cluster|rhinfra|rhacs|marketplace"`. Otherwise we'll also return cloud accounts that have purchased different products.

pkg/client/ocm/client.go
```diff
@@ -600,3 +602,26 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
 
 	return res, nil
 }
+
+func (c *client) GetCustomerCloudAccounts(externalID string) ([]*amsv1.CloudAccount, error) {
```
stehessel, line 606: Can you add a unit test for this function using the generated mocks?

internal/dinosaur/pkg/api/public/api/openapi.yaml
```diff
@@ -608,6 +608,37 @@ paths:
       security:
       - Bearer: []
       summary: Returns the list of supported regions of the supported cloud provider
+  /api/rhacs/v1/cloud_accounts:
+    get:
+      operationId: getCloudAccounts
```
sachaudh, line 613: The endpoint looks good to me. The only thing I'm wondering about is this `operationId`. @stehessel mentioned in his comment how to access the `organization.id` using `ocm` command. Is `operationId` supposed to be that? Also, accessing that id using the ocm command is helpful, but I can't do that through the browser so I'll need to find another way to do that. If you or @stehessel might know how that's possible, let me know. I am searching through https://api.openshift.com/ to see if there might be some API endpoint that might give me that information

sachaudh: Is `/api/accounts_mgmt/v1/organizations` the API i'm looking for?

ivan-degtiarenko: @sachaudh `organization.id` will be extracted by `fleet-manager` from AMS using `orgId` provided in the Bearer token.  The endpoint doesn't take any parameters.

`operationId` is swagger parameter that is used to uniquely identify the operation within the API spec:
https://swagger.io/docs/specification/paths-and-operations/

pkg/client/ocm/client.go
ivan-degtiarenko: Do you mean `cloud_accounts_test.go` for the endpoint handler?  There's no way to mock `sdkClient.Connection` used in `client.go`

stehessel: Hm yes. I think you could write a unit test with the `cloud_accounts_test.mock`, and you could also take a look at https://github.com/stackrox/acs-fleet-manager/blob/main/internal/dinosaur/pkg/services/quota/ams_quota_service_test.go for how to mock the OCM client. With the latter you could test the SKU filter logic.

internal/dinosaur/pkg/api/public/api/openapi.yaml
sachaudh: Ok, then I think that's fine with me. Thanks for adding the endpoint @ivan-degtiarenko 👍🏼

pkg/client/ocm/client.go
```diff
@@ -600,3 +602,31 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
 
 	return res, nil
 }
+
+func (c *client) GetCustomerCloudAccounts(externalID string, quotaIDs []string) ([]*amsv1.CloudAccount, error) {
+	var res []*amsv1.CloudAccount
+	organizationClient := c.connection.AccountsMgmt().V1().Organizations()
```
stehessel, line 608: There is already `GetQuotaCostsForProduct` - seems like it's what we want, it's only missing a `fetchCloudAccounts=true` flag. Maybe we could reuse this here?

internal/dinosaur/pkg/handlers/cloud_accounts_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/stackrox/acs-fleet-manager/pkg/client/ocm"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	JwtKeyFile = "test/support/jwt_private_key.pem"
+	JwtCAFile  = "test/support/jwt_ca.pem"
+)
+
+func TestGetSuccess(t *testing.T) {
```
stehessel, line 23: It would be good to test `GetCustomerCloudAccounts` as well, in particular testing that the filtering by resource id works. I think this is possible by mocking the other OCM client methods like `QuotaCost` (or `GetQuotaCostsForProduct`).

internal/dinosaur/pkg/handlers/cloud_accounts.go
```diff
@@ -0,0 +1,48 @@
...
+	return &cloudAccountsHandler{
+		client: client,
+	}
+}
+
+// Get ...
+func (h *cloudAccountsHandler) Get(w http.ResponseWriter, r *http.Request) {
+	cfg := &handlers.HandlerConfig{
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+			ctx := r.Context()
```
stehessel, line 30: If you extract inside of this function (`func() (i interface{}, serviceError *errors.ServiceError) { ... }`, you could test it separately from the http related parts of `Get`. This would save you some http setup in the `cloud_accounts_test.go`.

pkg/client/ocm/client.go
ivan-degtiarenko: Not really - `GetQuotaCostsForProduct` does filtration by `resourceName` and `product`

internal/dinosaur/pkg/handlers/cloud_accounts_test.go
ivan-degtiarenko: `GetQuotaCostsForProduct` can't be reused for the reasons mentioned above. I'm not sure how one can mock `QuotaCost` method on the organizationClient - it is created by `Connection`:
https://github.com/stackrox/acs-fleet-manager/blob/ivan%2FROX-12151-aws-accounts-endpoint/pkg/client/ocm/client.go#L608-L608

pkg/client/ocm/client.go
stehessel: In principle that is what we also want. But I think you are right, we would want to filter for `billing_model==marketplace` as well. Easier to go the `quota_id` route then.

internal/dinosaur/pkg/handlers/cloud_accounts_test.go
stehessel: Hm yes that is impractical, I thought I saw it in the method list of the client, but you are right it's not. That would leave us with mocks on http level, which maybe are supported via the "mock modes" (https://github.com/stackrox/acs-fleet-manager/blob/ivan/ROX-12151-aws-accounts-endpoint/pkg/client/ocm/config.go#L12). But probably a lot of effort for a simple feature...

stehessel (approved):

PR #518, opened 2022-10-25
ROX-12076: Probe service for blackbox monitoring [1/3]

stehessel:
This is the first batch of changes for the probe service. It encompasses the Makefile, Dockerfile and the underlying cli and config setup.

See https://github.com/stackrox/acs-fleet-manager/pull/408 for the preceding PR. I'm planning (at least) two follow up PRs for the runtime logic and the metrics logic.

## Test manual

```
# To run tests locally run:
```


probe/cmd/cmd.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+
+			for {
+				if err := executeProbe(config); err != nil {
+					if errors.Is(err, errInterruptSignal) {
+						return err
+					}
+					glog.Errorf("%+v", err)
+				}
+				time.Sleep(config.RuntimeRunWaitPeriod)
```
stehessel, line 54: @misberner I didn't understand your point of avoiding `time.Sleep`. Can you comment if this is appropriate here? The purpose is to wait some time in between probe runs.

probe/config/config_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSingleton_Success(t *testing.T) {
+	t.Setenv("FLEET_MANAGER_ENDPOINT", "http://127.0.0.1:8888")
```
SimonBaeumer, line 12: Please use the envisolator to avoid side-effects in unit tests.
The acs-fleet-manager repo can import the stackrox env isolator: https://github.com/stackrox/acs-fleet-manager/blob/4ce487d2ad2ce453580d0fb6e71a743f257bec94/fleetshard/pkg/central/reconciler/proxy_test.go#L16-L16

probe/cmd/cmd.go
SimonBaeumer: `time.Sleep` can't be cancelled because it does not return a channel which you can select on.

As the probe polls like the fleetshard-sync why not using the `concurrency.RetryTicker` which takes care of cancellation, retries etc?

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,35 @@
...
+
+// New creates a new runtime.
+func New(config *config.Config) (*Runtime, error) {
+	return &Runtime{
+		config: config,
+	}, nil
+}
+
+// RunSingle executes a single probe run.
+func (r *Runtime) RunSingle(ctx context.Context) error {
```
SimonBaeumer, line 24: Imho the endless loop created should be wrapped in the runtime, the runtime calls a package to execute the probe.
The `probe` package could implement a function `Execute` which executes a single probe run. The probe does not need to know whether it is called from a user once, in a for-loop or from a REST/gRPC endpoint.

I think you implemented it nearly and only small changes necessary:

- Rename `runtime` to `probe` (or something which fits)
- Move `for {}` endless loop from `cmd/` pkg to a runtime package
- Move cancellation to composition root in cmd package (or something else which is "above" a layer above runtime)

probe/cmd/cmd.go
```diff
@@ -0,0 +1,112 @@
...
+	case <-sigs:
+		return errInterruptSignal
+	case err := <-runResult:
+		if err != nil {
+			return errors.Wrap(err, "probe run failed")
+		}
+		return nil
+	case <-ctx.Done():
+		return errors.Wrap(ctx.Err(), "probe run failed")
+	}
```
SimonBaeumer, line 111: I think this should called from outside the for loop. In this implementation every probe call setups up the listeners again which makes it inefficient.
This implementation needs refactors if multiple go routines are started and need cancellation.

SimonBaeumer (changes requested):
Added some comments, the overall approach to me looks good.

Summarized:
- Cancellation is created for each probe run, this is not necessary
- Packages need a bit restructuring
-  Replacing the loop with `concurrency.NewRetryTicker`

probe/main.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	glog.Infof("probe service has been started")
+	defer func() {
+		glog.Info("probe service has been stopped")
+	}()
+
+	cmd := cmd.Command()
+	if err := cmd.Execute(); err != nil {
+		panic(err)
```
dhaus67, line 28: We should call `glog.Fatalf` here instead of relying on panic.

probe/cmd/cmd.go
dhaus67: Or, if the `concurrency.RetryTicker` doesn't fit your use-case, you can simply use `time.Ticker` with the specified timeout (we have references on how to use it within the stackrox repo, e.g. [a watcher for k8s config map changes](https://github.com/stackrox/stackrox/blob/d653406651df4331a714839ec2c0a23a93425c64/pkg/k8scfgwatch/poll_watcher.go#L33-L47)).

probe/config/config_test.go
stehessel: `t.Setenv` already restores the environment after the test, so I think this is unnecessary. Or does the env isolator have another purpose?

probe/cmd/cmd.go
stehessel: Went with the `RetryTicker`.

probe/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```sh
+./probe/bin/probe run
+```
+
+or run a endless loop of probes
+
+```sh
+./probe/bin/probe start
+```
```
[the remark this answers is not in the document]
stehessel: done

probe/pkg/runtime/runtime.go
stehessel: Tried to follow the suggestion. Can you take another look?

probe/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
+
+COPY bin/probe /stackrox/
+
+EXPOSE 7070
+
+CMD ["/stackrox/probe", "start"]
```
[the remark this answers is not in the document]
stehessel: done

probe/cmd/cmd.go
```diff
@@ -0,0 +1,112 @@
...
+	return c
+}
+
+func startCommand() *cobra.Command {
+	c := &cobra.Command{
+		SilenceUsage: true,
+		Use:          "start",
+		Short:        "Start a continuous loop of probe runs.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
```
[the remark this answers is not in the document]
stehessel: done

probe/main.go
stehessel: done

probe/internal/cli/cli.go
```diff
@@ -0,0 +1,97 @@
...
+			runtime, err := runtime.New()
+			if err != nil {
+				return errors.Wrap(err, "failed to create runtime")
+			}
+			defer runtime.Stop()
+
+			go func() {
+				if err := runtime.Start(); err != nil {
+					glog.Fatal(err)
+				}
```
dhaus67, line 47: This doesn't currently work with the way the function is currently implemented.

Right now, we call `runtime.Start` which in turn will start a `concurrency.RetryTicker`. We will return any error within start. Since that will not produce an error, we will never actually receive an error here.

Now, if within the retry ticker, any error occurs, the retry ticker will be stopped (which can be retrieved from running `ticker.Stopped()`.
However, this will not be propagated to this go routine, and we do not have a channel which we use to indicate any issues that _may_ have occurred during execution.

For me, the best would be the following:
- Share the code for handling context cancellation, errors, and interrupt signal from both `run` / `start`.
- Improve the existing go routine inline you created for the `run` part, specifically:
```go
go func() {
  select {
    case runResult <- runtime.RunSingle(ctxTimeout):
    case ctxTimeout.Done():
  }
}()
```
This way we ensure the go routine won't be blocking when the context timeout was hit before an error got returned from `RunSingle`.

I'm not sure whether it's required to prepare the current code to allow the execution of multiple go routines instead of the single `probe.Execute()` (e.g. splitting up the probes into multiple sub-routines?), but this is probably something we can iterate upon once we desire it.

dhaus67 (changes requested):

probe/config/config_test.go
dhaus67: I don't think we need to re-use `envisolator` anymore. This was something we had prior go1.17 (where [t.Setenv](https://pkg.go.dev/testing#T.Setenv) only has been added) and mimics the behavior of `t.Setenv` (with the caveat that `t.Setenv` only works for non-parallel tests compared to `envisolator`).

misberner: Yeah, `envisolator` shouldn't be used any more. And `envisolator` also does not work for parallel tests, at least it doesn't do the right thing.

probe/cmd/probe/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// Always log to stderr by default, required for glog.
+	if err := flag.Set("logtostderr", "true"); err != nil {
+		glog.Info("Unable to set logtostderr to true.")
+	}
+
+	glog.Infof("probe service has been started")
+	defer func() {
+		glog.Info("probe service has been stopped")
+	}()
```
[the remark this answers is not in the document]
stehessel: Indeed this looks a bit weird, I can remove them.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,61 @@
...
+	return &Runtime{
+		Config: config,
+	}, nil
+}
+
+// Start a continuous loop of probe runs.
+func (r *Runtime) Start() error {
+	ticker := concurrency.NewRetryTicker(func(ctx context.Context) (timeToNextTick time.Duration, err error) {
+		if err := r.RunSingle(ctx); err != nil {
+			return 0, errors.Wrap(err, "failed to execute single probe run")
```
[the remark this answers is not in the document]
stehessel: > As exit condition, the ticker currently will only exit [if the error is deemed as non-recoverable error.](https://github.com/stackrox/stackrox/blob/74c227af5798d3dbc93ec7b2cefe847fe266b855/pkg/concurrency/retry_ticker.go#L101-L104).

But isn't this exactly what we want? I.e. if the probe fails with a non-recoverable error then fail the service, otherwise just register a failed run in the metrics and continue with the next run?

probe/internal/cli/cli.go
stehessel: > Share the code for handling context cancellation, errors, and interrupt signal from both run / start.

Can you be more explicit? I'm not sure how to share the code because we want to do different things:
* In run we always want to fail on every error.
* In start we want to continue the loop on most errors.

probe/config/config_test.go
SimonBaeumer: Did not know that, thanks for the input!

probe/internal/cli/cli.go
```diff
@@ -0,0 +1,97 @@
...
+			}()
+
+			sigs := make(chan os.Signal, 1)
+			signal.Notify(sigs, os.Interrupt)
+			defer signal.Stop(sigs)
+
+			select {
+			case <-sigs:
+				return errInterruptSignal
+			case err := <-runResult:
```
SimonBaeumer, line 89: Is `err == nil` the success case in a single probe run? If yes, I think the success case is not handled here? 🤔

probe/pkg/runtime/runtime.go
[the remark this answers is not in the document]
SimonBaeumer: +1, the probe should continue running after an error in the probe occurred. 
@stehessel would a failed probe run trigger an alert (in a later implementation)?

probe/pkg/runtime/runtime.go
SimonBaeumer: Yes, looks very good, nice!

One small nit-pick: The `RunSingle` is not necessary as it wraps only the `probe.Execute`. The `probe run` can call `probe.Execute` directly.
This way the runtime's responsibility is to guarantee the continuous execution of probe runs, the probe pkg is responsible for executing the probe.

SimonBaeumer (approved):

probe/internal/cli/cli.go
stehessel: If `err == nil`, then `errors.Wrap(err, ...)` is also `nil`.

probe/pkg/runtime/runtime.go
stehessel: After the next step (adding the actual probe logic) it will be necessary, because `runtime` will also set up the fleet manager client etc.

probe/pkg/runtime/runtime.go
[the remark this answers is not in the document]
stehessel: Yeah you are right, I think we need an error channel for `start` as well, but it should only be filled with non recoverable errors.

probe/pkg/runtime/runtime.go
misberner: I think there may have been some conflicting advice from both Simon and me in the respective review round, so before iterating further, it might be good to agree on an approach first. We need:
1. a way to run a single probe execution
2. a way to run probe executions in a loop
3. a way to (cleanly) terminate an ongoing (single) probe execution, e.g., for `^C` in oneshot mode
4. a way to terminate the loop execution

I think it's a no-brainer that 2 should be based upon 1. I think I initially suggested to pull 2 out of `runtime`, but if there's a separate `probe` component, I think it makes sense to have the loop be part of runtime. In general I agree with Simon that the portion in `cmd` should contain as little logic as possible.

Furthermore, I think 3 and 4 should be the same thing from a caller's perspective. When a stop is requested, the single execution can return a special kind of error, and the loop just terminates. Or, the stop facility triggers a signal/cancels a context that is both used to abort a single execution, but also checked before starting a new loop iteration.

The bottom-line is, I think that there shouldn't be a `Stop()` method that only works in conjunction with `Start()`, because the use case of stopping execution is not tied to the loop mode only. For this reason, I think that one-shot and loop execution should have a uniform signature and both be synchronous, rather than differing in whether or not they take a context and whether or not they are asynchronous.

probe/internal/cli/cli.go
dhaus67: I think my main point is that I don't expect `RunSingle` or `Start` to have different function signatures.

Additionally, there are still conditions where we want to error out, e.g. as you mentioned above when the ticker could not be started.

So, for me it would be best to share this between both commands (creating the runtime, starting the routine that will _either_ do the one-shot or start the loop, react on signals and act upon them (e.g. stop the runtime, exit the progam). I don't think we should treat those differently conceptually, so I do think its possible to share the code for both.

probe/internal/cli/cli.go
```diff
@@ -0,0 +1,97 @@
...
+	"os/signal"
+
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/stackrox/acs-fleet-manager/probe/pkg/runtime"
+)
+
+// errInterruptSignal corresponds to a received SIGINT signal.
+var errInterruptSignal error = errors.New("received interrupt signal")
```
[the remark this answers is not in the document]
stehessel: Hmm my IDE uses `gofumpt`, which is a bit stricter and removes the variable block as opposed to `gofmt`. I'd prefer use `gofumpt` formatting, but I can disable it if there are already established conventions like this one.

probe/internal/cli/cli.go
```diff
@@ -0,0 +1,97 @@
...
+		Short:        "Run a single probe run.",
+		Args:         cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			runtime, err := runtime.New()
+			if err != nil {
+				return errors.Wrap(err, "failed to create runtime")
+			}
+			defer runtime.Stop()
+
+			ctxTimeout, cancel := context.WithTimeout(context.Background(), runtime.Config.RuntimeRunTimeout)
```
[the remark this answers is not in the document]
stehessel: The probe timeout in the loop case currently comes from the `RetryTicker`, which calls its function with a timeout context, which is then also passed to `RunSingle`.

probe/internal/cli/cli.go
```diff
@@ -0,0 +1,97 @@
...
+
+			go func() {
+				if err := runtime.Start(); err != nil {
+					glog.Fatal(err)
+				}
+			}()
+
+			sigs := make(chan os.Signal, 1)
+			signal.Notify(sigs, os.Interrupt)
+			defer signal.Stop(sigs)
```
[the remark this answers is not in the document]
stehessel: You are right, I think this can be removed.

stehessel:
> I looked at the docs, and if I am not mistaken, the "retry" in the name refers to the fact that a back-off strategy is applied on failures. I.e., you can have different waiting intervals after successful and failed executions, and also have the interval after a failed execution change dynamically based on how many executions have failed in a row. Nifty, but I'm not sure if that is what Stephan wants/needs?

The backoff is definitely not needed here.

probe/internal/cli/cli.go
SimonBaeumer: Uh, did not know that. Today I learned. 🧑‍🏫

probe/pkg/runtime/runtime.go
stehessel: Alright, let me summarize:
* We need the 4 types of execution/termination you mentioned. Just to add some more example cases:
  - a single probe run is cleanly terminated upon any error, like a context timeout (`Config.RuntimeRunTimeout`) or on `^C`.
  - in one-shot mode, the program will exit with non-zero exit code upon any error.
  - in loop mode, a recoverable error like context timeout during the probe run does not cancel the loop. It just cleans up the current run and waits for the next one.
  - in loop mode, a non-recoverable error like `^C` cleans up, cancels the loop and the program will exit with non-zero exit code.

Implementation:
* `Start` (loop) and `RunSingle` (one-shot) are part of the runtime package.
* `Start` and `RunSingle` have the same signature, take a context and run in the main go routine. The context is passed further down to the probe package, which will do the requests to fleet manager.
* `Start` calls `RunSingle` internally.
* The interrupt and error handling is done in a separate function, which is run in a separate go routine.

Did I miss anything or misunderstand the proposal?

probe/pkg/probe/probe.go
```diff
@@ -0,0 +1,17 @@
...
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// Execute the probe of the fleet manager API.
+func Execute(ctx context.Context) error {
+	// Dummy run
+	glog.Info("probe run has been started")
+	time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
stehessel: Thanks, I see now. I thought: But it does exit on timeout?!

probe/config/config_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSingleton_Success(t *testing.T) {
```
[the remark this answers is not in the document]
stehessel: Tbh I copied this part from the fleetshard config code. Agreed the name doesn't make sense here, I'll change it.

stehessel:
I refactored the code such that:

* `runtime.Start` and `runtime.RunSingle` have identical signatures and are now synchronously run on the main go routine.
* The interrupt handling has been factored out into a separate function.
* The `RetryTicket` has been removed in favor of regular `time.Ticker`.
* Smaller changes based on the nit comments.

probe/cmd/probe/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+func main() {
+	// This is needed to make `glog` believe that the flags have already been parsed, otherwise
+	// every log messages is prefixed by an error message stating the the flags haven't been
+	// parsed.
+	_ = flag.CommandLine.Parse([]string{})
+
+	// Always log to stderr by default, required for glog.
+	if err := flag.Set("logtostderr", "true"); err != nil {
+		glog.Info("Unable to set logtostderr to true.")
```
misberner, line 18: just curious - where would we log this to, given the failure reason?

probe/pkg/probe/probe.go
```diff
@@ -0,0 +1,24 @@
...
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/concurrency"
+)
+
+var (
+	// ErrNotRecoverable should lead to a graceful service shutdown.
+	ErrNotRecoverable = errors.New("not recoverable error")
```
misberner, line 14: Hmm, this is a bit of an antipattern as this doesn't say _what_ the error actually is. An error should always be or contain a cause. I saw that you are planning to wrap it (using `errors.Is` for checking), but the _fundamental_ error should always be an error condition, not a label we attach to some other error condition.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,79 @@
...
+
+// RunSingle executes a single probe run.
+func (r *Runtime) RunSingle(ctx context.Context) error {
+	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+	defer cancel()
+	defer r.CleanUp()
+
+	err := probe.Execute(ctxTimeout)
+	if ctxErr := ctxTimeout.Err(); ctxErr != nil {
+		ctxErr = errors.Wrap(ctxErr, "probe context cancelled")
```
misberner, line 65: Also nitpicking here, but given that you didn't call `cancel()`, it is more likely the context timed out/exceeded its deadline rather than being cancelled (this is of course possible if the parent context was canceled).

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, errors.Wrap(err, "failed to load configuration")
+	}
+
+	return &Runtime{
+		Config: config,
+	}, nil
+}
+
+// Start a continuous loop of probe runs.
+func (r *Runtime) Start(ctx context.Context) error {
```
misberner, line 40: I would call this `RunLoop` or `RunForever` to emphasize the analogy to `RunSingle`

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// RunSingle executes a single probe run.
+func (r *Runtime) RunSingle(ctx context.Context) error {
+	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+	defer cancel()
+	defer r.CleanUp()
+
+	err := probe.Execute(ctxTimeout)
+	if ctxErr := ctxTimeout.Err(); ctxErr != nil {
```
misberner, line 64: I would avoid this check:
- if the probe finished successfully (`err == nil`) just before the context expired, there's no reason to return an error.
- if the probe hit an error _other_ than context expiration, this error is most likely to be more relevant for understanding the failure condition; i.e., if `err != nil`, `err` should not be masked.
- if the probe hits a timeout/context cancellation, `Execute` should return `ctxTimeout.Err()` anyway.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,79 @@
...
+		return ctxErr
+	}
+	if err != nil {
+		return errors.Wrap(err, "probe run failed")
+	}
+	return nil
+}
+
+// CleanUp remaining probe resources.
+func (r *Runtime) CleanUp() error {
```
misberner, line 76: The unfortunate reality is that `CleanUp` will also perform blocking operations, and thus should also receive `context.Context` parameter. This also means you need a second "cleanup" timeout. It also means that when the context is cancelled, you won't be able to exit right away (that's why I initially said I would prefer an asynchronous call to `Stop()` being made rather than a context being cancelled, as the former does not suggest responsiveness - but no need to change it now).

You can add the context param when you start doing something here, but just as a heads-up.

misberner, line 76: Separate question: is `Runtime` the right place for `CleanUp`? Since `probe` does the setup, shouldn't it also do the cleanup?

probe/cmd/probe/main.go
stehessel: glog defaults to logging to temporary files. See https://pkg.go.dev/github.com/golang/glog#section-documentation.

probe/pkg/runtime/runtime.go
stehessel: I was planning that `runtime.CleanUp` calls not yet existing functions in `probe` to do the actual clean up work (just passing the fleet manager client and so on). If `runtime.CleanUp` turns out unnecessary I can remove it in favor of `probe.CleanUp`.

probe/pkg/probe/probe.go
stehessel: Tbh I don't know yet what unrecoverable error we even expect from the probe run (other than `^C`, which is handled by the parent context). So probably it makes sense to removes this now and revisit with a specific error at hand, which we want to cancel the loop.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+	ticker := time.NewTicker(r.Config.RuntimeRunWaitPeriod)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Wrap(ctx.Err(), "probe context invalid")
+		case <-ticker.C:
+			err := r.RunSingle(ctx)
+			glog.Warning(err)
```
dhaus67, line 50:
```suggestion
			if err := r.RunSingle(ctx); err != nil {
			        glog.Warning(err)
			}
```

probe/internal/cli/cli.go
```diff
@@ -0,0 +1,86 @@
...
+	runtime *runtime.Runtime
+}
+
+// New creates a CLI.
+func New() (*CLI, error) {
+	runtime, err := runtime.New()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create runtime")
+	}
+	return &CLI{runtime}, nil
```
dhaus67, line 30: nit (to keep consistency with runtime.New()):
```suggestion
	return &CLI{runtime: runtime}, nil
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var backoff = wait.Backoff{
+	Duration: 1 * time.Second,
+	Factor:   1.5,
+	Jitter:   0.1,
+	Steps:    15,
+	Cap:      10 * time.Minute,
+}
```
dhaus67, line 20:
```suggestion
```

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+			glog.Warning(err)
+		}
+	}
+}
+
+// RunSingle executes a single probe run.
+func (r *Runtime) RunSingle(ctx context.Context) error {
+	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+	defer cancel()
+	defer r.CleanUp()
```
dhaus67, line 59: I saw @misberner's comment suggesting to change this later on - specifically making the call to `r.CleanUp()` async and expose e.g. a concurrency.Signal indicating whether cleanup has been finalized. So I can assume it's a TODO for later on?

Irrespective of the changes to be made, it's probably still worthwhile to use `defer utils.Should(r.CleanUp())` instead of ignoring the error from `r.CleanUp()` here.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+
+// RunLoop a continuous loop of probe runs.
+func (r *Runtime) RunLoop(ctx context.Context) error {
+	ticker := time.NewTicker(r.Config.RuntimeRunWaitPeriod)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Wrap(ctx.Err(), "probe context invalid")
```
dhaus67, line 47: I think this qualifies as a nit, albeit it would be quite nice to have proper error and log messages on using the interrupt:

Right now, if you would cancel the loop, you would receive the following logs and error message:
```log
^CE1027 03:23:49.503328    3184 cli.go:81] received SIGINT signal, shutting down ...
Error: probe context invalid: context canceled
F1027 03:23:49.503399    3184 main.go:27] probe context invalid: context canceled
```

I thought we initially had plans to use a specific error for this, which is currently unused: `probe/internal/cli.errInterruptSignal`.

Additionally, things get even more confusing when you interrupt the loop when you are currently _within the probe run_:
```log
^CE1027 03:28:16.125293    3330 cli.go:81] received SIGINT signal, shutting down ...
I1027 03:28:16.125347    3330 probe.go:19] probe run has ended
I1027 03:28:16.125360    3330 runtime.go:69] probe resources have been cleaned up
W1027 03:28:16.125368    3330 runtime.go:50] probe run failed: probe run expired: context canceled
Error: probe context invalid: context canceled
```
This will now yield a `probe run expired` error. Reason for this is due to the way `concurrency.WaitWithTimeout` is handling the `ctx.Done` signal. It will return `true`, resulting in us assuming the context deadline was exceeded. Instead, we should check if `errors.Is(ctx.Err(), context.Cancelled)` and handle this to return our "special" interrupt error.

Shall we make it more clear and re-use`probe/internal/cli.errInterruptSignal` _iff_ the context was cancelled? If not, we should remove it since its unused anyways.

probe/pkg/runtime/runtime.go
stehessel: Yeah I'll do this in the next iteration, want to get this one done to avoid moving target.

probe/pkg/runtime/runtime.go
stehessel: done

probe/internal/cli/cli.go
stehessel: done

probe/pkg/runtime/runtime.go
stehessel: I added the check to the interrupt handler in `runtime` since this is where we cancel the context.

probe/pkg/runtime/runtime.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+		}
+	}
+}
+
+// RunSingle executes a single probe run.
+func (r *Runtime) RunSingle(ctx context.Context) error {
+	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+	defer cancel()
+	defer utils.Should(r.CleanUp())
```
misberner, line 52: `utils.Should` should only be used when you think that an `error` can only occur because of a programming error. I think cleaning up can fail for a variety of reasons (ACSMS down?) that are not programming errors.
```suggestion
	defer r.CleanUp()
```

misberner (approved):
I think especially the cleanup portion needs some more thinking, but that will probably be easier once it has a real implementation behind this, so no objections against merging this as-is for now.

dhaus67 (approved):

PR #519, opened 2022-10-25
Add Kubernetes Label to Tenant Namespaces

mtesseract:
## Description

In multiple places we have now logic for retrieving the set of all tenant namespaces by doing some grep-work on `kubectl get ns` output of some sort.

Would be nicer and more robust to have a dedicated label for this.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -385,7 +385,7 @@ func (r *CentralReconciler) getNamespace(name string) (*corev1.Namespace, error)
 
 func (r *CentralReconciler) createTenantNamespace(ctx context.Context, namespace *corev1.Namespace) error {
 	namespace.Labels = make(map[string]string)
-	namespace.Labels["rhacscs/type"] = "tenant"
+	namespace.Labels["rhacs.redhat.com/tenant"] = ""
 	err := r.client.Create(ctx, namespace)
 	if err != nil {
 		return fmt.Errorf("creating namespace %q: %w", namespace.ObjectMeta.Name, err)
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -383,15 +383,21 @@ func (r *CentralReconciler) getNamespace(name string) (*corev1.Namespace, error)
 	return namespace, nil
 }
 
+func (r *CentralReconciler) createTenantNamespace(ctx context.Context, namespace *corev1.Namespace) error {
+	namespace.Labels = make(map[string]string)
+	namespace.Labels["rhacscs/type"] = "tenant"
+	err := r.client.Create(ctx, namespace)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -383,15 +383,21 @@ func (r *CentralReconciler) getNamespace(name string) (*corev1.Namespace, error)
 	return namespace, nil
 }
 
+func (r *CentralReconciler) createTenantNamespace(ctx context.Context, namespace *corev1.Namespace) error {
```
SimonBaeumer, line 386: I think this does not update existing namespace right?

SimonBaeumer (changes requested):
Very good idea @mtesseract!

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -383,15 +383,21 @@ func (r *CentralReconciler) getNamespace(name string) (*corev1.Namespace, error)
...
+	namespace.Labels = make(map[string]string)
+	namespace.Labels["rhacscs/type"] = "tenant"
```
SimonBaeumer, line 388:
```suggestion
	namespace.Labels["rhacs.redhat.com/tenant"] = ""
```

fleetshard/pkg/central/reconciler/reconciler.go
mtesseract: that's correct. I thought it would be ok to not implement update logic to keep it simple and instead do a one-time adjustment to the existing namespaces manually.

fleetshard/pkg/central/reconciler/reconciler.go
mtesseract: Changed.

SimonBaeumer (approved):

fleetshard/pkg/central/reconciler/reconciler.go
mtesseract: @johannes94 I think you mean "If this is rolled out before LA [...]", right?

Do we agree that his change is ok to merge before LA and that it is acceptable to deal with updating tenant namespaces manually until then?

@SimonBaeumer I don't quite follow. What is your point exactly with the bash scripts?

johannes94: Yes I mean LA 👍 

> I am wondering if that is a use-case - applying a bash script to adjust existing instances instead of implementing an update logic in the operator which is only executed once.

I'd prefer a declarative approach, e.g fleetshard-sync has a state it wants the namespaces/resources in general to be, including labels and annotations, and makes sure it gets all resources to that state.

SimonBaeumer: @johannes94 agree, declarative is better. It's the nature of operators to do it programmatically which makes it more complex, the bash script would be a duck-tape solution.

@mtesseract +1 on merging + manual adjustments of labels.

PR #529, opened 2022-10-27
ROX-13157: First Iteration on integrating structured logging

mtesseract:
## Description

See https://issues.redhat.com/browse/ROX-13157.

This is the initial work integrating structured logging into ACSCS components.

**Goals:** Replace currently used logging framework with the `pkg/logging` (built on zap) package from the `stackrox` repo while trying to keep the amount of manual changes small by doing most code transformations automatically using regex substitutions.

**Non-Goals:** Actually leveraging structured logging through the Fleet* components in an idiomatic way (passing around loggers and enriching them with key/value pairs, integrating it deeply into the HTTP handler framework, respecting flow IDs, etc.). This is all to be done iteratively in the future instead of increasing the size of this PR even further.

**Limitations:**
* Currently in most situations the default `root logger` is used. In some situations module-specific loggers are used (support for this is coming straight from the `pkg/logging` package).

**Design Decisions:**
* I have decided to remove the `.V(n)`-style logging coming from the `glog` package. I believe this is adding unnecessary complexity for a use-case where standard debugging levels (`DEBUG`, `WARN`, `INFO`, etc.) should be sufficient.
* As a (dirty) trick to prevent the PR from growing further, I decided to make the necessary adjustments in `pkg/logging` to be able to mimic the currently used `glog` API. Hence I was able to get away in many situations by importing the `pkg/logging` package aliases as `glog`. When iterating on this, this aliasing should be removed.
* The commit history clearly points out those commits which were done using regex substitutions.

In a nut-shell: The Fleet* components now respect the `LOGENCODING` variable which supports the settings `console` or `json`. By default the dev environment uses `LOGENCODING=console`, but it can easily be overwritten by exporting `LOGENCODING=json`.

## Test manual

```
# To run tests locally run:
```


stehessel:
Merging the PR in its current state is too much work. We'll use the PR as a base once we come back to the structured logging topic.

PR #532, opened 2022-10-30
ROX-12076: Probe service - add clients [2]

stehessel:
## Description

Follow up to PR #518. This adds the fleet manager client and http client including mocks for tests.

## Test manual

```
# To run tests locally run:
```


stehessel:
* main
  * **PR #532** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/532" target="_blank">Graphite</a>  👈
    * **PR #533** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/533" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/532?utm_source=stack-comment).

probe/pkg/fleetmanager/client.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"net/http"
+
+	"github.com/pkg/errors"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/public"
+	"github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager"
+	"github.com/stackrox/acs-fleet-manager/probe/config"
+)
+
+// Client is an interface for the fleetmanager client.
```
SimonBaeumer, line 13:
```suggestion
// Client is a wrapper interface for the fleetmanager client.
```

probe/pkg/fleetmanager/client.go
```diff
@@ -0,0 +1,41 @@
...
+//
+//go:generate moq -out client_moq.go . Client
+type Client interface {
```
SimonBaeumer, line 16: +1, this is a very good idea to have a mock for the client!

Could you add an interface check to have compile time checks for the interface implementation? 
See: https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

I think this package can be moved to `pkg/` in general where we maintain a shared client between all components.
Something like `pkg/fleetmanager/client`

probe/pkg/fleetmanager/client.go
```diff
@@ -0,0 +1,41 @@
...
+		ClientID:     config.RHSSOClientID,
+		ClientSecret: config.RHSSOClientSecret, // pragma: allowlist secret
+		Realm:        config.RHSSORealm,
+		Endpoint:     config.RHSSOEndpoint,
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create fleet manager authentication")
+	}
+
+	client, err := fleetmanager.NewClient(config.FleetManagerEndpoint, auth, fleetmanager.WithUserAgent("probe-service"))
```
SimonBaeumer, line 35:
```suggestion
	client, err := fleetmanager.NewClient(config.FleetManagerEndpoint, auth, fleetmanager.WithUserAgent("fleet-manager-probe-service"))
```

SimonBaeumer (approved):

probe/pkg/fleetmanager/client.go
stehessel: done

probe/pkg/fleetmanager/client.go
stehessel: done

dhaus67 (approved):
Just minor nit's, otherwise LGTM 🚀

probe/pkg/fleetmanager/client.go
```diff
@@ -0,0 +1,27 @@
+package fleetmanager
```
SimonBaeumer, line 1: I think you can either omit the client part from `pkg/client/fleetmanager` or reversing it to `.../fleetmanager/client`.
This is because this makes the code explorable, a package "client" is very generic and without context (fleet-manager, fleetshard) not a useful information.

misberner: +1 on that `pkg/fleetmanager/client` would be a less surprising way to structure the packages

probe/pkg/httpclient/client.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"net/http"
+
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/stackrox/acs-fleet-manager/probe/config"
+)
+
+// New creates a http.Client with pre-configured retry and timeout.
+func New(config *config.Config) *http.Client {
+	retryClient := retryablehttp.NewClient()
```
misberner, line 12: How do you know that the default retry settings from a third-party library make sense for this use case?

probe/pkg/httpclient/client_mock.go
```diff
@@ -0,0 +1,18 @@
...
+// RoundTripFunc declares a type for the round trip function.
+type RoundTripFunc func(req *http.Request) *http.Response
+
+// RoundTrip mocks the round trip of the http client.
+func (f RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req), nil
+}
+
+// NewMockClient returns *http.Client with a mocked Transport.
+func NewMockClient(fn RoundTripFunc) *http.Client {
```
misberner, line 14: I would generally avoid placing test-only functions in regular packages. Maybe put this in `probe/pkg/httpclienttest` or `probe/pkg/httpclient/testutils` or so.

misberner (changes requested):

probe/pkg/fleetmanager/client.go
stehessel: I followed the existing structure with different clients placed in `pkg/client/*`. Are you suggesting to refactor the whole structure in https://github.com/stackrox/acs-fleet-manager/tree/main/pkg/client to `pkg/*/client` (for fleetmanager and other clients)? That seems way out of scope to me for this PR. If you mean just the new fleet manager interface in `pkg/fleetmanager/client`, that I can do.

With `.../fleetmanager/client` do you mean reversing it to `probe/pkg/fleetmanager/client`? That I could also do.

probe/pkg/httpclient/client.go
stehessel: I took a quick peak at the implementation in my IDE and it seemed reasonable to me. I can pin the retry config if you prefer.

probe/pkg/httpclient/client_mock.go
```diff
@@ -0,0 +1,18 @@
...
+	return &http.Client{
+		Transport: RoundTripFunc(fn),
+	}
+}
```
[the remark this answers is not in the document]
stehessel: Removed

probe/pkg/httpclient/client.go
misberner: I think the question was indirectly targeted at something else: why are those settings good -> what qualifies "good" settings -> what problems do "good" settings solve, therefore: what made you believe that we need intra-request-level retry? Not having looked at the library, I'm assuming now (for good reasons) that this only applies to `GET` (and maybe `HEAD`, but irrelevant) requests, and only to 5xx and transport-level failures. I didn't check the PR 3 line-by-line, but I would assume that only polling/wait-for-ready/wait-for-$status requests are GET request. However, these are retried anyway (as multiple requests) by the probe code. If we think that intermittent failures are a real risk, not retrying intra-request does not seem to hurt much, and retrying intra-request cannot help with the type of requests (mutating ones) where a single failure means the entire probe run fails anyway.

I guess you could argue that in case of a transient failure condition, the intra-request retry would happen quicker than just waiting for the next poll. But there's little reason to assume that this quicker retry is necessary - the status changing to the value we are waiting for is not any more likely after a transport/5xx failure than it is after a 2xx response with an undesired value, and yet we don't retry quickly after the latter.

Long story short: no objection to keeping it. Maybe we should even use it everywhere. But it does introduce complexity and makes things a tad bit less predictable.

pkg/client/fleetmanager/client.go
```diff
@@ -10,6 +11,19 @@ import (
...
+	CreateCentral(ctx context.Context, async bool, request public.CentralRequestPayload) (public.CentralRequest, *http.Response, error)
+	DeleteCentralById(ctx context.Context, id string, async bool) (*http.Response, error)
+	GetCentralById(ctx context.Context, id string) (public.CentralRequest, *http.Response, error)
+	GetCentrals(ctx context.Context, localVarOptionals *public.GetCentralsOpts) (public.CentralRequestList, *http.Response, error)
+}
+
+// Compile-time check that the interface is implemented by the fleet-manager OpenAPI client.
+var _ PublicClient = (*public.DefaultApiService)(nil)
+
 var _ http.RoundTripper = (*authTransport)(nil)
```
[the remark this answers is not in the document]
misberner: +1 to omitting the comment

probe/pkg/fleetmanager/client.go
misberner: IMHO no need to refactor `pkg/client`. I'm a bit confused if the structure has changed now or if I just misread simon's PR, but the current structure seems fine to me.

misberner (approved):

probe/pkg/httpclient/client.go
stehessel: Alright fair enough. Let me remove the library and add explicit retries as you suggest in PR [3].

stehessel:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/532?utm_source=stack-comment).


Job Reference: WSxW0jmiiYoVqUGHV0ml

PR #533, opened 2022-10-30
ROX-12076: Probe service - probe requests [3]

stehessel:
## Description

Follow up to PR #518. This adds the primary probe package.

Notable changes:
 * Hopefully better use of tickers and contexts.
 * The clean up routine has been moved to the probe package.
 * The clean up is now done in a separate go routine, while the main routine is blocked by a `concurrency.Signal`.
 * Probes are now named via config parameters. The `ProbeName` can be injected with the K8s pod name on deployment.
 * Instead of passing the state of the current probe run for central deletion back the clean up routine, the probe
   requests the list of currently owned centrals from fleet manager. It then iterates over this list and deletes
   the centrals owned by this instance of the probe. The clean up loop is currently executed sequentially, but
   could be optimized by running it concurrently in the future.
 * Added tests to cli and runtime package to test interrupt and timeout handling.

## Test manual

Tested the probe against a dev data plane cluster.

```
# To run tests locally run:
```


stehessel:
* main
  * **PR #532** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/532" target="_blank">Graphite</a> 
    * **PR #533** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/533" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/533?utm_source=stack-comment).

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,208 @@ package probe
...
+			return nil, errors.Wrap(ctx.Err(), "ensure central state timed out")
+		case <-ticker.C:
+			currentCentral, _, err := p.fleetManagerClient.GetCentralById(ctx, central.Id)
+			if err != nil {
+				glog.Warningf("central instance %s not reachable: %s", central.Id, err.Error())
+				continue
+			}
+
+			if currentCentral.Status == targetState {
+				glog.Infof("central instance %s is in `%s` state", currentCentral.Id, targetState)
```
dhaus67, line 167:
```suggestion
				glog.Infof("central instance %s is in %q state", currentCentral.Id, targetState)
```

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,208 @@ package probe
...
+		central := centralList.Items[i]
+		if central.Owner == fmt.Sprintf("service-account-%s", p.config.RHSSOClientID) &&
+			strings.HasPrefix(central.Name, fmt.Sprintf("%s-%s", p.config.ProbeNamePrefix, p.config.ProbeName)) {
+
+			if err := p.deleteCentral(ctx, &central); err != nil {
+				glog.Errorf("failed to clean up central instance %s: %s", central.Id, err.Error())
+			}
+		}
+	}
+	glog.Info("finished clean up of probe resources")
```
dhaus67, line 89: nit:
```suggestion
	glog.Info("finished attempting clean up of probe resources")
```

probe/pkg/probe/probe_test.go
```diff
@@ -0,0 +1,402 @@
...
+		t.Run(tc.testName, func(t *testing.T) {
+			probe, err := New(testConfig, tc.mockFMClient, nil)
+			require.NoError(t, err, "failed to create probe")
+			ctx, cancel := context.WithTimeout(context.TODO(), testConfig.ProbeRunTimeout)
+			defer cancel()
+
+			central, err := probe.createCentral(ctx)
+
+			if tc.wantErr {
+				assert.ErrorContains(t, err, tc.errContains, "expected an error during probe run")
```
dhaus67, line 114: I'm not a fan of these string checks within errors. Ideally, we would do an error comparison here (e.g. `assert.ErrorIs()`.

Unless its really critical, I would favour dropping the specific error check via string and instead simply check for an error being returned, in the lack of absence of concrete error types.

probe/pkg/probe/probe_test.go
```diff
@@ -0,0 +1,402 @@
...
+			} else {
+				assert.Equal(t, constants.CentralRequestStatusReady.String(), central.Status, "central not ready")
+				assert.NoError(t, err, "failed to create central")
```
dhaus67, line 117: You probably want those to be switched and a `required`. If you were to receive an unexpected error in `probe.CreateCentral`, you would get `central == nil` which would lead to a panic when accessing `central.Status`.
```suggestion
				require.NoError(t, err, "failed to create central")
				assert.Equal(t, constants.CentralRequestStatusReady.String(), central.Status, "central not ready")

```

probe/pkg/runtime/runtime.go
```diff
@@ -46,18 +51,22 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
 
 // RunSingle executes a single probe run.
 func (r *Runtime) RunSingle(ctx context.Context) error {
-	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.ProbeRunTimeout)
 	defer cancel()
-	defer r.CleanUp()
+	defer func() {
+		ctxCleanup, cancel := context.WithTimeout(context.Background(), r.Config.ProbeCleanUpTimeout)
```
dhaus67, line 57: nit:
```suggestion
		cleanupCtx, cancel := context.WithTimeout(context.Background(), r.Config.ProbeCleanUpTimeout)
```

probe/pkg/runtime/runtime.go
```diff
@@ -46,18 +51,22 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
 
 // RunSingle executes a single probe run.
 func (r *Runtime) RunSingle(ctx context.Context) error {
-	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.ProbeRunTimeout)
```
dhaus67, line 54:
```suggestion
	probeRunCtx, cancel := context.WithTimeout(ctx, r.Config.ProbeRunTimeout)
```

probe/pkg/runtime/runtime.go
```diff
@@ -46,18 +51,22 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
...
+	defer func() {
+		ctxCleanup, cancel := context.WithTimeout(context.Background(), r.Config.ProbeCleanUpTimeout)
+		defer cancel()
+		cleanupDone := concurrency.NewSignal()
+		go func() {
+			if err := r.probe.CleanUp(ctxCleanup, cleanupDone); err != nil {
+				glog.Error(err)
+			}
+		}()
+		cleanupDone.Wait()
```
dhaus67, line 65:
```suggestion
		select {
		        case <-cleanupCtx.Done():
		            // You probably want to catch the cleanup context being done here as well and log it, otherwise you will not catch it with the blocking call on `cleanupDone.Wait()`.
		        case <- cleanupDone.Done():
		           // The concurrency.Signal does expose this handy `Done()` function, which is a channel that will be closed if the signal was triggered.
		}
```

probe/pkg/runtime/runtime_test.go
```diff
@@ -0,0 +1,76 @@
...
+			require.NoError(t, err, "failed to create runtime")
+			ctx, cancel := context.WithTimeout(context.TODO(), testConfig.ProbeRunTimeout)
+			defer cancel()
+
+			err = runtime.RunSingle(ctx)
+
+			assert.ErrorContains(t, err, "probe run failed", "expected an error during probe run")
+			assert.ErrorContains(t, err, "context deadline exceeded", "expected timeout error")
+			assert.Equal(t, 1, len(tc.mockFMClient.GetCentralsCalls()), 1, "must retrieve central list for clean up")
+			assert.Equal(t, 1, len(tc.mockFMClient.DeleteCentralByIdCalls()), 1, "must delete central for clean up")
```
dhaus67, line 72: Otherwise, you would have an index out of bounds panic in l.73 in case `tc.mockFMClient.DeleteCentralByIdCalls() == 0`
```suggestion
			require.Equal(t, 1, len(tc.mockFMClient.DeleteCentralByIdCalls()), 1, "must delete central for clean up")
```

probe/pkg/runtime/runtime_test.go
```diff
@@ -0,0 +1,76 @@
...
+	ProbePollPeriod:     10 * time.Millisecond,
+	ProbeCleanUpTimeout: 100 * time.Millisecond,
+	ProbeRunTimeout:     100 * time.Millisecond,
+	ProbeRunWaitPeriod:  10 * time.Millisecond,
+	ProbeName:           "pod",
+	ProbeNamePrefix:     "probe",
+	RHSSOClientID:       "client",
+}
+
+func TestRunSingle(t *testing.T) {
```
dhaus67, line 27: It would be great to additionally have a test case which tests the behavior upon a timeout within the Cleanup section.

I feel like this would require additional work to make the code better testable, meaning defining an interface for `Probe` and generating mocks for that, ensuring that runtime depends on the interface etc.

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,208 @@ package probe
...
+func (p *Probe) CleanUp(ctx context.Context, done concurrency.Signal) error {
+	defer done.Signal()
+
+	centralList, _, err := p.fleetManagerClient.GetCentrals(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err, "could not retrieve central list")
+	}
+
+	for i := range centralList.Items {
+		central := centralList.Items[i]
```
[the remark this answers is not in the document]
stehessel: This doesn't work because of `&central` below, which does not work if `central` is part of the loop scope.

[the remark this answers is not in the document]
dhaus67: Ah, I missed that, disregard this one then.

probe/pkg/runtime/runtime_test.go
stehessel: This could be tested by putting a timeout into a mocked `GetCentralsFunc`, but it's probably a better idea to do the testing improvement you mentioned. That way we would have to deal with a mocked fleetmanager client in `runtime_test.go` and `cli_test.go` at all.

stehessel:
> 1. Avoiding creating an infinite number of probe Centrals

This is already covered by a finite quota for the probe service account. I don't think we need to add additional logic to the probe or fleet manager.

> 2. Testing the probe

I would suggest to add a `probe run` against the stage fleet manager to the CI pipeline. But I would do that as a follow up.

probe/pkg/runtime/runtime.go
stehessel: I'm fine with this change, just curious: This is what I expect to happen in case the `CleanUp` context expires:
* The context sensitive requests return with context exceeded error.
* This leads to `CleanUp` returning and signalling `cleanupDone` because of the `defer` statement.
* This then ends the blocking `cleanupDone.Wait` call.

Is there any scenario where the signal is not triggered in the defer and we would dead lock without this change?

probe/pkg/runtime/runtime_test.go
stehessel: Added a test that times out in `CleanUp`.

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,219 @@ package probe
...
+	}
+
+	return p.deleteCentral(ctx, central)
+}
+
+// CleanUp remaining probe resources.
+func (p *ProbeImpl) CleanUp(ctx context.Context, done concurrency.Signal) error {
+	defer done.Signal()
+
+	centralList, _, err := p.fleetManagerClient.GetCentrals(ctx, nil)
```
misberner, line 84: There's a 15m timeout for cleaning up, yet we give up immediately if fleet manager spits out an error. This seems a bit odd to me. I would have expected cleaning up to work like a reconciliation loop: attempting the same sequence of steps (list - filter - delete - wait for deleted) over and over again, until we either have confirmation that all centrals are gone, or the timeout expires.

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,219 @@ package probe
...
-	if expired := concurrency.WaitWithTimeout(ctx, 5*time.Second); expired {
-		return errors.Wrap(ctx.Err(), "probe run expired")
+// Verify that the Central instance has the expected properties and that the
+// Central UI is reachable.
+func (p *ProbeImpl) verifyCentral(ctx context.Context, centralRequest *public.CentralRequest) error {
+	if centralRequest.InstanceType != types.STANDARD.String() {
+		return errors.Errorf("central has wrong instance type: expected %s, got %s", types.STANDARD.String(), centralRequest.InstanceType)
+	}
+
+	if err := p.pingURL(ctx, centralRequest.CentralUIURL); err != nil {
```
misberner, line 135: So, expanding on my [comment from the previous PR](https://github.com/stackrox/acs-fleet-manager/pull/532/files#r1012019030) a bit: I see now that there are no explicit retries here in the surrounding logic. I wonder, though, if relying on HTTP client-level retries makes sense, for three different reasons:
1. absence of logging. While the client attempts to ping central, nothing will be printed on the console. I think it would be helpful to log intermediate attempts.
2. lack of control. I think we're _expecting_ that the first few attempts will fail, because the image will need to be pulled, central will need to be initialized before marking itself ready, router configs will need to be updated etc. To me, 4 retries with a 1 sec minimum seem very low. Not sure what the exponential base and the jitter are, but naively, it looks like we might give up after as little as ~10-15s. That is not enough for image pull + central startup, generally speaking. Somewhat related: the retry parameters, especially the point at which to give up, are not related to the overall probe execution timeout at all. While there certainly is some value in detecting error conditions that won't resolve themselves automatically, I would say that, in the absence of any such condition, we should keep retrying until the time budget is exhausted.
3. adequacy of retry policy. The default retry policy by the library makes sense when assuming we are interacting with an established system, but I'm not sure whether it's lenient enough to deal with a system that is just in the process of starting up. For example, race conditions when provisioning routes might be an issue. DNS resolution failures are retriable, and when an OpenShift router gets hit on a route that it doesn't know, it sends a 503 (service unavailable), which is also retriable. However, among the non-retriable errors are certificate validation errors. And I think such an error _could_ occur spuriously because the `CentralUIURL` is the _external_ URL, not the one corresponding to a route's default DNS (`<route>-<namespace>.apps.<cluster DNS>`). A 503 does not help in this context if it's sent back with an invalid certificate (matching `*.apps.<cluster DNS>`, whereas we'd need `*.acs.rhcloud.com` or so). It's of course good to not re-invent the wheel when it comes to deciding whether an HTTP failure is retriable or not, but I do suspect that some customizations are needed.

Have you done any extended tests of running the probe? If so, and you observed no flakes, maybe these concerns are unwarranted. However, I do think they should be addressable in a fairly complete fashion by using techniques you've already used elsewhere in the probe logic, namely explicit retries with a ticker.

misberner (changes requested):

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,208 @@ package probe
...
+	glog.Infof("elapsed time: %v", time.Since(start))
+}
+
+func (p *Probe) newCentralName() (string, error) {
+	rnd := make([]byte, 8)
+	if _, err := rand.Read(rnd); err != nil {
+		return "", errors.Wrapf(err, "reading random bytes for unique central name")
+	}
+	rndString := hex.EncodeToString(rnd)
+	return fmt.Sprintf("%s-%s-%s", p.config.ProbeNamePrefix, p.config.ProbeName, rndString), nil
```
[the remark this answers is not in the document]
stehessel: My intention was to use Kubernetes `valueFrom` to set `ProbeName` to the pod name of the probe. Afaik this does not allow further string formatting, so to keep the `probe-` I added the prefix.

probe/pkg/probe/probe.go
stehessel: I have not seen any flakes so far. Note that the probe already ensures that the Central is "ready", although I'm not exactly sure how that is determined wrt to the errors you mentioned. But I'm fine with switching to an explicit ticker based retry.

probe/pkg/probe/probe_test.go
```diff
@@ -0,0 +1,402 @@
+package probe
```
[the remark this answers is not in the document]
stehessel: @SimonBaeumer I added more tests wrt timeouts in `probe_test.go` in [4e8c2a0](https://github.com/stackrox/acs-fleet-manager/pull/533/commits/4e8c2a0b7dfb445d6a7c1cfd7e9b6209a9d98d5d).

probe/pkg/probe/probe.go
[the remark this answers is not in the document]
misberner: It doesn't allow string formatting directly, but a little-known property of K8s env vars is that they actually allow [variable substitution](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/). Hence, you could do
```
env:
- name: POD_NAME
  valueFrom:
    fieldRef:
      fieldPath: metadata.name
- name: PROBE_NAME
  value: "probe-$(POD_NAME)"
```
Note that the order is crucial, variables must be defined before being referenced.

I don't want to say you _should_ be using something like the above, but it _is_ possible.

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,245 @@ package probe
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Wrap(ctx.Err(), "cleanup centrals timed out")
+		case <-ticker.C:
+			centralList, _, err := p.fleetManagerClient.GetCentrals(ctx, nil)
+			if err != nil {
+				glog.Warningf("could not list centrals: %s", err)
+				continue
```
misberner, line 92: I think it would be good for readability to factor the code under the `case <-ticker.C` branch out into a separate function.

misberner: Actually, it might be worthwhile factoring out also the outer loop into a `retryUntilSucceeded(ctx context.Context, fn func(context.Context) error, interval time.Duration)` or so.

probe/pkg/runtime/runtime.go
```diff
@@ -45,19 +43,28 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
...
-func (r *Runtime) RunSingle(ctx context.Context) error {
-	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+func (r *Runtime) RunSingle(ctx context.Context) (errReturn error) {
+	probeRunCtx, cancel := context.WithTimeout(ctx, r.Config.ProbeRunTimeout)
 	defer cancel()
-	defer r.CleanUp()
+	defer func() {
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), r.Config.ProbeCleanUpTimeout)
+		defer cancel()
+		cleanupDone := concurrency.NewSignal()
```
misberner, line 52: It isn't entirely clear to me why `CleanUp` runs in a goroutine and you use a signal here, given that `CleanUp` takes in a context and thus should respect context cancellations. I _suspect_ that the motivation is ensuring that the deferred cleanup does not run for longer than `ProbeCleanUpTimeout`, regardless of whether `CleanUp` is well-behaved (i.e., respecting context cancellations) or not. If this is the case, there should be a comment here motivating the goroutine/signal approach.

probe/pkg/runtime/runtime.go
```diff
@@ -45,19 +43,28 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
...
+		go func() {
+			defer cleanupDone.Signal()
```
misberner, line 54: This is a deferred call in a goroutine launched from a deferred func. I think factoring one of those functions out would be a good change.

probe/pkg/runtime/runtime.go
```diff
@@ -45,19 +43,28 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
...
 
 // RunSingle executes a single probe run.
-func (r *Runtime) RunSingle(ctx context.Context) error {
-	ctxTimeout, cancel := context.WithTimeout(ctx, r.Config.RuntimeRunTimeout)
+func (r *Runtime) RunSingle(ctx context.Context) (errReturn error) {
+	probeRunCtx, cancel := context.WithTimeout(ctx, r.Config.ProbeRunTimeout)
 	defer cancel()
-	defer r.CleanUp()
+	defer func() {
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), r.Config.ProbeCleanUpTimeout)
```
misberner, line 50: I think I said this in a previous PR already, but it is a bit unfortunate that `RunSingle` now deliberately disrespects the `ctx` passed to it due to cleanup. I don't have a good solution, though

misberner (changes requested):
I think a few refactorings would really help readability, but overall the approach looks good. Thanks for all the effort iterating on this!

This is probably the last review I will be able to provide. Feel free to simply dismiss my review once you have addressed the changes.

probe/pkg/runtime/runtime.go
stehessel: Maybe I misunderstood your previous comments, but I thought that you and @dhaus67 were explicitly asking for the go routine/signal approach for the clean up function. That's why I used it here, however to my understanding, the new `context.Background()` would be sufficient as well.

stehessel:
> I think a few refactorings would really help readability, but overall the approach looks good. Thanks for all the effort iterating on this!
This is probably the last review I will be able to provide. Feel free to simply dismiss my review once you have addressed the changes.

Thank you as well Malte, for putting in a lot of effort to give a detailed review. I learned quite a lot, and I think it will be worth it in the end.

probe/pkg/runtime/runtime.go
stehessel: done

probe/pkg/probe/probe.go
stehessel: Done, also for the other retry-type functions.

probe/pkg/runtime/runtime.go
```diff
@@ -45,19 +43,32 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
...
+		defer cancel()
+		cleanupDone := concurrency.NewSignal()
+
+		go r.scheduleCleanup(cleanupCtx, &cleanupDone)
+
+		select {
+		case <-cleanupCtx.Done():
+			errReturn = cleanupCtx.Err()
```
dhaus67, line 58: I haven't seen the usage of named return parameters in combination with go routine as such here - will this actually lead to the `context cancelled / context deadline exceeded` error being returned from `RunSingle` here?

Either way, probably nice to have a comment describing this here, it wasn't not totally obvious to me at first sight.

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,270 @@ package probe
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Wrap(ctx.Err(), "retry failed")
+		case <-ticker.C:
+			err := fn(ctx)
+			if err != nil {
+				glog.Warning(err)
```
dhaus67, line 246: Should we log the error here or silently ignore it?
Currently, we do log every error in `cleanUpFunc` and the default error we return pretty much doesn't say a lot (`central clean up not successful`).
I know that it's not necessarily given that this function knows the details of the passed `fn` - so either we should stop logging within `cleanUpFunc` and propagate the whole error context and log it here, or silently ignore here and log within `cleanUpFunc`.

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,270 @@ package probe
...
+
+	centralResp, err := retryUntilSucceededWithResponse(ctx, ensureFunc, p.config.ProbePollPeriod)
+	if err != nil {
+		return nil, errors.Wrap(err, "ensure central state failed")
+	}
+	return centralResp, nil
+}
+
+func (p *ProbeImpl) ensureCentralDeleted(ctx context.Context, centralRequest *public.CentralRequest) error {
+	ensureFunc := func(context.Context) error {
```
dhaus67, line 194: Can we factor this out to a separate function instead of a inline function?

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,270 @@ package probe
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "central instance %s not reachable", centralRequest.Id)
+		}
+
+		if centralResp.Status == targetState {
+			glog.Infof("central instance %s is in %q state", centralResp.Id, targetState)
+			return &centralResp, nil
+		}
+		return nil, errors.Errorf("central instance %s not in target state %q", centralRequest.Id, targetState)
+	}
```
dhaus67, line 184: Can we factor this out to a separate function instead of a inline function?

probe/pkg/probe/probe.go
```diff
@@ -2,21 +2,270 @@ package probe
...
+		return nil, errors.Wrap(err, "ensure central state failed")
+	}
+	return centralResp, nil
+}
+
+func (p *ProbeImpl) ensureCentralDeleted(ctx context.Context, centralRequest *public.CentralRequest) error {
+	ensureFunc := func(context.Context) error {
+		_, response, err := p.fleetManagerClient.GetCentralById(ctx, centralRequest.Id)
+		if err != nil {
+			if response != nil && response.StatusCode == http.StatusNotFound {
```
dhaus67, line 197: I suppose in case `response != nil` you also want to close the response body.

probe/pkg/runtime/runtime.go
dhaus67: I think the previous iteration had neither context nor signal - I suppose that's the misunderstanding here and where the suggestions came from (at the very least I can speak for the suggestions I made).

I think we now doubled up on the way the function should behave - we have the context cancellation _and_ the signal approach.

I personally like the signal approach more for these cleanup tasks, however, since you already implemented it, you can also keep the context cancellation and remove the signal part.

probe/pkg/probe/probe_test.go
```diff
@@ -0,0 +1,443 @@
...
+				},
+				GetCentralByIdFunc: func(ctx context.Context, id string) (public.CentralRequest, *http.Response, error) {
+					central := public.CentralRequest{}
+					err := errors.Errorf("%d", http.StatusInternalServerError)
+					return central, nil, err
+				},
+			},
+		},
+		{
+			testName: "timeout on context deadline exceeded",
```
dhaus67, line 108:
```suggestion
			testName: "create central times out",
```

probe/pkg/probe/probe_test.go
```diff
@@ -0,0 +1,443 @@
...
+					return public.CentralRequest{
+						Id:     "id-42",
+						Name:   "probe-pod-42",
+						Status: constants.CentralRequestStatusDeleting.String(),
+					}, nil, nil
+				},
+			},
+		},
+		{
+			testName: "timeout on context deadline exceeded",
```
dhaus67, line 291:
```suggestion
			testName: "central deletion times out",
```

probe/pkg/probe/probe_test.go
```diff
@@ -0,0 +1,443 @@
...
+			mockFMClient: &fleetmanager.PublicClientMock{
+				GetCentralsFunc: func(ctx context.Context, localVarOptionals *public.GetCentralsOpts) (public.CentralRequestList, *http.Response, error) {
+					centralList := public.CentralRequestList{}
+					err := errors.Errorf("%d", http.StatusInternalServerError)
+					return centralList, nil, err
+				},
+			},
+		},
+		{
+			testName: "timeout on context deadline exceeded",
```
dhaus67, line 412:
```suggestion
			testName: "clean up times out",
```

dhaus67 (approved):
Thanks for addressing all the feedback and iterating over all the comments!

Just minor comments from my side, otherwise LGTM.

probe/pkg/runtime/runtime.go
stehessel: The return is not assigned from the goroutine, but just from a deferred function. So yeah it should return the context error. However on second look, I this might not be such a good idea in case `Execute` already returned an error, because then we are overwriting that. So I think just logging the clean up error would be better.

probe/pkg/probe/probe.go
stehessel: The OpenAPI client always closes the body, that's why I removed it here.

SimonBaeumer (approved):

stehessel:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/533?utm_source=stack-comment).


Job Reference: WSxW0jmiiYoVqUGHV0ml

PR #535, opened 2022-10-31
ROX-12153: Add cloud account id to API, database and AMS call

ivan-degtiarenko:
## Description
This PR:
* introduces `cloudAccountID` field to `CentralRequest`(API, database)
* Checks `cloudAcccountID` is correctly set to one of the organizations cloud accounts, or set to none if organization has no cloud accounts; returns an error otherwise
* sends `cloudAccountID` in `ReserveQuota` call

## Test manual

```
curl -X POST -H "Authorization: Bearer $(ocm token)" -H "Content-Type: application/json" \
  http://127.0.0.1:8000/api/rhacs/v1/centrals\?async\=true \
  -d '{"name": "test-central-1", "multi_az": true, "cloud_account_id": "<correct cloud account id>", "cloud_provider": "aws", "region": "us-east-1"}'

```
resulted in successfully obtaining quota with billing model marketplace.

while 

```
curl -X POST -H "Authorization: Bearer $(ocm token)" -H "Content-Type: application/json" \
  http://127.0.0.1:8000/api/rhacs/v1/centrals\?async\=true \
  -d '{"name": "test-central-2", "multi_az": true, "cloud_account_id": "<incorrect cloud account id>", "cloud_provider": "aws", "region": "us-east-1"}'
```
resulted in an error
```
{"id":"122","kind":"Error","href":"/api/rhacs/v1/errors/122","code":"RHACS-MGMT-122","reason":"Cloud account ID is not set up properly: Request cloud account does not match organization cloud accounts%!(EXTRA string=<incorrect cloud account id>)","operation_id":"cdrqum07t2ugtro9lti0"}
```


internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -130,9 +132,14 @@ func (q amsQuotaService) ReserveQuota(dinosaur *dbapi.CentralRequest, instanceTy
 	rr.BillingModel(amsv1.BillingModel(bm))
 	glog.Infof("Billing model of Central request %s with quota type %s has been set to %s.", dinosaur.ID, instanceType.GetQuotaType(), bm)
 
+	if err := q.verifyCloudAccountInAMS(dinosaur, orgID); err != nil {
```
stehessel, line 135: We should only do something with the cloud accounts if the billing model is `marketplace`. Note that the billing model is determined via `getAvailableBillingModelFromDinosaurInstanceType` automagically by fleet manager. This is taken from the Kafka implementation, but I'm not a big fan of that. I can imagine in the future we will also add the billing model as an explicit request parameter. In any case, for `standard` billing model (e.g. yearly subscriptions, sold via regular Red Hat sales) we should just ignore the cloud account id.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -158,6 +165,27 @@ func (q amsQuotaService) ReserveQuota(dinosaur *dbapi.CentralRequest, instanceTy
...
+		svcErr := errors.ToServiceError(err)
+		return errors.NewWithCause(svcErr.Code, svcErr, "Error getting cloud accounts")
+	}
+
+	if dinosaur.CloudAccountID == "" && len(cloudAccounts) != 0 {
+		return errors.CloudAccountIDNotSetupProperly("Request cloud account is not setup even though organization has cloud accounts")
+	}
+	if dinosaur.CloudAccountID == "" && len(cloudAccounts) == 0 {
+		return nil
+	}
```
stehessel, line 180: I would probably write this as such to avoid the `&&`s. But feel free to ignore, I think this is personal taste.

```suggestion
	if dinosaur.CloudAccountID == "" {
	  if len(cloudAccounts) != 0 {
		return errors.CloudAccountIDNotSetupProperly("Request cloud account is not setup even though organization has cloud accounts")
	  }
	  return nil
}
```

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -158,6 +165,27 @@ func (q amsQuotaService) ReserveQuota(dinosaur *dbapi.CentralRequest, instanceTy
...
+	for _, ca := range cloudAccounts {
```
stehessel, line 181: This would be much nicer to write in Python 😅 . Not sure if there is a more elegant way in golang though

pkg/errors/errors.go
```diff
@@ -723,3 +727,9 @@ func FailedToCheckQuota(reason string, values ...interface{}) *ServiceError {
 	message := fmt.Sprintf("%s: %s", ErrorFailedToCheckQuotaReason, reason)
 	return New(ErrorFailedToCheckQuota, message, values...)
 }
+
+// CloudAccountIDNotSetupProperly ...
+func CloudAccountIDNotSetupProperly(reason string, values ...interface{}) *ServiceError {
```
stehessel, line 732: Maybe `InvalidCloudAccountID` is a better name? `NotSetupProperly` sounds weirdly informal.

stehessel:
LGTM overall. The situation with the billing model is a bit unfortunate, because currently we don't really make this distinction in the UI. Right now I think it's probably best to favour `marketplace` over `standard` billing model if both exist, such that we don't run into the situation where a cloud account id is selected in the UI, but the quota is registered with a `standard` billing model. As mentioned in the future it might make sense to make the billing model explicitly part of the request, as well as show a selection of billing model to the user in the UI.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -158,6 +165,27 @@ func (q amsQuotaService) ReserveQuota(dinosaur *dbapi.CentralRequest, instanceTy
 	return "", errors.InsufficientQuotaError("Insufficient Quota")
 }
 
+func (q amsQuotaService) verifyCloudAccountInAMS(dinosaur *dbapi.CentralRequest, orgID string) *errors.ServiceError {
+	cloudAccounts, err := q.amsClient.GetCustomerCloudAccounts(orgID, []string{RHACSMarketplaceQuotaID})
+	if err != nil {
+		svcErr := errors.ToServiceError(err)
+		return errors.NewWithCause(svcErr.Code, svcErr, "Error getting cloud accounts")
```
SimonBaeumer, line 172: Can you log the cloud account IDs requested here please?

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -158,6 +165,27 @@ func (q amsQuotaService) ReserveQuota(dinosaur *dbapi.CentralRequest, instanceTy
...
+	}
+
+	if dinosaur.CloudAccountID == "" && len(cloudAccounts) != 0 {
+		return errors.CloudAccountIDNotSetupProperly("Request cloud account is not setup even though organization has cloud accounts")
```
SimonBaeumer, line 176: Can you log the cloud account IDs requested here please?

internal/dinosaur/pkg/services/quota/ams_quota_service.go
SimonBaeumer: Is the loop necessary because AMS returns all customer's cloud accounts? Can't we request an account by providing provider + ID?

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -462,10 +506,169 @@ func Test_AMSReserveQuota(t *testing.T) {
...
+				"testUser",
+				"cloudAccountID",
+			},
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					ClusterAuthorizationFunc: func(cb *v1.ClusterAuthorizationRequest) (*v1.ClusterAuthorizationResponse, error) {
+						sub := v1.SubscriptionBuilder{}
+						sub.ID("1234")
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
```
SimonBaeumer, line 643: Can you rename `ca` to `cloudAccount`? ca is an abbreviation used to often for certificate authority.

```suggestion
						cloudAccount, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
```

internal/dinosaur/pkg/services/quota/ams_quota_service.go
SimonBaeumer, line 181: Can you rename `ca` to `cloudAccount`? ca is an abbreviation used to often for certificate authority.

```suggestion
	for _, cloudAccount := range cloudAccounts {
```

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -462,10 +506,169 @@ func Test_AMSReserveQuota(t *testing.T) {
...
+				"12231",
+				"testUser",
+				"different cloudAccountID",
+			},
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					ClusterAuthorizationFunc: func(cb *v1.ClusterAuthorizationRequest) (*v1.ClusterAuthorizationResponse, error) {
+						sub := v1.SubscriptionBuilder{}
+						sub.ID("1234")
+						sub.Status("Active")
```
SimonBaeumer, line 602: Can you add constants for the string fields here?
Is it worth adding constants for the`sub.Status` field in a non-test package? 🤔

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -462,10 +506,169 @@ func Test_AMSReserveQuota(t *testing.T) {
...
+					},
+					GetCustomerCloudAccountsFunc: func(externalID string, quotaIDs []string) ([]*v1.CloudAccount, error) {
+						return nil, errors.New("unsuccessful cloud accounts test call")
+					},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "cloud account id in request is empty while cloud_accounts response is not",
```
SimonBaeumer, line 551: Can you also write what is expected?

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -462,10 +506,169 @@ func Test_AMSReserveQuota(t *testing.T) {
...
+						return ca, nil
+					},
+					GetOrganisationIDFromExternalIDFunc: func(externalId string) (string, error) {
+						return fmt.Sprintf("fake-org-id-%s", externalId), nil
+					},
+					GetQuotaCostsForProductFunc: func(organizationID, resourceName, product string) ([]*v1.QuotaCost, error) {
+						rrbq1 := v1.NewRelatedResource().BillingModel(string(v1.BillingModelMarketplace)).Product(string(ocm.RHACSTrialProduct)).ResourceName(resourceName).Cost(0)
+						qcb1, err := v1.NewQuotaCost().Allowed(0).Consumed(2).OrganizationID(organizationID).RelatedResources(rrbq1).Build()
+						if err != nil {
+							panic("unexpected error")
```
SimonBaeumer, line 573: Use `require.NoError(err)` instead

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -462,10 +506,169 @@ func Test_AMSReserveQuota(t *testing.T) {
...
+						return []*v1.CloudAccount{
+							cloudAccount,
+						}, nil
+					},
 				},
 			},
 			wantErr: true,
 		},
+		{
+			name: "cloud account id in request does not match cloud_accounts response",
```
SimonBaeumer, line 591:
```suggestion
			name: "cloud account id in request does not match ids in cloud_accounts response",
```

SimonBaeumer (changes requested):

internal/dinosaur/pkg/services/quota/ams_quota_service.go
ivan-degtiarenko: @SimonBaeumer unfortunately, AMS API does not allow for this level of granularity, see how  [GetCustomerCloudAccounts](https://github.com/stackrox/acs-fleet-manager/blob/ivan%2Frox-12153-add-aws-account-central-request/pkg/client/ocm/client.go#L606-L628) is implemented

internal/dinosaur/pkg/services/quota/ams_quota_service.go
ivan-degtiarenko: Not sure what you mean by cloud account IDs requested - here we request all cloud account IDs for organization

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
ivan-degtiarenko: Here `ca` stands for `clusterAuthorization` - should we change to `cloudAuthorizationResponse`?

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
ivan-degtiarenko: I extracted this subscription into a separate method.

> Is it worth adding constants for the sub.Status field in a non-test package

considering it is used twice within this repo, I wouldn't bother

internal/dinosaur/pkg/services/quota/ams_quota_service.go
SimonBaeumer: Ah okay, makes sense. Would you think it is useful to log the orgId?

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
SimonBaeumer: Good suggestion, +1 for `cloudAuthorizationResp`

SimonBaeumer (approved):

internal/dinosaur/pkg/services/quota/ams_quota_service.go
ivan-degtiarenko: Counter-proposal: maybe we can add `orgID` to error? Our HTTP handlers should log all errors with HTTP code between 400 and 499

stehessel (approved):

PR #541, opened 2022-11-03
ROX-12480: run terraform script from a github action

porridge:

- [x] ~Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~ :point_right: planning to do docs overhaul once this is well tested and applies to production too

## Test manual

The script is already [tested in a previous PR](https://github.com/stackrox/acs-fleet-manager/pull/505), but no good way to test the integration before the actual push.

### DIFF
--- a/.github/workflows/deploy-stage.yaml
+++ b/.github/workflows/deploy-stage.yaml
@@ -6,8 +6,8 @@ on:
       - main
 
 jobs:
-  oidc-smoke-test:
-    name: "Verify GitHub Authentication"
+  deploy:
+    name: "Re-terraform stage clusters"
     runs-on: ubuntu-latest
     permissions:
       id-token: write
@@ -29,10 +29,10 @@ jobs:
         with:
           aws-region: ${{ secrets.AWS_REGION }}
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
-      - name: List Parameter Store services
+      - name: Run terraforming THE on stage cluster
         env:
           USE_AWS_VAULT: false
         run: |
-          source ./scripts/lib/external_config.sh
-          init_chamber
-          run_chamber list-services
+          set -euo pipefail
+          cd dp-terraform/helm/rhacs-terraform
+          ./terraform_cluster.sh stage acs-stage-dp-01
--- a/.github/workflows/deploy-stage.yaml
+++ b/.github/workflows/deploy-stage.yaml
@@ -29,7 +29,7 @@ jobs:
         with:
           aws-region: ${{ secrets.AWS_REGION }}
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-stage.yaml
```diff
@@ -29,10 +29,10 @@ jobs:
...
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
-      - name: List Parameter Store services
+      - name: Run terraforming THE on stage cluster
         env:
           USE_AWS_VAULT: false
         run: |
-          source ./scripts/lib/external_config.sh
-          init_chamber
-          run_chamber list-services
+          set -euo pipefail
```
kovayur, line 36: nit: for the below commands this should be enough.
```suggestion
          set -e
```

.github/workflows/deploy-stage.yaml
```diff
@@ -29,10 +29,10 @@ jobs:
         with:
           aws-region: ${{ secrets.AWS_REGION }}
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
-      - name: List Parameter Store services
+      - name: Run terraforming THE on stage cluster
```
kovayur, line 32: "THE" is in the right place? I'm not sure I understand the meaning. 😅

kovayur (approved):
All good. Just minor remarks, they can be ignored

kovayur:
Thinking about the change another minute, I see that there were no changes introduced for `FLEETSHARD_SYNC_IMAGE` variable. It means that we'll redeploy the same image on each merge. Do you plan to change it in a future?

porridge:
FTR, this is waiting for:
- https://github.com/stackrox/acs-fleet-manager/pull/550
- https://github.com/stackrox/acs-fleet-manager/pull/548 (not strictly required for staging, but better safe than sorry)

porridge:
> Thinking about the change another minute, I see that there were no changes introduced for the `FLEETSHARD_SYNC_IMAGE` variable. It means that we'll redeploy the same image on each merge. Do you plan to change it in a future?

Yes, I'm taking baby steps for now.

.github/workflows/deploy-stage.yaml
porridge: Fixed. The capitalization was a poor joke, hinting at the fact that in the future there might be more clusters, especially in production, and we'll need to figure out how to process them all.

.github/workflows/deploy-stage.yaml
porridge: Yeah, I'm being "future-proof" :sweat_smile:

porridge: I mean, scripts tend to grow organically and people (myself included) often forget to extend the range of useful checks.

kovayur (approved):

porridge:
Force-merging since the e2e test is irrelevant here, and the cause was an infrastructure problem anyway.

PR #542, opened 2022-11-03
ROX-13004: Provision RDS instances and clusters

vladbologa:
## Description

Add support for provisioning / deprovisioning RDS Aurora Serverless DBs, to use with managed Centrals. 

The way it's implemented now is via an interface that has a **blocking** method to provision DBs, and a **non-blocking** method to deprovision them. This is because we need to have a provisioned DB to be able to successfully start Central (and to create its CR), but for deletion I considered that this can also be done in a fire-and-forget way (it's also unlikely that deletion would fail, once the process was started successfully).

The steps to provision a DB are the following:
1. A DB password is created and the secret is stored in etcd. This is done in the first step to ensure that the process can be interrupted. E.g. if we initiate the creation of a DB and store the password only at the end, if the process is restarted (e.g. fleetshard-sync restarts) we won't know the password anymore.
2. Initiate creation of a DB cluster using the aws-go-sdk (asynchronously, we don't wait for the creation to be finished - this is what happens also when creating a DB instance in the AWS console)
3. Initiate the creation of a DB instance (in the cluster created at step 2) using the aws-go-sdk
4. Wait until the the DB is available
5. Create the Central CR (append the DB related info to its CR, including the connection string that we retrieve from the DB cluster)

Deprovisioning the DB is more straightforward, we simply delete the instance and cluster using the aws-sdk-go, and then delete the password secret.

The name of the database is not store anywhere, because it's based on the central namespace name, and it therefore can be recreated at any time.

The managed DB functionality is disabled by default, the `MANAGED_DB_ENABLED` env var must be set to true for it to be enabled. There are several other env vars that need to be set for this to be working (AWS credentials of an account that has permissions to create and delete DBs, a security group to assign to the DB instances, and the DB subnet in which to create them). The other options are currently hardcoded (region, DB settings etc)

One problem that I see is that there is no support for unrecoverable errors. If something goes wrong and an error is returned during the execution of Reconcile, the current Reconcile will fail, but the Runtime will try again until the Central is deleted. This is not an issue specific to RDS provisioning, but I thought it's worth mentioning.

## Test manual

Before the official 3.73 release, testing of the RDS feature is possible only by running an operator built from source that supports the feature (e.g. take the `release-3.73` branch).

The way I'm testing in a local cluster:
```
# run the operator locally (from stackrox/operator)
make install run

# all other commands are run from the acs-fleet-manager root dir
# bootstrap the test env
INSTALL_OPERATOR=NO INSTALL_OLM=NO ./dev/env/scripts/bootstrap.sh

# retrieve AWS credentials for the test account
kinit <username>
aws-saml.py

# set the appropriate env variables with the AWS credentials (from `~/.aws/credentials`)

# bring up fleet-manager and fleet-shard
INSTALL_OPERATOR=NO MANAGED_DB_ENABLED=TRUE MANAGED_DB_SECURITY_GROUP=sg-04dcc23a03646041c ./dev/env/scripts/up.sh

# create a test central
./scripts/create-central.sh

# verify that the Central is up and running
# verify in the AWS console that the DB is being created, has active connections and is finally deleted

#delete test central
curl -X DELETE -H "Authorization: Bearer $(ocm token)" -H "Content-Type: application/json" \
  http://127.0.0.1:8000/api/rhacs/v1/centrals/cdvq4mmsbpn001pjdfkg\?async\=true
```


fleetshard/pkg/central/rdsclient/rdsclient.go
```diff
@@ -0,0 +1,307 @@
...
+	dbInstanceQuery := &rds.DescribeDBInstancesInput{
+		DBInstanceIdentifier: aws.String(instanceID),
+		//TODO: add cluster Filter
+	}
+
+	dbClusterQuery := &rds.DescribeDBClustersInput{
+		DBClusterIdentifier: aws.String(clusterID),
+	}
+
+	for {
```
vladbologa, line 185: I want to check if this can be rewritten using event subscriptions (instead of polling for the state every 10 seconds).

fleetshard/pkg/central/rdsclient/rdsclient.go
```diff
@@ -0,0 +1,307 @@
...
+func newDeleteCentralDBClusterInput(clusterID string, skipFinalSnapshot bool) *rds.DeleteDBClusterInput {
+	return &rds.DeleteDBClusterInput{
+		DBClusterIdentifier: aws.String(clusterID),
+		SkipFinalSnapshot:   aws.Bool(skipFinalSnapshot),
+	}
+}
+
+func newRdsClient() (*rds.RDS, error) {
+	cfg := &aws.Config{
+		Region: aws.String(awsRegion),
```
vladbologa, line 298: passing of the credentials is missing here

fleetshard/pkg/central/rdsclient/rdsclient.go
porridge: I'd be cautious against adding too much AWS-specific code at this point, given:
- this might be a premature optimization,
- the possibility of having to support other cloud providers in the future, who might not have an events mechanism.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,313 @@
...
+	instanceID := centralNamespace + dbInstanceSuffix
+
+	if c.instanceExists(instanceID) {
+		status, err := c.instanceStatus(instanceID)
+		if err != nil {
+			return false, fmt.Errorf("getting DB instance status: %v", err)
+		}
+		if status != dbDeletingStatus {
+			//TODO: do not skip taking a final DB snapshot
+			glog.Infof("Deprovisioning RDS database cluster.")
```
ebensh, line 82: I think this is swapped with the message below - first we delete the instance, then the cluster

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	if c.clusterExists(clusterID) {
+		status, err := c.clusterStatus(clusterID)
+		if err != nil {
+			return false, fmt.Errorf("getting DB cluster status: %v", err)
+		}
+		if status != dbDeletingStatus {
+			//TODO: do not skip taking a final DB snapshot
+			glog.Infof("Deprovisioning RDS database instance.")
```
ebensh, line 97: Flipped here

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,313 @@
...
+type AWSCredentials struct {
+	// AccessKeyID is the AWS access key identifier.
+	AccessKeyID string
+	// SecretAccessKey is the AWS secret access key.
+	SecretAccessKey string
+	// SessionToken is a token required for temporary security credentials retrieved via STS.
+	SessionToken string
+}
+
+// EnsureDBProvisioned is a blocking function that makes sure that an RDS database was provisioned for a Central
```
ebensh, line 51: In runtime.go:134 this is provided ctx.Background(), and it looks like we don't ensure a timeout anywhere; this means the reconcile for this Central could get stuck and never recover, correct?

Should we at least ensure a timeout is present on the AWS API call? (or in the larger runtime.go set a timeout on the context there)

ebensh (changes requested):

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Right, I forgot to write a TODO about this, and then I forgot about it entirely. I'll look into it now.

vladbologa: I think I prefer setting a timeout in runtime.go, even though that will change behaviour unrelated to this ticket. But if someone has a strong reason not to do it, I'll move it to the AWS API call.

SimonBaeumer:
@vladbologa Can you describe the overall approach in the PR description? If possible highlighting areas which are more critical?

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+			return "", fmt.Errorf("waiting for RDS instance to be available: %v", ctx.Err())
+		}
+
+		result, err := c.rdsClient.DescribeDBInstances(dbInstanceQuery)
+		if err != nil {
+			return "", fmt.Errorf("retrieving DB instance state: %v", err)
+		}
+
+		if len(result.DBInstances) != 1 {
+			return "", fmt.Errorf("unexpected number of DB instances: %v", err)
```
SimonBaeumer, line 203: I would assume that in this case `err` is `nil` no?
Would it be more useful to print the returned DB instances instead?

```
"unable to determine DB instances, found more than 1 instance. Instances: ...."
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func (c *RDSProvisioningClient) waitForInstanceToBeAvailable(ctx context.Context, instanceID string, clusterID string) (string, error) {
+	dbInstanceQuery := &rds.DescribeDBInstancesInput{
+		DBInstanceIdentifier: aws.String(instanceID),
+	}
+
+	dbClusterQuery := &rds.DescribeDBClustersInput{
+		DBClusterIdentifier: aws.String(clusterID),
+	}
```
SimonBaeumer, line 190: WDYTH about moving these to the function call directly?
Temporary variables are helpful to me if the execution path gets too long and bloated, in this case to me it created an indirection which was a bit harder to create a mental mapping on.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+
+	for {
+		if ctx.Err() != nil {
+			return "", fmt.Errorf("waiting for RDS instance to be available: %v", ctx.Err())
+		}
+
+		result, err := c.rdsClient.DescribeDBInstances(dbInstanceQuery)
```
SimonBaeumer, line 197:
```suggestion
		dbInstanceStatuses, err := c.rdsClient.DescribeDBInstances(dbInstanceQuery)
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+
+	return *clusterResult.DBClusters[0].Status, nil
+}
+
+func (c *RDSProvisioningClient) instanceStatus(instanceID string) (string, error) {
+	dbInstanceQuery := &rds.DescribeDBInstancesInput{
+		DBInstanceIdentifier: aws.String(instanceID),
+	}
+
+	instanceResult, err := c.rdsClient.DescribeDBInstances(dbInstanceQuery)
```
SimonBaeumer, line 175: Here as well like described (below or above :D), I like the effort of creating temporary variables for readability, in this case though it creates a mental mapping overhead for readers of the code in a simple execution path.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+		result, err := c.rdsClient.DescribeDBInstances(dbInstanceQuery)
+		if err != nil {
+			return "", fmt.Errorf("retrieving DB instance state: %v", err)
+		}
+
+		if len(result.DBInstances) != 1 {
+			return "", fmt.Errorf("unexpected number of DB instances: %v", err)
+		}
+
+		dbInstanceStatus := *result.DBInstances[0].DBInstanceStatus
```
SimonBaeumer, line 206: Can you use the `clusterStatus` for L197-L206? Looks like duplicate code on a first glance.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+	return err == nil
+}
+
+func (c *RDSProvisioningClient) instanceExists(instanceID string) bool {
+	dbInstanceQuery := &rds.DescribeDBInstancesInput{
+		DBInstanceIdentifier: aws.String(instanceID),
+	}
+
+	_, err := c.rdsClient.DescribeDBInstances(dbInstanceQuery)
+	return err == nil
```
SimonBaeumer, line 154: This check looks not accurate. If `err` contains a network related error it does not says anything about the existence of the database but would return `false` here.
Instead a specific `not exist-err` should be checked or the amount of db instances returned.

Also in other cases you verified that only 1 DB instance is returned from the query, here the check is not done. Is there a reason? 🤔

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+
+	return nil
+}
+
+func (c *RDSProvisioningClient) clusterExists(clusterID string) bool {
+	dbClusterQuery := &rds.DescribeDBClustersInput{
+		DBClusterIdentifier: aws.String(clusterID),
+	}
+
+	_, err := c.rdsClient.DescribeDBClusters(dbClusterQuery)
```
SimonBaeumer, line 144: See below feedback on `instanceExists`. I think it is not sufficient to only check for an error if a cluster exists.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	return c.waitForInstanceToBeAvailable(ctx, instanceID, clusterID)
+}
+
+// EnsureDBDeprovisioned is a function that initiates the deprovisioning of the RDS database of a Central
+// Unlike EnsureDBProvisioned, this function does not block until the DB is deprovisioned
+func (c *RDSProvisioningClient) EnsureDBDeprovisioned(centralNamespace string) (bool, error) {
+	clusterID := centralNamespace + dbClusterSuffix
+	instanceID := centralNamespace + dbInstanceSuffix
```
SimonBaeumer, line 73: I think coupling the namespace name is not a good idea. If the namespace schema changes our database names need a migration as well.
If you want to refer to the `rhacs-<instance-id>` model I think a function which compiles it explicitly is better to avoid side effects in later refactorings.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+
+	if c.instanceExists(instanceID) {
+		status, err := c.instanceStatus(instanceID)
+		if err != nil {
+			return false, fmt.Errorf("getting DB instance status: %v", err)
+		}
+		if status != dbDeletingStatus {
+			//TODO: do not skip taking a final DB snapshot
+			glog.Infof("Deprovisioning RDS database instance.")
```
SimonBaeumer, line 82: I think this log info is meaningless because it lacks context which database is provisioning.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+	}
+
+	if c.clusterExists(clusterID) {
+		status, err := c.clusterStatus(clusterID)
+		if err != nil {
+			return false, fmt.Errorf("getting DB cluster status: %v", err)
+		}
+		if status != dbDeletingStatus {
+			//TODO: do not skip taking a final DB snapshot
```
SimonBaeumer, line 96: Is this a requirement? If yes please create a ticket for it.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+			glog.Infof("Deprovisioning RDS database cluster.")
+			_, err := c.rdsClient.DeleteDBCluster(newDeleteCentralDBClusterInput(clusterID, true))
+			if err != nil {
+				return false, fmt.Errorf("deleting DB cluster: %v", err)
```
SimonBaeumer, line 100:
```suggestion
				return false, errors.Wrap(err, "deleting DB cluster")
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+			if err != nil {
+				return false, fmt.Errorf("deleting DB instance: %v", err)
+			}
+		}
+	}
+
+	if c.clusterExists(clusterID) {
+		status, err := c.clusterStatus(clusterID)
+		if err != nil {
+			return false, fmt.Errorf("getting DB cluster status: %v", err)
```
SimonBaeumer, line 93:
```suggestion
			return false, errors.Wrap(err, "getting DB cluster status")
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+		status, err := c.instanceStatus(instanceID)
+		if err != nil {
+			return false, fmt.Errorf("getting DB instance status: %v", err)
+		}
+		if status != dbDeletingStatus {
+			//TODO: do not skip taking a final DB snapshot
+			glog.Infof("Deprovisioning RDS database instance.")
+			_, err := c.rdsClient.DeleteDBInstance(newDeleteCentralDBInstanceInput(instanceID, true))
+			if err != nil {
+				return false, fmt.Errorf("deleting DB instance: %v", err)
```
SimonBaeumer, line 85:
```suggestion
				return false, errors.Wrap(err, "deleting DB instance")
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+// EnsureDBDeprovisioned is a function that initiates the deprovisioning of the RDS database of a Central
+// Unlike EnsureDBProvisioned, this function does not block until the DB is deprovisioned
+func (c *RDSProvisioningClient) EnsureDBDeprovisioned(centralNamespace string) (bool, error) {
+	clusterID := centralNamespace + dbClusterSuffix
+	instanceID := centralNamespace + dbInstanceSuffix
+
+	if c.instanceExists(instanceID) {
+		status, err := c.instanceStatus(instanceID)
+		if err != nil {
+			return false, fmt.Errorf("getting DB instance status: %v", err)
```
SimonBaeumer, line 78:
```suggestion
			return false, errors.Wrap(err, "getting DB instance status")
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+		}
+	}
+
+	return true, nil
+}
+
+func (c *RDSProvisioningClient) ensureDBClusterCreated(ctx context.Context, clusterID, centralNamespace, centralDbSecretName string) error {
+	if !c.clusterExists(clusterID) {
+		// cluster does not exist, create it
```
SimonBaeumer, line 110: Can be removed, comment is redundant to if-case.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+		glog.Infof("Provisioning RDS database cluster.")
```
SimonBaeumer, line 111: Log message lacks context, please add DB cluster name. 
Is it a useful information? Don't we only need that information in case a provisioning failed?

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+				return false, fmt.Errorf("deleting DB cluster: %v", err)
+			}
+		}
+	}
+
+	return true, nil
+}
+
+func (c *RDSProvisioningClient) ensureDBClusterCreated(ctx context.Context, clusterID, centralNamespace, centralDbSecretName string) error {
+	if !c.clusterExists(clusterID) {
```
SimonBaeumer, line 109: Can you change the check to an early return?

```
if c.clusterExists(clusterID) {
    return nil
}

// Logic to create DB cluster
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+		if err != nil {
+			return fmt.Errorf("creating DB cluster: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func (c *RDSProvisioningClient) ensureDBInstanceCreated(instanceID string, clusterID string) error {
+	if !c.instanceExists(instanceID) {
```
SimonBaeumer, line 126: Here also an early return to reduce n-path complexity.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	dbAvailableStatus = "available"
+	dbDeletingStatus  = "deleting"
+
+	awsRegion        = "us-east-1" // TODO: this should not be hardcoded
```
SimonBaeumer, line 22: I second the TODO, can you make it configurable?

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+	return nil
+}
+
+func (c *RDSProvisioningClient) ensureDBInstanceCreated(instanceID string, clusterID string) error {
+	if !c.instanceExists(instanceID) {
+		// instance does not exist, create it
+		glog.Infof("Provisioning RDS database instance.")
+		_, err := c.rdsClient.CreateDBInstance(newCreateCentralDBInstanceInput(clusterID, instanceID))
+		if err != nil {
+			// TODO: delete cluster if instance cannot be created?
```
SimonBaeumer, line 131: Why? Shouldn't the reconciler just retry to create a new instance?
If the timeout is reached our operations team gets notified because of a failed provisioning.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+			return fmt.Errorf("creating DB instance: %v", err)
```
SimonBaeumer, line 132:
```suggestion
			return errors.Wrap(err, "creating DB instance")
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+	instanceID := centralNamespace + dbInstanceSuffix
+
+	err := c.ensureDBClusterCreated(ctx, clusterID, centralNamespace, centralDbSecretName)
+	if err != nil {
+		return "", fmt.Errorf("ensuring DB cluster %s exists: %v", clusterID, err)
+	}
+
+	err = c.ensureDBInstanceCreated(instanceID, clusterID)
+	if err != nil {
+		return "", fmt.Errorf("ensuring DB instance %s exists in cluster %s: %v", instanceID, clusterID, err)
```
SimonBaeumer, line 63:
```suggestion
		return "", errors.Wrapf(err, "ensuring DB instance %s exists in cluster %s", instanceID, clusterID)
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// EnsureDBProvisioned is a blocking function that makes sure that an RDS database was provisioned for a Central
+func (c *RDSProvisioningClient) EnsureDBProvisioned(ctx context.Context, centralNamespace, centralDbSecretName string) (string, error) {
+	clusterID := centralNamespace + dbClusterSuffix
+	instanceID := centralNamespace + dbInstanceSuffix
+
+	err := c.ensureDBClusterCreated(ctx, clusterID, centralNamespace, centralDbSecretName)
+	if err != nil {
+		return "", fmt.Errorf("ensuring DB cluster %s exists: %v", clusterID, err)
```
SimonBaeumer, line 58:
```suggestion
		return "", errors.Wrapf(err, "ensuring DB cluster %s exists", clusterID)
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+	// SessionToken is a token required for temporary security credentials retrieved via STS.
+	SessionToken string
+}
+
+// EnsureDBProvisioned is a blocking function that makes sure that an RDS database was provisioned for a Central
+func (c *RDSProvisioningClient) EnsureDBProvisioned(ctx context.Context, centralNamespace, centralDbSecretName string) (string, error) {
+	clusterID := centralNamespace + dbClusterSuffix
+	instanceID := centralNamespace + dbInstanceSuffix
+
+	err := c.ensureDBClusterCreated(ctx, clusterID, centralNamespace, centralDbSecretName)
```
SimonBaeumer, line 56: You can move the err and ensure call into the if-statement directly.

```
if err := c.ensureDBClusterCreated(ctx, clusterID, centralNamespace, centralDbSecretName); err != nil {
    // ...
}
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+	if err != nil {
+		return "", fmt.Errorf("ensuring DB cluster %s exists: %v", clusterID, err)
+	}
+
+	err = c.ensureDBInstanceCreated(instanceID, clusterID)
```
SimonBaeumer, line 61: Can be moved the if case directly.

```
if err := c.ensureDBInstanceCreated(instanceID, clusterID); err != nil {
    // ....
}
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func (c *RDSProvisioningClient) clusterStatus(clusterID string) (string, error) {
+	dbClusterQuery := &rds.DescribeDBClustersInput{
+		DBClusterIdentifier: aws.String(clusterID),
+	}
+
+	clusterResult, err := c.rdsClient.DescribeDBClusters(dbClusterQuery)
+	if err != nil {
+		return "", fmt.Errorf("getting cluster status: %v", err)
```
SimonBaeumer, line 164:
```suggestion
		return "", errors.Wrap(err, "getting cluster status")
```

fleetshard/pkg/central/dbprovisioning/dbprovisioningclient.go
```diff
@@ -0,0 +1,12 @@
+// Package dbprovisioning provides functionality to provision and deprovision databases on cloud providers
+package dbprovisioning
```
SimonBaeumer, line 2: The naming is duplicate of the package name. To me the package would look something like this:

```
cloudprovider/
  dbclient.go         // has interface
  awsclient/
    rds.go
```

`dbprovisioning` sounds overly specific to me which would make all other queries related to instance management not being located here.

fleetshard/pkg/central/dbprovisioning/dbprovisioningclient.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+)
+
+// Client defines an interface for clients that can provision and deprovision databases on cloud providers
+type Client interface {
+	EnsureDBProvisioned(ctx context.Context, name, passwordSecretName string) (string, error)
```
SimonBaeumer, line 10: Docs should be on the interface, i.e. if the call is blocking.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -96,6 +95,68 @@ func TestReconcileCreate(t *testing.T) {
...
+			ManagedDBEnabled: true})
+
+	status, err := r.Reconcile(context.TODO(), simpleManagedCentral)
+	require.NoError(t, err)
+	managedDBProvisioningClient.AssertNumberOfCalls(t, "EnsureDBProvisioned", 1)
+
+	if readyCondition, ok := conditionForType(status.Conditions, conditionTypeReady); ok {
+		assert.Equal(t, "True", readyCondition.Status)
+	} else {
+		assert.Fail(t, "Ready condition not found in conditions", status.Conditions)
```
SimonBaeumer, line 115: Why not printing a message in `assert.Equal` if the assert failed? There is no need to call `assert.Fail` manually here

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -96,6 +95,68 @@ func TestReconcileCreate(t *testing.T) {
...
+	r := NewCentralReconciler(fakeClient, private.ManagedCentral{}, managedDBProvisioningClient,
+		CentralReconcilerOptions{
+			UseRoutes:        true,
+			ManagedDBEnabled: true})
+
+	status, err := r.Reconcile(context.TODO(), simpleManagedCentral)
+	require.NoError(t, err)
+	managedDBProvisioningClient.AssertNumberOfCalls(t, "EnsureDBProvisioned", 1)
+
+	if readyCondition, ok := conditionForType(status.Conditions, conditionTypeReady); ok {
```
SimonBaeumer, line 112:
```suggestion
readyCondition, ok := conditionForType(status.Conditions, conditionTypeReady)
require.True(t, ok)
assert.Equal(t, "True", readyCondition.Status, "Ready condition not found in conditions", status.Conditions)
```

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -96,6 +95,68 @@ func TestReconcileCreate(t *testing.T) {
...
+		assert.Equal(t, "True", readyCondition.Status)
+	} else {
+		assert.Fail(t, "Ready condition not found in conditions", status.Conditions)
+	}
+
+	central := &v1alpha1.Central{}
+	err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: centralName, Namespace: centralNamespace}, central)
+	require.NoError(t, err)
+	assert.Equal(t, centralName, central.GetName())
+	assert.Equal(t, "1", central.GetAnnotations()[revisionAnnotationKey])
```
SimonBaeumer, line 122: Why do you assert the revision? `Reconcile` is called once.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -96,6 +95,68 @@ func TestReconcileCreate(t *testing.T) {
...
+	assert.Equal(t, "true", central.GetAnnotations()[managedServicesAnnotation])
+	assert.Equal(t, true, *central.Spec.Central.Exposure.Route.Enabled)
+
+	route := &openshiftRouteV1.Route{}
+	err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: centralReencryptRouteName, Namespace: centralNamespace}, route)
+	require.NoError(t, err)
+	assert.Equal(t, centralReencryptRouteName, route.GetName())
+	assert.Equal(t, openshiftRouteV1.TLSTerminationReencrypt, route.Spec.TLS.Termination)
+	assert.Equal(t, testutils.CentralCA, route.Spec.TLS.DestinationCACertificate)
```
SimonBaeumer, line 131: Why verifying so many properties on the created route?

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -96,6 +95,68 @@ func TestReconcileCreate(t *testing.T) {
...
+
+	secret := &v1.Secret{}
+	err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: centralDbSecretName, Namespace: centralNamespace}, secret)
+	require.NoError(t, err)
+	_, ok := secret.Data["password"]
```
SimonBaeumer, line 136: I think an assertion is missing that the DB password is not empty.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -231,6 +292,54 @@ func TestReconcileDelete(t *testing.T) {
 	assert.True(t, k8sErrors.IsNotFound(err))
 }
 
+func TestReconcileDeleteWithManagedDB(t *testing.T) {
+	fakeClient := testutils.NewFakeClientBuilder(t).Build()
+
+	managedDBProvisioningClient := new(testutils.DBProvisioningClientMock)
+	managedDBProvisioningClient.On("EnsureDBProvisioned", mock.Anything, mock.Anything, centralDbSecretName).Return("connectionString", nil)
```
SimonBaeumer, line 299: Can you add a meaningful example of a connection string with example data?

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -231,6 +292,54 @@ func TestReconcileDelete(t *testing.T) {
...
+	managedDBProvisioningClient.On("EnsureDBDeprovisioned", mock.Anything).Return(true, nil)
+	statusTrigger, err := r.Reconcile(context.TODO(), deletedCentral)
+	require.Error(t, err, ErrDeletionInProgress)
+	require.Nil(t, statusTrigger)
+	managedDBProvisioningClient.AssertNumberOfCalls(t, "EnsureDBDeprovisioned", 1)
+
+	// deletion completed needs second reconcile to check as deletion is async in a kubernetes cluster
+	statusDeletion, err := r.Reconcile(context.TODO(), deletedCentral)
+	require.NoError(t, err)
+	require.NotNil(t, statusDeletion)
```
SimonBaeumer, line 322: Should this be a check on a specific status?

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -422,6 +463,61 @@ func (r *CentralReconciler) ensureNamespaceDeleted(ctx context.Context, name str
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      centralDbSecretName,
+					Namespace: remoteCentralNamespace,
+				},
+				Data: map[string][]byte{"password": []byte(generatedPassword)},
+			}
+
+			err = r.client.Create(ctx, secret)
+			if err != nil {
+				return fmt.Errorf("creating Central DB secret: %v", err)
```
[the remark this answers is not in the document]
vladbologa: I replaced %v with %w, because it's the preferred way in Go 1.13+ here and in all other places.

fleetshard/pkg/central/dbprovisioning/dbprovisioningclient.go
vladbologa: Done, I think your naming suggestion is much better.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: I checked the docs and it's not possible to get more than 1 instances if you search by name (there must be an exact match). And if the instance is not found, the `DescribeDBInstances` will fail. So I removed this check entirely.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: I think it's useful to see exactly when we called the AWS API to provision / deprovision databases and instances.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Agreed, I removed this comment.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+
+			connectionString := fmt.Sprintf("host=%s port=%d user=%s dbname=%s sslmode=require",
+				*clusterResult.DBClusters[0].Endpoint, 5432, dbUser, "postgres")
+
+			return connectionString, nil
+		}
+
+		glog.Infof("RDS instance status: %s", dbInstanceStatus)
+		time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
vladbologa: The time is 8-10 minutes... With a 10 sec delay we're making about 50 requests to AWS.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: `DescribeDBInstances` / `DescribeDBClusters` belong to the `aws-sdk-go` package. I could make a wrapper function, to make the code easier to follow.

fleetshard/pkg/central/dbprovisioning/dbprovisioningclient.go
vladbologa: Done

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Fixed.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Tbh not sure, will have to check. I think it should be a requirement, otherwise if a Central is deleted by mistake, it will not be possible to recover it to its most recent state.

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: These checks were copied over from the `TestReconcileCreate` test.

I thought I should do all the checks from that test (+ the DB checks), to make sure there were no side-effects (i.e. causing unrelated features to stop working).

kovayur (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -422,6 +463,61 @@ func (r *CentralReconciler) ensureNamespaceDeleted(ctx context.Context, name str
...
+
+	generatedPassword, err := random.GenerateString(25, random.AlphanumericCharacters)
+	if err != nil {
+		return fmt.Errorf("generating Central DB password: %v", err)
+	}
+
+	err = r.client.Get(ctx, ctrlClient.ObjectKey{Namespace: remoteCentralNamespace, Name: centralDbSecretName}, secret)
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			secret = &corev1.Secret{
```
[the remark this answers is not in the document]
vladbologa: Done

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: You're right, fixed.

I got tricked by the function doc, which only lists `ErrCodeDBClusterNotFoundFault` as possible error codes, but that doesn't account for network related errors etc.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: I'd leave it for another PR, because we only support this region anyway atm. If I don't hardcode it here, I'd have to hardcode it somewhere in the upper layers. Should I do that?

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: I agree, this was a hack from the original demo that somehow stayed, but now it's not easy to get rid of.

I have now refactored the RDS client to remove references to namespaces (I was using them to name the clusters & instances, but also to retrieve the password secret - functionality which does not belong here). 

But then I noticed that we don't have the `instance-id` anywhere in fleetshard, except for the namespace. Possible solutions:

1.  keep using the namespace for now, create a ticket to remove coupling to namespace (I suppose we can assume the namespaces will be stable short-medium term)
2. propagate the `CentralID` from fleet-manager to fleetshard (requires internal API changes)
3. generate the DB name and store it (in etcd?) I would prefer not to do this, I want to avoid keeping state, and instead having deterministic names for the DB.

WDYT? I'd go for 1, followed by 2.

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: This code was part of the original `TestReconcileDelete`, which I duplicated and added the DB-related checks. I think you wrote these asserts. :)

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: I did not write this code. There is the original test `TestReconcileCreate`, I added `TestReconcileCreateWithManagedDB` which is similar to the original one, but uses the Managed DB and has some extra checks. Same goes for several other comments here.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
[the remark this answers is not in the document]
SimonBaeumer: Can we than use 1 min? 🤔 
If the provisioning misses the first attempt it needs at leat 20 Minutes which would result in a timeout and a pager notification.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
SimonBaeumer: I think the check should not removed than, it is always possible to have an error in AWS SDK. If Go tries to access an invalid index the application crashes.
See: https://go.dev/play/p/-IfKZQJVojL

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
SimonBaeumer: Initializing the struct in the function call is what I've tried to describe. That would be independent of the aws sdk, the caller has full control over it:

```
_, err := c.rdsClient.DescribeDBClusters(&rds.DescribeDBClustersInput{
    DBClusterIdentifier: aws.String(clusterID),
})
```

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
SimonBaeumer: It is about the `dbInstanceQuery`. Sorry my comment was not explicit enough.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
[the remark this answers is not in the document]
vladbologa: I'm not sure I understand. This is just the time interval between two AWS calls. Additionally there's a 15 minute timeout  for the `Reconcile` function.

What is it that causes the Pager notification? If there's no status reported from fleetshard -> Fleet Manager in 20 minutes? Or if there's no running Central in 20 minutes?

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
SimonBaeumer: > propagate the CentralID from fleet-manager to fleetshard (requires internal API changes)

Option 2 sounds the best to me. Imho follow-up PR is okay, a TODO with the ticket number to it is sufficient.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Ah ok, I actually made this change. See rds.go:197 (describeDBInstance) and rds.go:209(describeDBCluster)

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
SimonBaeumer: Please remove the TODO if it is not necessary. We should not keep the database after deletion, so I suspect that this is a requirement.
Can you check with Doron?

fleetshard/pkg/central/reconciler/reconciler_test.go
SimonBaeumer: Interesting, no need to tackle it in this PR.

fleetshard/pkg/central/reconciler/reconciler_test.go
SimonBaeumer: We also have e2e tests. No strong feelings on having too many assertions here, just wanted to be cautiousness about them because it creates unnecessary noise in CI logs.
Maybe reducing the amount of assertions is fine here? Do you expect any side effects except the route not being created? 🤔

fleetshard/pkg/central/reconciler/reconciler_test.go
SimonBaeumer: Okay, then we can do better than my old me here 🥳 
Please let us not use "copied the code" as an excuse to keep logic which we can do simply better now. Also this means the author of the copied code did not fully understood the assertion itself as well. We are all stuck in the past😱

fleetshard/pkg/central/reconciler/reconciler_test.go
SimonBaeumer: > These checks were copied over from the TestReconcileCreate test.

As I read it again I have strong feelings. Copying code and leaving the state is not good if we know it already better. Let's spent the time on housekeeping. 🧹

vladbologa: I don't expect any sides effects, but I don't want to test only for things that I expect. There might be side effects in the future for who knows what reason.

I think the test for "Central creation" should be just as thorough as the test for "Central creation with Managed DB". I don't think that duplicated code in tests is as bad as in production code, but we could also factor some things out.

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: I agree, but I couldn't resist pointing it out, sorry. 😂

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: We don't keep the database. It's a snapshot stored on S3, with a retention period of 30 days (set with BackupRetentionPeriod).

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Added a Jira ticket in the TODO.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done, added ticket number.

fleetshard/pkg/central/dbprovisioning/rdsprovisioningclient.go
vladbologa: Done

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: The revision value is set by the reconciler, so this checks that the value is correct.

dev/env/manifests/fleetshard-sync/02-fleetshard-sync-deployment.yaml
```diff
@@ -43,6 +43,18 @@ spec:
...
+            - name: MANAGED_DB_ENABLED
+              value: "$MANAGED_DB_ENABLED"
+            - name: MANAGED_DB_SECURITY_GROUP
+              value: "$MANAGED_DB_SECURITY_GROUP"
+            - name: MANAGED_DB_SUBNET_GROUP
+              value: "$MANAGED_DB_SUBNET_GROUP"
+            - name: MANAGED_DB_ACCESS_KEY_ID
+              value: "$MANAGED_DB_ACCESS_KEY_ID"
+            - name: MANAGED_DB_SECRET_ACCESS_KEY
+              value: "$MANAGED_DB_SECRET_ACCESS_KEY"
```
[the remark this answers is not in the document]
vladbologa: Done.

dev/env/manifests/fleetshard-sync/02-fleetshard-sync-deployment.yaml
```diff
@@ -43,6 +43,18 @@ spec:
...
+            - name: MANAGED_DB_SESSION_TOKEN
+              value: "$MANAGED_DB_SESSION_TOKEN"
```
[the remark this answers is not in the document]
vladbologa: Done.

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: Actually the status is checked the following lines...

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: Done

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: Done

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
```diff
@@ -0,0 +1,340 @@
...
+	result, err := r.rdsClient.DescribeDBInstances(
+		&rds.DescribeDBInstancesInput{
+			DBInstanceIdentifier: aws.String(instanceID),
+		})
+	if err != nil {
+		return nil, fmt.Errorf("retrieving DB instance state: %w", err)
+	}
+
+	if len(result.DBInstances) != 1 {
+		// this should never happen (DescribeDBInstances should return either 1 instance, or ErrCodeDBInstanceNotFoundFault)
```
SimonBaeumer, line 208: +1000, defensive programming!

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -94,6 +125,35 @@ func TestReconcileCreate(t *testing.T) {
 	assert.Equal(t, centralReencryptRouteName, route.GetName())
 	assert.Equal(t, openshiftRouteV1.TLSTerminationReencrypt, route.Spec.TLS.Termination)
 	assert.Equal(t, testutils.CentralCA, route.Spec.TLS.DestinationCACertificate)
+
+	secret := &v1.Secret{}
+	err = fakeClient.Get(context.TODO(), client.ObjectKey{Name: centralDbSecretName, Namespace: centralNamespace}, secret)
+	require.NoError(t, err)
+	password, ok := secret.Data["password"]
+	require.True(t, ok)
+	assert.True(t, len(password) > 0)
```
SimonBaeumer, line 134: There are pre-defined existing assertion (asset.NotEmpty could work too):

```suggestion
	assert.Len(t, len(password) > 0)
```

SimonBaeumer (changes requested):
Good job @vladbologa!

SimonBaeumer (approved):

ebensh (approved):
LGTM, let's get it in and test it out further

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: `Len` doesn't work in this case (it compares against an exact value), I changed to `NotEmpty`

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -96,6 +93,62 @@ func TestReconcileCreate(t *testing.T) {
...
+		})
+	require.NoError(t, err)
+
+	r := NewCentralReconciler(fakeClient, private.ManagedCentral{}, managedDBProvisioningClient,
+		CentralReconcilerOptions{
+			UseRoutes:        true,
+			ManagedDBEnabled: true})
+
+	_, err = r.Reconcile(context.TODO(), simpleManagedCentral)
+	require.ErrorContains(t, err, "InvalidClientTokenId")
```
kovayur, line 149: nit: Shouldn't we test here with AWS mocked services instead of real?

fleetshard/pkg/testutils/dbclient.go
```diff
@@ -0,0 +1,24 @@
+package testutils
+
+import (
+	"context"
+
+	"github.com/stretchr/testify/mock"
+)
+
+// DBProvisioningClientMock is a mock cloudprovider.DBClient
+type DBProvisioningClientMock struct {
```
kovayur, line 10: nit: shouldn't we use `moq` to generate mocked implementations?

kovayur (approved):
LGTM. Just a few small remarks on unit testing. Can be fixed in the following PRs.

kovayur:
Good job @vladbologa 🚀 !

fleetshard/pkg/central/reconciler/reconciler_test.go
vladbologa: actually I used the real one on purpose for this test, but maybe it's not good practice if a unit test depends on external services

PR #554, opened 2022-11-07
ROX-12380: Run the test step too.

porridge:
## Description

Run the toy step from before #541 to see what actually broke.


porridge:
That turned out to be something else.

PR #555, opened 2022-11-07
ROX-13004: Update Stackrox reference in go.mod

vladbologa:
## Description

Updating Stackrox so that we have access to new fields in the Central specs.

This PR also fixes a failing test, that started to fail after the Stackrox sha update.
The test expects a DB transaction to fail, and sets this expectation with `mocket.Catcher.Reset().NewMock().WithQuery("INSERT").WithExecException()`
This no longer works, as Gorm executes the insert operation without returning an error. However I noticed that if I set this expectation in a different way, the test works again. (I used `WithError` instead of `WithExecException`).

I am assuming that the bug is caused by the update of the gorm library. Perhaps there is an incompatibility between the new `gorm` version, and the `mocket` library that is used by these tests (this library was last updated 4 years ago). The function that should return an error is part of the Gorm library (see `internal/dinosaur/pkg/services/clusters.go:126`), so the wrong behaviour in this test does not seem to be related to our code, only to external libraries.

~~- [ ] Evaluated and added CHANGELOG.md entry if required~~

## Test manual

```
# To run tests locally run:
```


internal/dinosaur/pkg/services/clusters_test.go
```diff
@@ -151,7 +151,7 @@ func Test_Cluster_Create(t *testing.T) {
 				cluster: buildCluster(nil),
 			},
 			setupFn: func() {
-				mocket.Catcher.Reset().NewMock().WithQuery("INSERT").WithExecException()
+				mocket.Catcher.Reset().NewMock().WithQuery("INSERT").WithError(fmt.Errorf("database error"))
```
vladbologa, line 154: This test was suddenly failing after changing the Stackrox referencer and running `go mod tidy`, and I assume it's related to the update of the `gorm` library. Not sure if I should be spending too much time debugging this, and this workaround works.

vladbologa: Seems that the bug is indeed causes by the gorm version bump, the test also fails in the dependabot commit:
https://github.com/stackrox/acs-fleet-manager/pull/551

go.mod
```diff
@@ -38,39 +38,39 @@ require (
...
 	golang.org/x/net v0.1.0
-	golang.org/x/oauth2 v0.0.0-20220718184931-c8730f7fcb92
+	golang.org/x/oauth2 v0.1.0
 	golang.org/x/sys v0.1.0
 	gopkg.in/resty.v1 v1.12.0
 	gopkg.in/yaml.v2 v2.4.0
-	gorm.io/driver/postgres v1.4.4
-	gorm.io/gorm v1.23.8
+	gorm.io/driver/postgres v1.4.5
+	gorm.io/gorm v1.24.1-0.20221019064659-5dd2bb482755
```
kovayur, line 62: Why not 1.24.1? Like in #551

vladbologa: it's like this in the Stackrox repo, and was added by `go mod tidy`. I see that if change it to 1.24.1 it stays like that, even after running `go mod tidy`, so I can change it to that.

kovayur (approved):

PR #561, opened 2022-11-07
Move cluster params to the Parameter Store

kovayur:
## Description
This PR removes wrongly defined `CLUSTER_ID`s. They are now loaded from the Parameter Store.

Related Slack channel: https://srox.slack.com/archives/C04AGPG4GMN

## Test manual
Diff between yamls

```
# To run tests locally run:
```

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
+++ b/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
@@ -64,13 +64,9 @@ function assert_environment() {
 case $CLUSTER_NAME in
 acs-stage-dp-01)
   assert_environment stage
-  CLUSTER_ID="d3ccd7a0-9b33-41e3-9024-6677f42259ed"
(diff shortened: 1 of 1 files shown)


porridge (approved):
Nice!
Please also update the `How to create an OSD Cluster` checklist in [environments page](https://docs.engineering.redhat.com/pages/viewpage.action?pageId=288990393) to mention the need to put these parameters in the store, apart from the need to update this script.

Some non-blocking thoughts, possibly food for a followup PR:
- we could now replace the `case/esac` piece to not be per-cluster but just take the second component of `$CLUSTER_NAME` and match it against `$ENVIRONMENT`. Then we could completely remove the need to update this script on cluster creation.
- would it be possible/easy to have `load_external_config` declare the variables as read-only so that it's harder to accidentally overwrite them in the script?
- might be worth adding a reference to the incident channel in the PR description

SimonBaeumer (approved):

PR #562, opened 2022-11-08
Rox 12076/probe service metrics [4]

stehessel:
This adds Prometheus metrics to the probe service. Metrics are recorded for

* number of runs started, failed and successful
* timestamps of last failed and successful runs. This will be used to define alerting rules upon failed runs.
* duration of probe runs.

## Test manual

curl `/metrics` endpoint.

```
# To run tests locally run:
```


probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+			value := targetMetric.Metric[0].GetCounter().GetValue()
+			assert.Equalf(t, expectedIncrement, value, "metric %s has unexpected value", tc.metricName)
+		})
+	}
+}
+
+func TestTimestampGauges(t *testing.T) {
+	tt := []struct {
+		metricName       string
+		setTimestampFunc func(m *Metrics)
```
SimonBaeumer, line 58:
```suggestion
		callSetTimestampFunc func(m *Metrics)
```

probe/pkg/metrics/server_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+func serveMetrics(t *testing.T, customMetrics *Metrics) metricResponse {
+	rec := httptest.NewRecorder()
+	req, err := http.NewRequest(http.MethodGet, "/metrics", nil)
+	require.NoError(t, err, "failed creating metrics requests")
+
+	server := newMetricsServer(":8081", customMetrics)
+	server.Handler.ServeHTTP(rec, req)
```
SimonBaeumer, line 50: Closing the server is missing?

probe/pkg/runtime/runtime.go
```diff
@@ -44,6 +45,17 @@ func (r *Runtime) RunLoop(ctx context.Context) error {
 
 // RunSingle executes a single probe run.
 func (r *Runtime) RunSingle(ctx context.Context) (errReturn error) {
+	metrics.MetricsInstance().IncStartedRuns()
+	defer func() {
+		if errReturn != nil {
```
SimonBaeumer, line 50: Please do not use named err returns variables in defer functions, this is not idiomatic and creates dependencies between function calls which can be hard to debug.

Please add the metrics calls after the probe run. I do not see here a benefit in calling this via defer. It is the same flow, except that you the structure of the Go code does not represent the natural execution path anymore only to return an error on a failed probe execution which could easily be solved with a function-local variable.

probe/cmd/probe/main.go
```diff
@@ -16,10 +19,23 @@ func main() {
...
+		glog.Info("unable to set logtostderr to true.")
 	}
 
-	c, err := cli.New()
+	config, err := config.GetConfig()
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	metricsServer, close := metrics.NewMetricsServer(config.MetricsAddress)
```
SimonBaeumer, line 30: `close` collides with built-in function name to close channels.

You can defer on the metricsServer. To guarantee nil-pointer safety a nil-check must be introduced anyway.

probe/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	prometheusNamespace = "acs"
+	prometheusSubsystem = "probe"
```
SimonBaeumer, line 13: Can you make these constants configurable?

probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCounterIncrements(t *testing.T) {
+	const expectedIncrement = 1.0
+
+	tt := []struct {
+		metricName        string
+		callIncrementFunc func(m *Metrics)
```
SimonBaeumer, line 17: +1 for naming

probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+			count := targetMetric.Metric[0].GetHistogram().GetSampleCount()
+			sum := targetMetric.Metric[0].GetHistogram().GetSampleSum()
+			assert.Equalf(t, expectedCount, count, "expected metric: %s to have a count of %v", tc.metricName, expectedCount)
+			assert.Equalf(t, expectedSum, sum, "expected metric: %s to have a sum of %v", tc.metricName, expectedSum)
+		})
+	}
+}
+
+func requireMetric(t *testing.T, metrics metricResponse, metricName string) *io_prometheus_client.MetricFamily {
+	targetMetric := metrics[metricName]
```
SimonBaeumer, line 124: You can use `assert.Contains`. To me this is more a `getMetric` func with some nil-checks, only that nil-checks in tests can be written as `required` which improves readability.

probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+			count := targetMetric.Metric[0].GetHistogram().GetSampleCount()
+			sum := targetMetric.Metric[0].GetHistogram().GetSampleSum()
+			assert.Equalf(t, expectedCount, count, "expected metric: %s to have a count of %v", tc.metricName, expectedCount)
+			assert.Equalf(t, expectedSum, sum, "expected metric: %s to have a sum of %v", tc.metricName, expectedSum)
+		})
+	}
+}
+
+func requireMetric(t *testing.T, metrics metricResponse, metricName string) *io_prometheus_client.MetricFamily {
```
SimonBaeumer, line 123:
```suggestion
func getMetric(t *testing.T, metrics metricResponse, metricName string) *io_prometheus_client.MetricFamily {
```

probe/pkg/metrics/server.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// NewMetricsServer returns the metrics server.
+func NewMetricsServer(address string) (*http.Server, func()) {
+	metricsServer := newMetricsServer(address, MetricsInstance())
+	closeFunc := func() {
+		if err := metricsServer.Close(); err != nil {
```
SimonBaeumer, line 15: Can you pass the metricsServer to the go func instead of accessing the variable directly?

probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+		t.Run(tc.metricName, func(t *testing.T) {
+			m := newMetrics()
+			expectedCount := uint64(2)
+			expectedSum := 480.0
+			tc.observeFunc(m)
+
+			metrics := serveMetrics(t, m)
+			targetMetric := requireMetric(t, metrics, tc.metricName)
+
+			count := targetMetric.Metric[0].GetHistogram().GetSampleCount()
```
SimonBaeumer, line 115: `require.NotEmpty(targetMetrics.Metric)`

SimonBaeumer (changes requested):

probe/pkg/metrics/metrics_test.go
stehessel: removed the function all together

probe/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+			Name:      "runs_failed",
+			Help:      "The number of failed probe runs.",
+		}),
+		lastSuccessTimestamp: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "last_success_timestamp",
+			Help:      "The Unix timestamp of the last successful probe run.",
+		}),
+		lastFailureTimestamp: prometheus.NewGauge(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
stehessel: done

probe/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,133 @@
...
+	m.lastFailureTimestamp.SetToCurrentTime()
+}
+
+// ObserveTotalDuration sets the total duration gauge for probe runs.
+func (m *Metrics) ObserveTotalDuration(duration time.Duration) {
+	m.totalDurationHistogram.Observe(duration.Seconds())
+}
+
+// MetricsInstance returns the global Singleton instance for Metrics.
+func MetricsInstance() *Metrics {
```
dhaus67, line 80: nit: Within `stackrox/stackrox` we named those functions `...Singleton()`, but since we don't have a strict guideline within this repo, you don't have to change it and can keep the name.

dhaus67, line 80: Does this one have to be exported? I'm not sure about potential other use-cases where we need it exported, since it's currently only called by the `NewMetricsServer`, but maybe there are?

probe/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,133 @@
...
+		startedRuns: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "runs_started",
+			Help:      "The number of started probe runs.",
+		}),
+		successfulRuns: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "runs_success",
```
dhaus67, line 98: nit: `runs_succeeded` to be in line with `runs_failed`?

probe/cmd/probe/main.go
```diff
@@ -16,10 +19,28 @@ func main() {
...
+	}
+
+	if metricsServer := metrics.NewMetricsServer(config.MetricsAddress); metricsServer != nil {
+		defer func(server *http.Server) {
+			if err := server.Close(); err != nil {
+				glog.Errorf("failed to close metrics server: %v", err)
+			}
+		}(metricsServer)
+		go func() {
+			if err := metricsServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
```
dhaus67, line 37: nit: Personally favor the `error.Is` call more, but looking at other sources and places this one's also valid, so you don't have to change it.
```suggestion
			if err := metricsServer.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
```

probe/cmd/probe/main.go
```diff
@@ -16,10 +19,28 @@ func main() {
...
+		glog.Info("unable to set logtostderr to true.")
 	}
 
-	c, err := cli.New()
+	config, err := config.GetConfig()
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	if metricsServer := metrics.NewMetricsServer(config.MetricsAddress); metricsServer != nil {
```
dhaus67, line 30: It'd be nice to factor out the go routine into a different function + the defer function.

probe/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Metrics holds the prometheus.Collector instances for the probe's custom metrics
+// and provides methods to interact with them.
+type Metrics struct {
+	startedRuns            prometheus.Counter
+	successfulRuns         prometheus.Counter
+	failedRuns             prometheus.Counter
+	lastStartedTimestamp   prometheus.Gauge
+	lastSuccessTimestamp   prometheus.Gauge
+	lastFailureTimestamp   prometheus.Gauge
```
dhaus67, line 29: From a naming point of view, it'd be nice if the properties would match the name of the metric, i.e. `failure != failed` to be consistent there.

probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCounterIncrements(t *testing.T) {
```
dhaus67, line 11: Maybe it'd be worthwhile to lint our metrics using https://github.com/prometheus/client_golang/blob/main/prometheus/testutil/lint.go#L26 ?

probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,132 @@
...
+			m := newMetrics()
+			tc.callIncrementFunc(m)
+
+			metrics := serveMetrics(t, m)
+			require.Contains(t, metrics, tc.metricName)
+			targetMetric := metrics[tc.metricName]
+
+			// Test that the metrics value is 1 after calling the incrementFunc.
+			require.NotEmpty(t, targetMetric.Metric)
+			value := targetMetric.Metric[0].GetCounter().GetValue()
```
dhaus67, line 50: It's a bit of a bummer to only get a `float64` from the counter instead of the integer you would expect 

sadge

probe/pkg/metrics/metrics.go
stehessel: I'd keep it like this to be consistent with https://github.com/stackrox/acs-fleet-manager/blob/main/fleetshard/pkg/fleetshardmetrics/metrics.go#L74.

stehessel: It's also called whenever we want to collect metrics. E.g. `metrics.MetricsInstance().IncStartedRuns()`. We could put the singleton behind the scenes, however that makes testing more difficult because we would then somehow need to reset the singleton after every test. In stackrox/stackrox we do this, but there we also just seem to skip the tests entirely so we don't have that problem.

probe/pkg/metrics/metrics_test.go
stehessel: https://www.youtube.com/watch?v=y9r_pZL4boE

probe/pkg/metrics/metrics_test.go
stehessel: Feel free to provide an example ;). Not sure how this method is supposed to be called.

dhaus67: So what I had in mind is using the metrics we defined within `newMetrics` and have all of them run through the linter. This way we would get some linting errors regarding i.e. naming issues.

Code would be something as such:
```go
func TestMetricConformity(t *testing.T) {
	metrics := newMetrics()

	for _, metric := range []prometheus.Collector{
		metrics.startedRuns, metrics.succeededRuns, metrics.failedRuns, metrics.lastStartedTimestamp,
		metrics.lastSuccessTimestamp, metrics.lastFailureTimestamp, metrics.totalDurationHistogram,
	} {

		problems, err := testutil.CollectAndLint(metric)
		assert.NoError(t, err)
		assert.Empty(t, problems)
	}
}

```

This code would yield for example problems for the counter metrics and their naming:
```
    metrics_test.go:68: 
        	Error:      	Should be empty, but was [{acs_probe_runs_started counter metrics should have "_total" suffix}]
        	Error:      	Should be empty, but was [{acs_probe_runs_succeeded counter metrics should have "_total" suffix}]
        	Error:      	Should be empty, but was [{acs_probe_runs_failed counter metrics should have "_total" suffix}]
```

dhaus67 (approved):

probe/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,120 @@
+// Package metrics ...
```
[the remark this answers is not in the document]
SimonBaeumer: I think this can be removed

probe/pkg/metrics/metrics.go
SimonBaeumer: My bad, I confused this with a kubernetes namespace 🤦

probe/pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+func TestMetricsConformity(t *testing.T) {
+	metrics := newMetrics()
+
+	for _, metric := range []prometheus.Collector{
+		metrics.startedRuns, metrics.succeededRuns, metrics.failedRuns, metrics.lastStartedTimestamp,
+		metrics.lastSuccessTimestamp, metrics.lastFailureTimestamp, metrics.totalDurationHistogram,
+	} {
+		problems, err := testutil.CollectAndLint(metric)
+		assert.NoError(t, err)
```
SimonBaeumer, line 144:
```suggestion
		require.NoError(t, err)
```

probe/cmd/probe/main.go
```diff
@@ -16,10 +18,20 @@ func main() {
...
+		defer metrics.CloseMetricsServer(metricsServer)
```
SimonBaeumer, line 30: The functions look a bit weird to me because errors are silently omitted.

1. Cleaner would be a `utils.Should` function similar to stackrox
2. imho if the metrics server errors the prove service should not start at all

SimonBaeumer (changes requested):

stehessel:
> I think this can be removed

You'd think wrong ;). We have a pretty aggressive linter that requires every package to be commented. But now I see that I accidentally edited the wrong `metrics.go` comment 🤦 .

probe/pkg/metrics/metrics_test.go
stehessel: Why `require` here? I think `assert` would be more fitting here because it's one of the primary assertions of the test and not a requirement for other assertions.

SimonBaeumer: You are right, no need to change

SimonBaeumer:
Can you adjust PR title to `ROX-12076: probe service metrics [4]`?

Good job on splitting the PRs @stehessel, I liked how I could do a structured review and provide meaningful feedback.

probe/cmd/probe/main.go
stehessel: They are not omitted silently but logged. But I can add the `utils.Should`.

SimonBaeumer (approved):

PR #563, opened 2022-11-08
probe service: use correct context when invoking the cleanup

stehessel:
This is a bug that sadly got introduced in my last iteration of #533. The cause is a simple typo (`ctx` vs `cleanupCtx`), however this leads to the cleanup not having its own context. While that would be more "golang native", the issue is that we actually want the a blocking clean up call in this case. Otherwise after a timeout or cancellation of the main context, the probe resources would not get cleaned up properly.

I used `NotContains` in the regression test because I didn't manage to find a clean way to introduce a typed error and propagate it correctly with the other wrapped errors of the main routine.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/probe/pkg/runtime/runtime.go
+++ b/probe/pkg/runtime/runtime.go
@@ -60,9 +60,9 @@ func (r *Runtime) RunSingle(ctx context.Context) (errReturn error) {
 			// If ONLY the clean up failed, the context error is wrapped and
 			// returned in `SingleRun`.
 			if errReturn != nil {
-				errReturn = errors.Wrapf(errReturn, "%s: %s", errCleanupFailed, cleanupCtx.Err())
+				errReturn = errors.Wrapf(errReturn, "%s: %s", errCleanupFailed, err)
 			} else {
-				errReturn = errors.Wrap(cleanupCtx.Err(), errCleanupFailed.Error())
+				errReturn = errors.Wrap(err, errCleanupFailed.Error())
 			}
 		}
 	}()
--- a/probe/pkg/runtime/runtime.go
+++ b/probe/pkg/runtime/runtime.go
@@ -11,6 +11,10 @@ import (
 	"github.com/stackrox/acs-fleet-manager/probe/pkg/probe"
 )
 
+var (
+	errCleanupFailed = errors.New("cleanup failed")
+)
+
 // Runtime orchestrates probe runs against fleet manager.
 type Runtime struct {
 	Config *config.Config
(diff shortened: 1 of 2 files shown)


probe/pkg/runtime/runtime.go
```diff
@@ -50,15 +54,15 @@ func (r *Runtime) RunSingle(ctx context.Context) (errReturn error) {
...
 			// If clean up failed AND the original probe run failed, wrap the
 			// original error and return it in `SingleRun`.
 			// If ONLY the clean up failed, the context error is wrapped and
 			// returned in `SingleRun`.
 			if errReturn != nil {
-				errReturn = errors.Wrapf(errReturn, "cleanup failed: %s", cleanupCtx.Err())
+				errReturn = errors.Wrapf(errReturn, "%s: %s", errCleanupFailed, cleanupCtx.Err())
 			} else {
-				errReturn = errors.Wrap(cleanupCtx.Err(), "cleanup failed")
+				errReturn = errors.Wrap(cleanupCtx.Err(), errCleanupFailed.Error())
```
dhaus67, line 65: Not as something you should change, but as how to make the error type checking work:
When you call `errors.Wrap`, a `errors.withMessage` will be created. This will have a `cause`, which is the error you want to wrap, and a `message`, which is whatever you passed as the additional argument (i.e. the `errCleanupFailed`).
When checking if an error is a part of the chain via `errors.Is / assert.ErrorIs`, each error in the chain will be checked and [Unwrap](https://pkg.go.dev/errors#example-Unwrap)  will be used to get the next error in the chain. Only the `cause` of the error will be checked during that, which explains why this doesn't work at the moment with regards to checking for the specific error type.
If you want to make it work, you have to wrap the `errCleanupFailed` instead of the `errReturn / cleanupCtx.Err()` respectively:
```go
				errReturn = errors.Wrapf(errCleanupFailed, "%s: %s", errReturn, cleanupCtx.Err())
			} else {
				errReturn = errors.Wrap(errCleanupFailed, cleanupCtx.Err().Error())
```

This is suboptimal to be honest - maybe with the error handling proposals for Go2 there will be improvement for this.

probe/pkg/runtime/runtime.go
```diff
@@ -50,15 +54,15 @@ func (r *Runtime) RunSingle(ctx context.Context) (errReturn error) {
...
 
-		if err := r.probe.CleanUp(ctx); err != nil {
+		if err := r.probe.CleanUp(cleanupCtx); err != nil {
 			// If clean up failed AND the original probe run failed, wrap the
 			// original error and return it in `SingleRun`.
 			// If ONLY the clean up failed, the context error is wrapped and
 			// returned in `SingleRun`.
 			if errReturn != nil {
-				errReturn = errors.Wrapf(errReturn, "cleanup failed: %s", cleanupCtx.Err())
+				errReturn = errors.Wrapf(errReturn, "%s: %s", errCleanupFailed, cleanupCtx.Err())
```
dhaus67, line 63: In case of running the tests currently, the rendered error will be `cleanup failed: %!s(<nil>): probe run failed: context canceled`.
Let's add a check to not print `cleanupCtx.Err()` if it's nil (you can't use `cleanupCtx.Err().Error()` since I'm pretty sure this will end up with a nil-pointer exception).

dhaus67 (approved):
Minor comments, otherwise LGTM.

PR #566, opened 2022-11-11
ROX-12076: docker image for probe service [5]

stehessel:
## Description

Add support for building and pushing Docker images for the probe service.
The images are pushed to `rhacs-eng:blackbox-monitoring-probe-service`
inside the stackrox quay.io organization.

In contrast to fleetmanager and fleetshard-sync, which are bundled together
in the same image, a new image for the probe is created. Consequently, the
relevant Makefile commands have been split into `/fleetmanager` and `/probe`
components.

In addition, the probe Dockerfile has been updated to reflect the latest
changes to the fleetmanager/fleetshard-sync Dockerfile.

## Test manual

Pushed image to repo via Makefile.

### DIFF
--- a/probe/Dockerfile
+++ b/probe/Dockerfile
@@ -21,7 +21,7 @@ ENV GOFLAGS=${GOFLAGS}
 
 RUN mkdir /src
 WORKDIR /src
-RUN CGO_ENABLED=0 go install -ldflags "-s -w -extldflags '-static'" github.com/go-delve/delve/cmd/dlv@latest
+RUN CGO_ENABLED=0 go install -ldflags "-s -w -extldflags '-static'" github.com/go-delve/delve/cmd/dlv@v1.9.1
 COPY go.*  ./
 RUN go mod download
 COPY . ./
@@ -32,14 +32,14 @@ RUN GOARGS="-gcflags 'all=-N -l'" make probe
 FROM build as build-standard
 RUN make probe
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6 as debug
+FROM registry.access.redhat.com/ubi8/ubi-micro:8.7 as debug
 COPY --from=build-debug /go/bin/dlv /src/probe/bin /stackrox/
 COPY --from=build-debug /src /src
 EXPOSE 7070
 ENTRYPOINT [ "/stackrox/dlv" , "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "exec", "/stackrox/probe"]
 CMD ["start"]
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6 as standard
+FROM registry.access.redhat.com/ubi8/ubi-micro:8.7 as standard
 COPY --from=build-standard /src/probe/bin /stackrox/
 EXPOSE 7070
 ENTRYPOINT ["/stackrox/probe"]
--- a/probe/Dockerfile
+++ b/probe/Dockerfile
@@ -32,14 +32,14 @@ RUN GOARGS="-gcflags 'all=-N -l'" make probe
 FROM build as build-standard
 RUN make probe
 
-FROM registry.access.redhat.com/ubi8/ubi-micro:8.7 as debug
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as debug
 COPY --from=build-debug /go/bin/dlv /src/probe/bin /stackrox/
 COPY --from=build-debug /src /src
 EXPOSE 7070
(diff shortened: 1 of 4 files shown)


stehessel:
* main
  * **PR #566** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/566" target="_blank">Graphite</a>  👈
    * **PR #567** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/567" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/566?utm_source=stack-comment).

probe/Dockerfile
```diff
@@ -1,13 +1,49 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
+FROM registry.access.redhat.com/ubi8/s2i-base:1-388 AS build
```
janisz, line 1: What is that?

probe/Dockerfile
```diff
@@ -1,13 +1,49 @@
...
+
+ARG GOROOT=/usr/local/go
+ENV GOROOT=${GOROOT}
+
+ARG GOFLAGS=-mod=mod
+ENV GOFLAGS=${GOFLAGS}
+
+RUN mkdir /src
+WORKDIR /src
+RUN CGO_ENABLED=0 go install -ldflags "-s -w -extldflags '-static'" github.com/go-delve/delve/cmd/dlv@latest
```
janisz, line 24: Don't we want to pin it? I'm fine with latest it might cause a problems in the future

probe/Dockerfile
```diff
@@ -1,13 +1,49 @@
...
+RUN go mod download
+COPY . ./
+
+FROM build as build-debug
+RUN GOARGS="-gcflags 'all=-N -l'" make probe
 
+FROM build as build-standard
+RUN make probe
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6 as debug
```
janisz, line 35: Can we use `ubi-micro`?

janisz, line 35:
```suggestion
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as debug
```

probe/Dockerfile
```diff
@@ -1,13 +1,49 @@
...
+COPY --from=build-debug /go/bin/dlv /src/probe/bin /stackrox/
+COPY --from=build-debug /src /src
+EXPOSE 7070
+ENTRYPOINT [ "/stackrox/dlv" , "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "exec", "/stackrox/probe"]
 CMD ["start"]
 
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6 as standard
```
janisz, line 42:
```suggestion
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as standard
```

How about configuring dependabot to maintain it?

probe/Dockerfile
stehessel: I copied the whole Dockerfile the "main" Dockerfile https://github.com/stackrox/acs-fleet-manager/blob/main/Dockerfile. I wanted the Dockerfiles to be consistent, but tbh for the probe we don't need anything complicated. I believe s2i-base is the universal Red Hat builder image https://catalog.redhat.com/software/containers/ubi8/s2i-base/5c83976a5a13464733ec6790.

stehessel: I checked, in this case the builder image is needed because of the `curl` and `tar` during the build.

probe/Dockerfile
stehessel: Yeah I think so

probe/Dockerfile
stehessel: We can do this as a follow up, I don't have permissions to configure dependabot in the rpo.

It would also be good for the "main" image. When it's set up for the main image we can just add it for the probe image as well.

janisz (approved):

probe/Dockerfile
janisz: https://github.com/stackrox/acs-fleet-manager/pull/576

stehessel:
@janisz I had to add two more commits based on learnings from the end2end test of the probe. Specifically, the `ubi-micro` sadly does not work because it's missing certificates. Switched back to `ubi-minimal`.

I also had to modify the `docker login` logic a bit to avoid overwriting logins. The plan is to use `quay.io/rhacs-eng/blackbox-monitoring-probe-service` as the registry. For the main image we use `quay.io/app-sre` registry - however, I don't think it's worth it to add the probe service to the app-interface build jobs (questionable if app-sre would even allow that...). So I will add the pull secrets for `quay.io/rhacs-eng` to the data plane cluster once we deploy the probe service.

dhaus67 (approved):

PR #567, opened 2022-11-11
ROX-12076: helm chart for probe service [7]

stehessel:
## Description

Add a probe service helm sub-chart to the data plane terraforming charts.

The reason for deploying the probe service to the data plane cluster is two-fold:

1. Easy deployment via Helm that is fully under our control. In particular, we do
   not have to deal with app-interface or get approvals from App-SRE.
2. Deploying the probe outside of the cluster it monitors (fleetmanager in App-SRE cluster)
   can be beneficial because presents more of an outside view. Turning up a new cluster just
   for the probe is costly and seems unnecessary.

Alternatively, we may also consider creating a separate helm chart for the probe service to avoid bloating the terraforming chart. Not sure what the helm best practice for this is. Opinions?

## Test manual

Tested Helm chart in dev cluster.


stehessel:
* main
  * **PR #566** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/566" target="_blank">Graphite</a> 
    * **PR #567** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/567" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/567?utm_source=stack-comment).

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -57,3 +57,12 @@ logging:
   aws:
     accessKeyId: ""
     secretAccessKey: ""
+
+# See available parameters in charts/probe/values.yaml
+# - enabled flag is used to completely enable/disable logging sub-chart
```
dhaus67, line 62:
```suggestion
# - enabled flag is used to completely enable/disable probe sub-chart
```

dp-terraform/helm/rhacs-terraform/Chart.yaml
```diff
@@ -29,3 +29,5 @@ dependencies:
     condition: observability.enabled
   - name: logging
     condition: logging.enabled
+  - name: probe
+    condition: probe.enabled
```
dhaus67, line 33: Irrespective of the overarching question of whether this should or shouldn't be a separate helm chart or sub chart:
will there be a follow-up to also create some sample script to deploy the probe service similar to the refactored `terraform_cluster.sh` incl. an appropriate GitHub action for it, or will this not be done at all and it will be "different"?
(I don't expect it to be done within this PR, would be a bit too much for sure)

dhaus67 (commented):
Personally, I wouldn't have seen the probe service as part of the "main" helm chart that is used for terraforming the clusters.
We already have the problem of different lifecycles within the existing chart due to the fleet manager & fleet shard sync. 
Also, if we were to "enable" the probe service on existing clusters, it would currently mean to add more within `terraform_cluster.sh`, which currently isn't the case.

I suppose leaving the chart separate and having it be applied in a GitHub action similar to the rhacs-terraform chart would be more worthwhile to me.

Something I'm missing currently would be the one-shot probe: should this be deployable via Helm chart as well, or do you have something different in mind of how people should use this?

dp-terraform/helm/rhacs-terraform/Chart.yaml
stehessel: Yeah, the hopefully last PR will be the actual deployment of the probe, i.e. either attached the existing terraforming or as a separate thing.

stehessel:
> I suppose leaving the chart separate and having it be applied in a GitHub action similar to the rhacs-terraform chart would be more worthwhile to me.

Yeah I'm fine with that.

stehessel:
> Something I'm missing currently would be the one-shot probe: should this be deployable via Helm chart as well, or do you have something different in mind of how people should use this?

I don't think the one-shot mode would be run on Kubernetes at all. It's meant for local execution of the probe, or within a CI run. As a follow up we can take a look at the current e2e and run the probe one-shot as a part of that, or replace some parts with it. But tbh I would see this as a separate story even.

dhaus67 (approved):
>I don't think the one-shot mode would be run on Kubernetes at all. It's meant for local execution of the probe, or within a CI run. As a follow up we can take a look at the current e2e and run the probe one-shot as a part of that, or replace some parts with it. But tbh I would see this as a separate story even.

Agreed that it should be a different story 👍 

Other than that, LGTM 🚢

stehessel:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/567?utm_source=stack-comment).


Job Reference: eWMyi9y2YiHkAp5CNKzt

PR #568, opened 2022-11-13
ROX-12076: Remove probe name prefix [6]

stehessel:
## Description

I did not realize that the HOSTNAME env variable inside the pod already includes `probe-`. An additional prefix is therefore not necessary.

## Checklist (Definition of Done)

- [x] Unit and integration tests added
- [ ] ~Added test description under `Test manual`~
- [ ] ~Evaluated and added CHANGELOG.md entry if required~
- [ ] ~Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~
- [x] CI and all relevant tests are passing
- [x] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [ ] ~Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.~

PR #582, opened 2022-11-17
ROX-12076: OCM auth [8]

stehessel:
## Description

Based on the learnings from e2e testing the probe on a dev cluster, I had to do some adjustments:
* Add OCM auth as a workaround. The reason behind this is that OCM does not seem to correctly support
  RHSSO service accounts when they represent the user instead of the calling service.
* I also added some minor adjustments:
  * Reduced clean up timeout to 5 minutes.
  * Catch SIGTERM signal.
  * Don't wait for deprovision in clean up.

## Test manual

Tested on dev cluster.

### DIFF
--- a/deploy/helm/probe/values.yaml
+++ b/deploy/helm/probe/values.yaml
@@ -1,9 +1,12 @@
 image: "quay.io/rhacs-eng/blackbox-monitoring-probe-service:main"
 namespace: "rhacs-probe"
 fleetManagerEndpoint: ""
-authType: ""
+# Must be either 'RHSSO' or 'OCM'.
+authType: "RHSSO"
 ocm:
+  # The username of the Red Hat SSO account.
   username: ""
+  # The refresh token obtained from `ocm token --refresh`.
   token: ""
 redHatSSO:
   clientId: ""
--- a/deploy/helm/probe/values.yaml
+++ b/deploy/helm/probe/values.yaml
@@ -1,6 +1,10 @@
 image: "quay.io/rhacs-eng/blackbox-monitoring-probe-service:main"
 namespace: "rhacs-probe"
 fleetManagerEndpoint: ""
+authType: ""
+ocm:
+  username: ""
+  token: ""
 redHatSSO:
   clientId: ""
   clientSecret: ""
--- a/probe/config/config.go
+++ b/probe/config/config.go
@@ -19,11 +19,7 @@ type Config struct {
 	FleetManagerEndpoint    string        `env:"FLEET_MANAGER_ENDPOINT" envDefault:"http://127.0.0.1:8000"`
 	MetricsAddress          string        `env:"METRICS_ADDRESS" envDefault:":7070"`
 	RHSSOClientID           string        `env:"RHSSO_SERVICE_ACCOUNT_CLIENT_ID"`
-	RHSSOClientSecret       string        `env:"RHSSO_SERVICE_ACCOUNT_CLIENT_SECRET"`
-	RHSSOEndpoint           string        `env:"RHSSO_ENDPOINT" envDefault:"https://sso.redhat.com"`
-	RHSSORealm              string        `env:"RHSSO_REALM" envDefault:"redhat-external"`
 	OCMUsername             string        `env:"OCM_USERNAME"`
-	OCMRefreshToken         string        `env:"OCM_TOKEN"`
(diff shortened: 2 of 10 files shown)


stehessel:
* main
  * **PR #566** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/566" target="_blank">Graphite</a> 
    * **PR #568** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/568" target="_blank">Graphite</a> 
      * **PR #567** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/567" target="_blank">Graphite</a> 
        * **PR #582** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/582" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/582?utm_source=stack-comment).

probe/config/config.go
```diff
@@ -38,11 +44,24 @@ func GetConfig() (*Config, error) {
...
+		}
+		c.ProbeUsername = fmt.Sprintf("service-account-%s", c.RHSSOClientID)
+	} else if c.AuthType == "OCM" {
+		if c.OCMUsername == "" {
+			configErrors.AddError(errors.New("OCM_USERNAME unset in the environment"))
+		}
+		if c.AuthType == "OCM" && c.OCMRefreshToken == "" {
+			configErrors.AddError(errors.New("OCM_TOKEN unset in the environment"))
+		}
+		c.ProbeUsername = c.OCMUsername
```
dhaus67, line 62: For the OCM username, it's a bit unfortunate that we require setting this statically instead of fetching the user name dynamically based on the token we got.
It'd be possible by using the following code:
```go
	l, err := sdk.NewGlogLoggerBuilder().Build()
	if err != nil {
		return nil, err
	}

	builder := sdk.NewConnectionBuilder().
		Client("cloud-services", "").
		Tokens(c.OCMRefreshToken).
		Logger(l)
	conn, err := builder.Build()
	if err != nil {
		return nil, err
	}
	resp, err := conn.AccountsMgmt().V1().CurrentAccount().Get().Send()
	if err != nil {
		return err
	}
	acc := resp.Body()
	acc.Username()
```

but this would mean duping parts of the auth code in `pkg/client/fleetmanager/auth_ocm.go`.

If you want to spend this extra effort of minimizing the required config for the OCM auth type, you can go for it. But, since the overall goal is to use RH SSO service accounts here and OCM is just a workaround, don't feel obliged to do it.

Just wanted to mention that it's also possible to retrieve this dynamically.

probe/pkg/probe/probe.go
```diff
@@ -162,11 +161,6 @@ func (p *ProbeImpl) deleteCentral(ctx context.Context, centralRequest *public.Ce
 		return errors.Wrapf(err, "deletion of central instance %s failed", centralRequest.Id)
 	}
 
-	_, err = p.ensureCentralState(ctx, centralRequest, constants.CentralRequestStatusDeprovision.String())
```
dhaus67, line 165: Just thinking about this and one of the issues we have observed regarding instances being stuck in deprovision status: do you think it makes sense to add the request status within the error message in l.172?
Meaning:
```
return errors.Wrapf(err, "central instance %s with status %s could not be deleted", centralRequest.Id, centralRequest.Status)
```

probe/config/config.go
stehessel: I think I would prefer to keep the OCM workaround as simple as possible.

deploy/helm/probe/values.yaml
```diff
@@ -1,6 +1,10 @@
 image: "quay.io/rhacs-eng/blackbox-monitoring-probe-service:main"
 namespace: "rhacs-probe"
 fleetManagerEndpoint: ""
+authType: ""
```
dhaus67, line 4: nit: should we set it specifically to `RHSSO` by default + add a comment that you currently only support either `RHSSO` or `OCM`?

dhaus67 (approved):
Minor nits otherwise LGTM 🚢

stehessel:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/582?utm_source=stack-comment).


Job Reference: eWMyi9y2YiHkAp5CNKzt

PR #584, opened 2022-11-18
Simplify the script a little.

porridge:
## Description

Simplify the script and elliminate the need to update it with every new cluster creation.

## Checklist (Definition of Done)

- [x] ~Unit and integration tests added~
- [x] Added test description under `Test manual`
- [x] ~Evaluated and added CHANGELOG.md entry if required~
- [x] ~Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~
- [ ] CI and all relevant tests are passing
- [x] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [x] ~Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.~

## Test manual

Ran the script with `echo` prepended to the `helm` invocation to make sure it runs and fails as expected.

PR #585, opened 2022-11-18
ROX-12076: Add continuous deployment for probe [9]

stehessel:
## Description

Add a deployment script to execute the Helm chart for the probe service. The stage deployment is done via CD in the stage deployment workflow. The pull secret is injected as a secret into the probe namespace and linked to a probe service account. The secret itself is stored in the AWS parameter store.

## Test manual

Tested the pull secret injection on dev cluster.

### DIFF
--- a/.github/workflows/deploy-stage.yaml
+++ b/.github/workflows/deploy-stage.yaml
@@ -9,9 +9,6 @@ jobs:
   cancel:
     name: Cancel previous runs
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write
-      contents: read
     steps:
       - name: Cancel Previous Runs
         uses: n1hility/cancel-previous-runs@v2
@@ -22,6 +19,9 @@ jobs:
     name: Re-terraform stage clusters
     needs: cancel
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
     environment: stage
     steps:
       - name: Set up Go 1.18
@@ -47,6 +47,9 @@ jobs:
     name: Deploy blackbox monitoring probe service
     needs: cancel
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
     environment: stage
     steps:
       - name: Set up Go 1.18
--- a/.github/workflows/deploy-stage.yaml
+++ b/.github/workflows/deploy-stage.yaml
@@ -6,18 +6,24 @@ on:
       - main
 
 jobs:
-  deploy:
-    name: "Re-terraform stage clusters"
+  cancel:
+    name: Cancel previous runs
     runs-on: ubuntu-latest
     permissions:
       id-token: write
       contents: read
-    environment: stage
     steps:
       - name: Cancel Previous Runs
         uses: n1hility/cancel-previous-runs@v2
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-stage.yaml
```diff
@@ -6,18 +6,24 @@ on:
...
 
 jobs:
-  deploy:
-    name: "Re-terraform stage clusters"
+  cancel:
+    name: Cancel previous runs
     runs-on: ubuntu-latest
     permissions:
       id-token: write
       contents: read
```
porridge, line 14: These permissions are needed to get a token for talking to AWS parameter store.
Now that there is a separate `terraform` step, I think these will either need to be moved there (and copied to the  `deploy-probe` step too), or moved up one level (if that makes sense w.r.t. the GH Actions workflow schema).

deploy/helm/probe/deploy.sh
```diff
@@ -0,0 +1,90 @@
...
+    PROBE_IMAGE="quay.io/rhacs-eng/blackbox-monitoring-probe-service:2b0c84d"
+    ;;
+
+  *)
+    echo "Unknown environment ${ENVIRONMENT}"
+    exit 2
+    ;;
+esac
+
+function assert_environment() {
```
porridge, line 42: Can I ask you to approve https://github.com/stackrox/acs-fleet-manager/pull/584/files and do the same here?

porridge (changes requested):
I was going to ask whether we have instructions for updating the pull secret (since apparently the base64-encoded blob needs to go into parameter store?) but then I started wondering whether we could somehow get rid of the need for the pull secret altogether, since the code is public anyway, and other images we deploy on the data plane clusters tend to be public as well, right?
That would be one less thing to worry about.

Also a couple of comments inline.

.github/workflows/deploy-stage.yaml
stehessel: Ah, somehow I thought these are only needed by the `cancel` step.

deploy/helm/probe/deploy.sh
stehessel: Sure

stehessel:
The other images are not public. But we pull them from `quay.io/app-sre` instead of `quay.io/rhacs-eng`, for which there seems to be already a pull secret by default on the cluster. However, I think making the probe image public is a good idea. I'll ask someone with quay admin credentials if this can be done.

kurlov:
BTW, we can push images under quay.io/app-sre org in app-interface (e.g. [public fleet manager image](https://console.redhat.com/openshift/token/show#)). There are already creds for that in JenkinsCI runners. We can introduce probe image build and push as a [dedicated step](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/acs-fleet-manager/cicd/jobs.yaml)

porridge:
> The other images are not public. But we pull them from `quay.io/app-sre` instead of `quay.io/rhacs-eng`, for which there seems to be already a pull secret by default on the cluster.

Hm, I thought these were public too, at least https://quay.io/repository/app-sre/acs-fleet-manager?tab=tags opens just fine in a fresh incognito window. :thinking:

porridge (approved):

PR #599, opened 2022-11-23
Drop temporary build and deploy script

kurlov:
## Description

[app-interface PR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/52332) is merged so `build_deploy.sh` script is no longer needed because `build_push_fleet_manager.sh` will be used instead


porridge (approved):
Wow, that was really temporary :-D

PR #605, opened 2022-11-24
ROX-12411: Make staging e2e use probe service

kurlov:
## Description

Make app-interface staging e2e tests run on top of the probe service.
App-interface e2e runs against real stage environment. This test is an upstream for build and deployment steps. So app-interface e2e runs right after successful stage deployment. 

Just for the context, OpenShift CI e2e builds whole environment from scratch (fleet manager, fleet shard) and run e2e in this isolated setup.

The goal for app-interface e2e is to spot if latest commit breaks the current stage environment.
The blackbox probe service will spot if some configuration/timed event breaks stage environment.

## Test manual

Local run staging e2e

```
# To run tests locally run:
export OCM_USERNAME=
export STATIC_OCM_TOKEN=
export ACS_FLEET_MANAGER_ENDPOINT=
e2e/run_e2e_stage.sh

...

W1124 15:51:33.429562       1 probe.go:195] central instance cdvp46glmtu28nlikv50 not in target state "ready"
I1124 15:51:38.465846       1 probe.go:191] central instance cdvp46glmtu28nlikv50 is in "ready" state
I1124 15:51:44.010257       1 probe.go:159] deletion of central instance cdvp46glmtu28nlikv50 requested
W1124 15:51:49.139794       1 probe.go:222] central instance cdvp46glmtu28nlikv50 not deleted
W1124 15:51:54.138467       1 probe.go:222] central instance cdvp46glmtu28nlikv50 not deleted
W1124 15:51:59.144168       1 probe.go:222] central instance cdvp46glmtu28nlikv50 not deleted
I1124 15:52:04.112945       1 probe.go:214] central instance cdvp46glmtu28nlikv50 has been deleted
I1124 15:52:04.112999       1 probe.go:52] elapsed time: 4m26.670280046s
I1124 15:52:04.113014       1 probe.go:80] probe run has ended
I1124 15:52:09.247799       1 probe.go:112] finished clean up attempt of probe resources
```


stehessel:
Can you give some more context on the nature of the app-interface CI run vs the openshift CI run? I'm a bit confused of what we are attempting to test with this change. If this is meant as a PR gate, shouldn't we run the probe against a fleet-manager compiled from the PR commit? If it's just running the probe against the stage fleet manager on a regular basis, we already do this with the probe service. Running duplicate probes via the CI seems redundant.

e2e/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM registry.access.redhat.com/ubi8-minimal:8.6 AS builder
```
stehessel, line 1: I think you can just use `probe/Dockerfile`, you don't have to build a dedicated image.

Makefile
```diff
@@ -576,6 +579,22 @@ test/run: image/build/test
...
+
+# Run the probe based e2e test in container
+test/e2e/probe/run: image/build/e2e/probe
+	$(DOCKER) run \
+	-e QUOTA_TYPE="OCM" \
+	-e AUTH_TYPE="OCM" \
+	-e OCM_USERNAME="${OCM_USERNAME}" \
+	-e OCM_TOKEN="${OCM_TOKEN}" \
+	-e FLEET_MANAGER_ENDPOINT="${FLEET_MANAGER_ENDPOINT}" \
+	--rm -it "$(e2e_probe_image)"
```
stehessel, line 595: The probe Dockerfile defines the probe binary as the entrypoint and `start` as the command. Therefore you need to add `run` at the end here to use `probe run` instead of `probe start`.

Makefile
```diff
@@ -576,6 +579,22 @@ test/run: image/build/test
 	$(DOCKER) run -u $(shell id -u) --net=host -p 9876:9876 -i "$(test_image)"
 .PHONY: test/run
 
+# Build the probe e2e test image
+image/build/e2e/probe:
```
stehessel, line 583: Can you reuse `image/build/multi-target/probe` here?

kurlov:
@stehessel I've updated PR description. I think app-interface e2e brings values in terms of spotting if some commit breaks stage env. On the other hand, the Probe service will notify if some settings or event breaks stage. App-interface e2e might look redundant but I believe it's a nice addition

stehessel:
> The goal for app-interface e2e is to spot if latest commit breaks the current stage environment.

But wouldn't it make more sense if we spot the breakage *before* we merge, i.e. as part of the CI pipeline? That's something valuable that the probe service is not able to deliver.

Otherwise I don't really see the point tbh. Let's say we do merge a bad commit that is spotted by the probe:
* We receive a PagerDuty alert because the probe service runs will fail.
* We receive another PagerDuty alert because the app-interface pipeline broke.

In this case we will just duplicate alerts. 

However, if the breakage is because AMS went down randomly, the probe service still alerts because it runs continuously. But the app-interface run will not alert because the last run was successful.

stehessel (approved):
Discussed offline with @kurlov. The value proposition of the app-interface CI run in addition to the probe service is to give immediate feedback to devs. After their changes are deployed, a Slack message in the `acs-fleet-manager` channel will confirm that the probe run was successful.

PR #606, opened 2022-11-24
Fallback to EVAL if quota check fails

stehessel:

## Test manual

```
# To run tests locally run:
```


stehessel:
> Might be an obvious one, but: besides the probe service checking that it receives the correct instance type, do we have any other alerts that would highlight an issue with AMS?

No I don't think so. We could add metrics on number of EVAL vs STANDARD instances requested and alert based on those.

PR #617, opened 2022-11-29
ROX-13593: Add telemetry storage options to managed Centrals

stehessel:
Add telemetry storage options to managed Centrals. The plan is to use Segment as telemetry backend. This does not include service annotations to identify Central instances in Segment, only the parameters needed to push data to Segment itself.

The telemetry options are passed as env variables via `Customize` part of the CRD. This adds the env vars to all pods (Central, Scanner, Scanner-DB). I did not find a way to only add the env vars to Central - I think this would require a change to the CRD/RHACS operator code.

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
@@ -48,12 +48,9 @@ spec:
             - name: RHSSO_ENDPOINT
               value: {{ .Values.fleetshardSync.redHatSSO.endpoint }}
             - name: TELEMETRY_ENDPOINT
-              value: {{ .Values.fleetshardSync.telemetry.storage.endpoint }}
+              value: {{ .Values.fleetshardSync.telemetry.storage.endpoint | quote }}
             - name: TELEMETRY_STORAGE_KEY
-              valueFrom:
-                secretKeyRef:
-                  name: rhacs-telemetry-storage-key
-                  key: KEY
+              value: {{ .Values.fleetshardSync.telemetry.storage.key | quote }}
           ports:
             - name: monitoring
               containerPort: 8080
--- a/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
@@ -19,34 +19,41 @@ spec:
     spec:
       serviceAccountName: fleetshard-sync
       containers:
-      - name: fleetshard-sync
-        image: {{ .Values.fleetshardSync.image | quote }}
-        imagePullPolicy: IfNotPresent
-        command:
-        - /usr/local/bin/fleetshard-sync
-        env:
-        - name: OCM_TOKEN
-          value: {{ .Values.fleetshardSync.ocmToken }}
-        - name: FLEET_MANAGER_ENDPOINT
-          value: {{ .Values.fleetshardSync.fleetManagerEndpoint }}
-        - name: CLUSTER_ID
-          value: {{ .Values.fleetshardSync.clusterId }}
-        - name: CREATE_AUTH_PROVIDER
-          value: "{{ .Values.fleetshardSync.createAuthProvider }}"
-        - name: AUTH_TYPE
-          value: {{ .Values.fleetshardSync.authType }}
-        - name: STATIC_TOKEN
-          value: {{ .Values.fleetshardSync.staticToken }}
-        - name: EGRESS_PROXY_IMAGE
-          value: {{ .Values.fleetshardSync.egressProxy.image | quote }}
-        - name: RHSSO_SERVICE_ACCOUNT_CLIENT_ID
-          value: {{ .Values.fleetshardSync.redHatSSO.clientId }}
-        - name: RHSSO_SERVICE_ACCOUNT_CLIENT_SECRET
-          value: {{ .Values.fleetshardSync.redHatSSO.clientSecret }}
-        - name: RHSSO_REALM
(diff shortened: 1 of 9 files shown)


dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -19,34 +19,41 @@ spec:
...
+              value: {{ .Values.fleetshardSync.redHatSSO.realm }}
+            - name: RHSSO_ENDPOINT
+              value: {{ .Values.fleetshardSync.redHatSSO.endpoint }}
+            - name: TELEMETRY_ENDPOINT
+              value: {{ .Values.fleetshardSync.telemetry.storage.endpoint }}
+            - name: TELEMETRY_STORAGE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: rhacs-telemetry-storage-key
+                  key: KEY
```
0x656b694d, line 56: Why secret? Other tokens above are set explicitly via values.

stehessel: Just best practice imo. The other secrets should also use secrets instead of raw envs. With the telemetry key in particular it's of course a bit debatable because the on-prem equivalent of the API key will not really be secret anyway, but alas...

fleetshard/config/config.go
```diff
@@ -25,6 +25,8 @@ type Config struct {
 	CreateAuthProvider   bool          `env:"CREATE_AUTH_PROVIDER" envDefault:"false"`
 	MetricsAddress       string        `env:"FLEETSHARD_METRICS_ADDRESS" envDefault:":8080"`
 	EgressProxyImage     string        `env:"EGRESS_PROXY_IMAGE"`
+	TelemetryEndpoint    string        `env:"TELEMETRY_ENDPOINT"`
```
stehessel, line 28: @0x656b694d Do we need to add `ROX_INIT_TELEMETRY_ENABLED` here as well?

0x656b694d: I'd use empty `TELEMETRY_STORAGE_KEY` as the sign of disabled telemetry. So no.

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
stehessel: Well it's not like we want to freely publish our API token. The only reason that it's "non-secret" is that we technically cannot prevent leakage. But I see your point, I'm fine with moving the segment key to plain env.

fleetshard/config/config.go
stehessel: I saw `ROX_INIT_TELEMETRY_ENABLED` in the stackrox code, so we should remove it there then.

stehessel:
The e2e test is still failing. Not sure why, but will investigate.

stehessel:
* main
  * **PR #617** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/617" target="_blank">Graphite</a>  👈
    * **PR #627** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/627" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/617?utm_source=stack-comment).

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -25,6 +25,11 @@ fleetshardSync:
     realm: "redhat-external"
   egressProxy:
     image: "registry.redhat.io/openshift4/ose-egress-http-proxy:v4.11.0"
+  # API key to push telemetry data to a remote backend. Leaving it empty disables telemetry.
+  telemetry:
+    storage:
+      endpoint: ""
+      key: ""
```
dhaus67, line 32:
```suggestion
  telemetry:
    storage:
      endpoint: ""
      # API key to push telemetry data to a remote backend. Leaving it empty disables telemetry.
      key: ""
```

dhaus67 (commented):
The code itself looks good to me, I'd have a more conceptual question:

The value of `TELEMETRY_STORAGE_KEY` also has to be set within other installation methods for on-prem installations (i.e. helm chart).

I would have a couple of questions for this before merging this PR:
- Do we plan to expose the telemtry options (such as opt-in/opt-out) via specific Helm settings?
- Shouldn't the opt-in/-out be a part of the operator then, allowing us here to simply reuse the setting instead of doing this workaround by using `customize.envVars`?
- Is the value of `TELEMETRY_SOTRAGE_KEY` different from on-prem / ACSCS?

Additionally, we could make this more specific by adjusting the operator to read from annotations ([like we have done for the managed service specific settings](https://github.com/stackrox/stackrox/commit/f72be438ecaad367d5f3c7e8abfdc05422a4d4b0)) and skip using the `customize.envVars` approach (but, to be fair, this might be a bit over-engineering and obsolete, depending on the answers to the above questions).

stehessel:
We haven't really discussed the on-prem plans yet (opt-in vs opt-out, flag in Helm chart vs UI prompt in Central, ...). I'll be honest I was going for the quick way here because of the extreme iteration times that a change in the operator brings. But I realize now that is probably too much of a hack.

I think your suggestion makes a lot of sense, i.e. have a flag in the Helm chart and Central CRD for the operator.

fleetshard/config/config.go
0x656b694d: Yes, with the reference in roxctl.

0x656b694d (approved):
LGTM

stehessel:
Will reopen after #140 and #4030 have landed.

PR #580, opened 2022-11-15
Fix removal of centrals that did not get assigned a client origin

dhaus67:
## Description

Currently, it's possible to run into an error within the deletion worker for centrals.

This is possible when the deletion will be triggered during provisioning of the central. At this specific point, no `ClientOrigin` will be associated with the central request.

Since the client origin will be unset, the deletion will currently fail.

Instead of being strict on the deletion of these central requests and erroring out, I've dediced to instead log the occurrence.

## Checklist (Definition of Done)
- [ ] Unit and integration tests added
- [x] Added test description under `Test manual`
- [ ] Evaluated and added CHANGELOG.md entry if required
- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)
- [x] CI and all relevant tests are passing
- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.

## Test manual

```
# To run tests locally run:
make db/teardown db/setup db/migrate

# Start the fleet manager.
./fleet-manager serve ...

# Start the fleet shard sync.
CLUSTER_ID=... ./fleetshard-sync

# Create a central request.
./scripts/create-central.sh

# Immediately after creating the central request, delete the central request.
curl -H "Authorization: Bearer $(ocm token)" http://127.0.0.1:8000/api/rhacs/v1/centrals/<id>?async=true -X DELETE

# Ensure that the central request is properly cleaned up and no error will be thrown.
```

PR #618, opened 2022-11-29
ROX-12921: New worker to clean up timed out central requests

ivan-degtiarenko:
## Description
This PR introduces time out to central requests. If the timeout is reached and a central request is still in progress(statuses `Accepted`, `Preparing`, `Provisioning`), then the request is failed and all associated resources are cleaned up.

What this PR has:
* new worker to reconcile timed-out central requests
* timeout configuration parameter

What will be introduced in other PRs:
* metrics and alerts
* introducing timeout environment variable

## Test manual

1. `./fleet-manager serve --central-request-expiration-timeout=1`
2.  Create central request via `curl`
3. Wait for 1 minute and see lines similar to
```
I1129 06:41:28.065697   99362 centrals_timeout_mgr.go:49] reconciling centrals reached timeout
I1129 06:41:28.071921   99362 centrals_timeout_mgr.go:59] timed out centrals count = 1
I1129 06:41:28.071938   99362 centrals_timeout_mgr.go:62] timed out central id = ce2pmeg7t2uo88ikr00g
```

### DIFF
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
@@ -36,7 +37,7 @@ type CentralConfig struct {
 	// Set in minutes. After timeout is reached, central request is cancelled and marked as failed.
 	// TODO: this parameter does not belong here, as it's configuration of central request, not central.
 	// TODO: However, for the time being there's no better place to put this parameter.
-	CentralRequestExpirationTimeout int `json:"central_request_expiration_timeout"`
+	CentralRequestExpirationTimeout time.Duration `json:"central_request_expiration_timeout"`
 }
 
 // NewCentralConfig ...
@@ -50,7 +51,7 @@ func NewCentralConfig() *CentralConfig {
 		Quota:                            NewCentralQuotaConfig(),
 		CentralIDPClientSecretFile:       "secrets/central.idp-client-secret", //pragma: allowlist secret
 		CentralIDPIssuer:                 "https://sso.redhat.com/auth/realms/redhat-external",
-		CentralRequestExpirationTimeout:  60, // minutes
+		CentralRequestExpirationTimeout:  60 * time.Minute,
 	}
 }
 
@@ -68,7 +69,7 @@ func (c *CentralConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.CentralIDPClientID, "central-idp-client-id", c.CentralIDPClientID, "OIDC client_id to pass to Central's auth config")
 	fs.StringVar(&c.CentralIDPClientSecretFile, "central-idp-client-secret-file", c.CentralIDPClientSecretFile, "File containing OIDC client_secret to pass to Central's auth config")
 	fs.StringVar(&c.CentralIDPIssuer, "central-idp-issuer", c.CentralIDPIssuer, "OIDC issuer URL to pass to Central's auth config")
-	fs.IntVar(&c.CentralRequestExpirationTimeout, "central-request-expiration-timeout", c.CentralRequestExpirationTimeout, "Timeout for central requests")
+	fs.DurationVar(&c.CentralRequestExpirationTimeout, "central-request-expiration-timeout", c.CentralRequestExpirationTimeout, "Timeout for central requests")
 }
 
 // ReadFiles ...
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -32,6 +32,11 @@ type CentralConfig struct {
 	CentralIDPClientSecret     string `json:"central_idp_client_secret"`
 	CentralIDPClientSecretFile string `json:"central_idp_client_secret_file"`
 	CentralIDPIssuer           string `json:"central_idp_issuer"`
+
+	// Set in minutes. After timeout is reached, central request is cancelled and marked as failed.
+	// TODO: this parameter does not belong here, as it's configuration of central request, not central.
+	// TODO: However, for the time being there's no better place to put this parameter.
+	CentralRequestExpirationTimeout int `json:"central_request_expiration_timeout"`
 }
 
 // NewCentralConfig ...
@@ -45,6 +50,7 @@ func NewCentralConfig() *CentralConfig {
 		Quota:                            NewCentralQuotaConfig(),
 		CentralIDPClientSecretFile:       "secrets/central.idp-client-secret", //pragma: allowlist secret
 		CentralIDPIssuer:                 "https://sso.redhat.com/auth/realms/redhat-external",
+		CentralRequestExpirationTimeout:  60, // minutes
 	}
 }
 
@@ -62,6 +68,7 @@ func (c *CentralConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.CentralIDPClientID, "central-idp-client-id", c.CentralIDPClientID, "OIDC client_id to pass to Central's auth config")
 	fs.StringVar(&c.CentralIDPClientSecretFile, "central-idp-client-secret-file", c.CentralIDPClientSecretFile, "File containing OIDC client_secret to pass to Central's auth config")
 	fs.StringVar(&c.CentralIDPIssuer, "central-idp-issuer", c.CentralIDPIssuer, "OIDC issuer URL to pass to Central's auth config")
+	fs.IntVar(&c.CentralRequestExpirationTimeout, "central-request-expiration-timeout", c.CentralRequestExpirationTimeout, "Timeout for central requests")
 }
 
 // ReadFiles ...
--- a/internal/dinosaur/pkg/config/central.go
+++ b/internal/dinosaur/pkg/config/central.go
@@ -34,7 +34,6 @@ type CentralConfig struct {
 	CentralIDPClientSecretFile string `json:"central_idp_client_secret_file"`
 	CentralIDPIssuer           string `json:"central_idp_issuer"`
 
-	// Set in minutes. After timeout is reached, central request is cancelled and marked as failed.
 	// TODO: this parameter does not belong here, as it's configuration of central request, not central.
 	// TODO: However, for the time being there's no better place to put this parameter.
 	CentralRequestExpirationTimeout time.Duration `json:"central_request_expiration_timeout"`
--- a/internal/dinosaur/pkg/migrations/202211282100_add_central_timeout_lease_type.go
+++ b/internal/dinosaur/pkg/migrations/202211282100_add_central_timeout_lease_type.go
@@ -1,0 +1,40 @@
+package migrations
+
+// Migrations should NEVER use types from other packages. Types can change
+// and then migrations run on a _new_ database will fail or behave unexpectedly.
+// Instead of importing types, always re-create the type in the migration, as
+// is done here, even though the same type is defined in pkg/api
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"github.com/stackrox/acs-fleet-manager/pkg/api"
+	"github.com/stackrox/acs-fleet-manager/pkg/db"
+	"gorm.io/gorm"
+)
+
+const centralTimeoutLeaseType = "timeout_central"
+
+// addCentralTimeoutLease adds a leader lease value for the timeout_Central lease and its
+// worker.
+// It is similar to addLeaderLease.
+func addCentralTimeoutLease() *gormigrate.Migration {
+	return &gormigrate.Migration{
+		ID: "202211282100",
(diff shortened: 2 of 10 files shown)


stehessel:
> then the request is failed and all associated resources are cleaned up.

I wonder when we should clean up these resources. We probably don't want to remove the failed instance immediately, because that's confusing for users looking for their instance. Maybe we just should leave the failed instances until users delete them manually themselves?

internal/dinosaur/pkg/workers/dinosaurmgrs/centrals_timeout_mgr.go
```diff
@@ -0,0 +1,85 @@
...
+	// list central requests eligible to be cancelled
+	lastCreatedAt := time.Now().Add(-k.timeout)
+	timedOutCentralRequests, serviceErr := k.centralService.ListInProgressCentralsOlderThan(lastCreatedAt)
+
+	if serviceErr != nil {
+		encounteredErrors = append(encounteredErrors, errors.Wrap(serviceErr, "failed to list timed out centrals"))
+	} else {
+		glog.Infof("timed out centrals count = %d", len(timedOutCentralRequests))
+	}
+	for _, centralRequest := range timedOutCentralRequests {
```
SimonBaeumer, line 62: If my understanding is correct this also effects customer instances. This would mean a customer who creates a RHACS instance and it fails the request just disappears from the UI? 🤔

internal/dinosaur/pkg/workers/dinosaurmgrs/centrals_timeout_mgr.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (k *CentralsTimeoutManager) reconcileTimedOutCentral(centralRequest *dbapi.CentralRequest) error {
+	// Force deletion on the side of dataplane cluster in case central already went into provisioning stage.
+	if centralRequest.Status == constants2.CentralRequestStatusProvisioning.String() {
+		now := time.Now()
+		centralRequest.DeletionTimestamp = &now
+	}
+	centralRequest.Status = constants2.CentralRequestStatusFailed.String()
+	centralRequest.FailedReason = "Creation time went over the timeout. Interrupting central initialization."
```
SimonBaeumer, line 80: How could the instance recover from that error?
A customer must not see this error, our operations team should get notified.

internal/dinosaur/pkg/config/central.go
```diff
@@ -32,6 +32,11 @@ type CentralConfig struct {
 	CentralIDPClientSecret     string `json:"central_idp_client_secret"`
 	CentralIDPClientSecretFile string `json:"central_idp_client_secret_file"`
 	CentralIDPIssuer           string `json:"central_idp_issuer"`
+
+	// Set in minutes. After timeout is reached, central request is cancelled and marked as failed.
+	// TODO: this parameter does not belong here, as it's configuration of central request, not central.
+	// TODO: However, for the time being there's no better place to put this parameter.
+	CentralRequestExpirationTimeout int `json:"central_request_expiration_timeout"`
```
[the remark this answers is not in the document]
SimonBaeumer: +1

internal/dinosaur/pkg/workers/dinosaurmgrs/centrals_timeout_mgr.go
SimonBaeumer: I think this should only be a feature for probe services, for customer instances fleet-manager should only send a notification.

ivan-degtiarenko: > the request just disappears from the UI

We discussed it with @stehessel and agreed that we can avoid timed out instances from disappearing from the UI(for example, by introducing `TimedOut` status or changing the way we filter central requests in API).

> for customer instances fleet-manager should only send a notification.

I see the following reasons for time out on a stuck central request:
1. Occupies resources allocated for organization
* we have a limited quota and SSO clients per organization
2. Can be confusing for user 
* Should they delete such instance? Should they wait some more?
3. Protect us from users who don’t delete stuck instances
* Timeout is better than manually cleaning up the database from stuck instances

We can introduce "soft" timeout for sending a notification and "hard" timeout for failing central request and cleaning up resources, but I think at some point we should drop any central request - whether it is from a probe or from a customer, WDYT?

internal/dinosaur/pkg/workers/dinosaurmgrs/centrals_timeout_mgr.go
ivan-degtiarenko: What do you mean by "recover from that error"? Maybe I missed something, but I thought we don't recover central requests from failures.

internal/dinosaur/pkg/migrations/202211282100_add_central_timeout_lease_type.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/stackrox/acs-fleet-manager/pkg/db"
+	"gorm.io/gorm"
+)
+
+const centralTimeoutLeaseType = "timeout_central"
+
+// addCentralTimeoutLease adds a leader lease value for the timeout_Central lease and its
+// worker.
+// It is similar to addLeaderLease.
+func addCentralTimeoutLease() *gormigrate.Migration {
```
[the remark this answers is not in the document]
ivan-degtiarenko: Every worker has his own lease, and it should be in the database before the startup. I took https://github.com/stackrox/acs-fleet-manager/commit/41efa0e96f6062d890e7c834a9198c496c864241 as an example

ivan-degtiarenko:
> That is, wouldn't it be simpler to add a check in three places when the instances in accepted, preparing, and provisioning are returned from the database and before continuing processing?

@ebensh the idea is to be able to time out central requests with any status which can be seen as `in progress`. I think introducing a single worker allows us to avoid code duplication, as well as simplifies changes in case new status is added.

ivan-degtiarenko:
> I wonder when we should clean up these resources. We probably don't want to remove the failed instance immediately, because that's confusing for users looking for their instance. Maybe we just should leave the failed instances until users delete them manually themselves?

@stehessel I will cross-post results from our discussion in Slack here: we can show timed-out central instances after they failed. We can also introduce separate "soft" timeout for when an alert is raised and "hard" timeout for when resources associated with the central request are deleted.

internal/dinosaur/pkg/workers/dinosaurmgrs/centrals_timeout_mgr.go
stehessel: I'd say deleting failed instances because they consume resources is essentially an optimization. Given how rare we anticipate these failures to be, I don't think it's a very important one. In any case, I wouldn't call this "soft" and "hard" timeout. Rather a single timeout until failure is reached, and then optionally an independent garbage collection to optimize resource usage.

internal/dinosaur/pkg/workers/dinosaurmgrs/centrals_timeout_mgr.go
stehessel: I think we need to decide the failure model. Do we want the time out/failure to be unrecoverable (essentially ask users to create a new instance), or do we want to retry in the background similar to `ImagePullBackOff`? As far as I understand, fleet-manager's state model currently does the former.

internal/dinosaur/pkg/workers/dinosaurmgrs/centrals_timeout_mgr.go
```diff
@@ -0,0 +1,84 @@
...
+			continue
+		}
+	}
+
+	return encounteredErrors
+}
+
+func (k *CentralsTimeoutManager) reconcileTimedOutCentral(centralRequest *dbapi.CentralRequest) error {
+	centralRequest.Status = constants2.CentralRequestStatusFailed.String()
+	centralRequest.FailedReason = "Creation time went over the timeout. Interrupting central initialization."
```
stehessel, line 77:
```suggestion
	centralRequest.FailedReason = "Central creation timeout expired. Aborting central initialization."
```
"Went over" sounds too informal imo.

pkg/metrics/metrics.go
```diff
@@ -416,6 +421,24 @@ var centralStatusSinceCreatedMetric = prometheus.NewGaugeVec(
 	centralStatusSinceCreatedMetricLabels,
 )
 
+var centralTimeoutCountMetric = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Subsystem: FleetManager,
+		Name:      "central_timeout_total_count",
```
stehessel, line 427:
```suggestion
		Name:      "central_timeouts_total",
```
Prometheus counters end in `_total` by convention, the `count` is redundant.

pkg/metrics/metrics.go
```diff
@@ -416,6 +421,24 @@ var centralStatusSinceCreatedMetric = prometheus.NewGaugeVec(
...
+		Help:      "number of total timed-out centrals",
+	},
+	centralTimeoutCountMetricLabels,
+)
+
+// IncreaseCentralTimeoutCountMetric - increase counter for the centralTimeoutCountMetric
+func IncreaseCentralTimeoutCountMetric(centralID, clusterID string) {
```
stehessel, line 434:
```suggestion
func IncreaseCentralTimeoutCounter(centralID, clusterID string) {
```

stehessel (commented):
The code looks fine to me assuming we go with the timeout worker approach. However I do share Evan's concern about introducing unnecessary concurrency. Tbh I'm more concerned about race conditions than code duplication. We already saw one problem caused by side-effects of state transitions in https://github.com/stackrox/acs-fleet-manager/pull/580.

What I have noticed is that looking at https://github.com/stackrox/acs-fleet-manager/blob/6d676ef432f1b4bc34418cfe3afa73d8b932a94b/internal/dinosaur/pkg/workers/dinosaurmgrs/preparing_dinosaurs_mgr.go#L86, it seems that there is already something like a time out mechanism for `preparing` and `accepted` states. From looking at the code, this only seems to abort retries of internal errors after some time, but maybe it could be expanded to actual timeouts.

ivan-degtiarenko:
@stehessel  I saw that too. The problem is that we want to handle 3 states: `accepted`, `preparing` and `provisioning`. While the first two already have some timeout logic embedded, the third one is not actually reconciled by a worker - state changes from provisioning based on status sent by `fleetshard-sync`: https://github.com/stackrox/acs-fleet-manager/blob/0bcefb52ada13167c9d2250d27515adbdeb2d7ec/internal/dinosaur/pkg/services/data_plane_dinosaur.go#L57-L111

If we want to timeout `provisioning` requests(and I think we do as this the one step `fleet-manager` does not have control over), we would still need to introduce concurrency. I could rewrite https://github.com/stackrox/acs-fleet-manager/blob/0bcefb52ada13167c9d2250d27515adbdeb2d7ec/internal/dinosaur/pkg/workers/dinosaurmgrs/provisioning_dinosaurs_mgr.go + re-use code in `accepted` and `preparing` workers, but it still will introduce concurrent behaviour to the code. WDYT?

ebensh:
> @stehessel I saw that too. The problem is that we want to handle 3 states: `accepted`, `preparing` and `provisioning`. While the first two already have some timeout logic embedded, the third one is not actually reconciled by a worker - state changes from provisioning based on status sent by `fleetshard-sync`:
> 
> https://github.com/stackrox/acs-fleet-manager/blob/0bcefb52ada13167c9d2250d27515adbdeb2d7ec/internal/dinosaur/pkg/services/data_plane_dinosaur.go#L57-L111
> 
> If we want to timeout `provisioning` requests(and I think we do as this the one step `fleet-manager` does not have control over), we would still need to introduce concurrency. I could rewrite https://github.com/stackrox/acs-fleet-manager/blob/0bcefb52ada13167c9d2250d27515adbdeb2d7ec/internal/dinosaur/pkg/workers/dinosaurmgrs/provisioning_dinosaurs_mgr.go + re-use code in `accepted` and `preparing` workers, but it still will introduce concurrent behaviour to the code. WDYT?

I don't think this is introducing more concurrency than already exists, but you're right that there's some concurrency issues, e.g. there's a race condition where the Central could be moved to deprovisioning after this check that it's in provisioning state (https://github.com/stackrox/acs-fleet-manager/blob/907b37fbf9708a5a4d89830db9e45a352f10f129/internal/dinosaur/pkg/services/data_plane_dinosaur.go#L100-L100), which would then move it back to Ready. This should be atomic and in a transaction, I think. In general the ownership around provisioning state (or anywhere the fleetshard sync has a say) is not ideal.

I do worry that introducing more states would be more copy&paste of timeout error handling code; I was hoping that the per-state code would be very straightforward (if central stuck > threshold set state to timedout). Can you try to go down that path in a separate PR and see if it's as much code as you think?

ivan-degtiarenko:
@ebensh @stehessel PTAL: https://github.com/stackrox/acs-fleet-manager/pull/667

ivan-degtiarenko:
Superseded by https://github.com/stackrox/acs-fleet-manager/pull/667. Closing the PR.

PR #622, opened 2022-11-30
Make probe service generate central name starting with char and max 32

kurlov:
## Description

Fleet Manager dos not allow creating a central which begins with a number:
```
curl -XPOST --header "Authorization: Bearer $(ocm token)" 'https://XXX.XXX.openshift.com/api/rhacs/v1/centrals?async=true' --data '{"region": "us-east-1", "cloud_provider": "aws", "multi_az": true, "name": "3akurlov-number-test"}' | jq
{
  "id": "32",
  "kind": "Error",
  "href": "/api/rhacs/v1/errors/32",
  "code": "RHACS-MGMT-32",
  "reason": "name does not match ^[a-z]([-a-z0-9]*[a-z0-9])?$",
  "operation_id": "XXX"
}
```
it is possible that the Probe service might generate a central name starting with number or longer that 32 characters on local machine or on CI. This PR fixes it. Every probe generated central names will start with `probe-` prefix.

Also, added small dinosaur renaming.

## Checklist (Definition of Done)

- [x] Unit and integration tests added
- [x] Added test description under `Test manual`
~- [ ] Evaluated and added CHANGELOG.md entry if required~
~- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~
- [x] CI and all relevant tests are passing
- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
~- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.~

## Test manual

**TODO:** Add manual testing efforts

```
make probe && ./probe/bin/probe run

...

W1130 16:50:27.156627   42410 probe.go:202] central instance ce3nmb0n2at7tjbvesi0 not in target state "ready"
W1130 16:50:32.148967   42410 probe.go:202] central instance ce3nmb0n2at7tjbvesi0 not in target state "ready"
W1130 16:50:37.144199   42410 probe.go:202] central instance ce3nmb0n2at7tjbvesi0 not in target state "ready"
I1130 16:50:42.157684   42410 probe.go:198] central instance ce3nmb0n2at7tjbvesi0 is in "ready" state
I1130 16:50:47.716239   42410 probe.go:166] deletion of central instance ce3nmb0n2at7tjbvesi0 requested
W1130 16:50:52.839356   42410 probe.go:229] central instance ce3nmb0n2at7tjbvesi0 not deleted
[...]

PR #626, opened 2022-11-30
Provide PROBE_NAME for e2e probe run

kurlov:
## Description

Fix app-interface e2e probe run flaky

Alternative to #622

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -591,6 +591,7 @@ test/e2e/probe/run:
 	$(DOCKER) run \
 	-e QUOTA_TYPE="OCM" \
 	-e AUTH_TYPE="OCM" \
+	-e PROBE_NAME="e2e-probe" \
 	-e OCM_USERNAME="${OCM_USERNAME}" \
 	-e OCM_TOKEN="${OCM_TOKEN}" \
 	-e FLEET_MANAGER_ENDPOINT="${FLEET_MANAGER_ENDPOINT}" \


stehessel:
Is it possible to inject something like a run ID into the probe name? If we just set it statically, there could be race conditions when multiple CI runs try to clean up the probe resources.

kurlov:
@stehessel I am not sure about unique CI env for each run but $RANDON should do the trick

Makefile
```diff
@@ -591,6 +591,7 @@ test/e2e/probe/run:
 	$(DOCKER) run \
 	-e QUOTA_TYPE="OCM" \
 	-e AUTH_TYPE="OCM" \
+	-e PROBE_NAME="e2e-probe${RANDOM}" \
```
stehessel, line 594: Where is `RANDOM` coming from? What is the length?

kurlov: `$RANDOM` is a part of unix or linux shell. It generates a random number between 0 and 32767. So it adds 5 additional characters at most

kurlov: I hope it will work on CI 🤞

stehessel: Huh TIL. It looks like it is only part of some shells. Of the ones I use (`fish`, `dash` and `bash`) only bash supports `$RANDOM`. So I'm not sure how POSIX compliant this is. Instead, I would propose to use `$$$$` to get the process id.

stehessel (approved):

Makefile
ebensh: TIL +1 :)

ebensh (approved):

PR #627, opened 2022-12-01
ROX-13593: Add tenant ID as a k8s label

stehessel:
## Description

The tenant ID may be used to query tenant resources in scripts and tooling. In addition, the tenant ID identifies the Segment group for telemetry users.

I decided to opt for a k8s label instead of an annotation because I think it fits better semantically as it identifies the resource and is useful in k8s queries.

As far as I understand the reconciler, the label should be set for all existing and new tenants.

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -40,6 +40,7 @@ const (
 	helmReleaseName = "tenant-resources"
 
 	managedServicesAnnotation = "platform.stackrox.io/managed-services"
+	tenantIDLabelKey = "rhacs.redhat.com/tenant"
 
 	centralDbSecretName = "central-db-password" // pragma: allowlist secret
 )
@@ -115,9 +116,12 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 
 	central := &v1alpha1.Central{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        remoteCentralName,
-			Namespace:   remoteCentralNamespace,
-			Labels:      map[string]string{k8s.ManagedByLabelKey: k8s.ManagedByFleetshardValue},
+			Name:      remoteCentralName,
+			Namespace: remoteCentralNamespace,
+			Labels: map[string]string{
+				k8s.ManagedByLabelKey: k8s.ManagedByFleetshardValue,
+				tenantIDLabelKey:      remoteCentral.Id,
+			},
 			Annotations: map[string]string{managedServicesAnnotation: "true"},
 		},
 		Spec: v1alpha1.CentralSpec{
@@ -431,7 +435,7 @@ func (r *CentralReconciler) getNamespace(name string) (*corev1.Namespace, error)
 
 func (r *CentralReconciler) createTenantNamespace(ctx context.Context, namespace *corev1.Namespace) error {
 	namespace.Labels = make(map[string]string)
-	namespace.Labels["rhacs.redhat.com/tenant"] = ""
+	namespace.Labels[tenantIDLabelKey] = ""
 	err := r.client.Create(ctx, namespace)
 	if err != nil {
 		return fmt.Errorf("creating namespace %q: %w", namespace.ObjectMeta.Name, err)
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -40,7 +40,7 @@ const (
 	helmReleaseName = "tenant-resources"
 
 	managedServicesAnnotation = "platform.stackrox.io/managed-services"
-	tenantIDLabelKey = "rhacs.redhat.com/tenant"
+	tenantIDLabelKey          = "rhacs.redhat.com/tenant"
 
 	centralDbSecretName = "central-db-password" // pragma: allowlist secret
 )
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -187,7 +187,7 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 		return nil, ErrDeletionInProgress
 	}
 
-	if err := r.ensureNamespaceExists(remoteCentralNamespace); err != nil {
+	if err := r.ensureNamespaceExists(remoteCentralNamespace, remoteCentral.Id); err != nil {
 		return nil, errors.Wrapf(err, "unable to ensure that namespace %s exists", remoteCentralNamespace)
 	}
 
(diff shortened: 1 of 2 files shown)


stehessel:
* main
  * **PR #617** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/617" target="_blank">Graphite</a> 
    * **PR #627** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/627" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/627?utm_source=stack-comment).

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -40,6 +40,7 @@ const (
 	helmReleaseName = "tenant-resources"
 
 	managedServicesAnnotation = "platform.stackrox.io/managed-services"
+	tenantIDLabelKey          = "rhacs.redhat.com/tenant"
```
0x656b694d, line 43: Why label and not annotation?

stehessel: I think it fits a bit better semantically because it identifies the resources. It's also more useful than an annotation because you can do something like `kubectl get pods -l tenant=xxx`. Not a big deal, but why not use it.

SimonBaeumer: +1 for label. Labels are used for identification, whereas annotations define different meta data and sometimes modify a pods behaviour (i.e. a controller).


Suggestion: being more explicit in the annotation key:

```suggestion
	tenantIDLabelKey          = "rhacs.redhat.com/tenant-id"
```

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -434,7 +438,7 @@ func (r *CentralReconciler) getNamespace(name string) (*corev1.Namespace, error)
 
 func (r *CentralReconciler) createTenantNamespace(ctx context.Context, namespace *corev1.Namespace) error {
 	namespace.Labels = make(map[string]string)
-	namespace.Labels["rhacs.redhat.com/tenant"] = ""
+	namespace.Labels[tenantIDLabelKey] = ""
```
SimonBaeumer, line 441: The `rhacs.redhat.com/tenant` had a different use-case. It was to select all namespaces for tenants, independent of the actual tenant configuration.
Please do not delete this, some queries require it.

Please add a new label and keep the `rhacs.redhat.com/tenant` label as it is.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -118,9 +119,12 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
 	central := &v1alpha1.Central{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        remoteCentralName,
-			Namespace:   remoteCentralNamespace,
-			Labels:      map[string]string{k8s.ManagedByLabelKey: k8s.ManagedByFleetshardValue},
+			Name:      remoteCentralName,
+			Namespace: remoteCentralNamespace,
+			Labels: map[string]string{
+				k8s.ManagedByLabelKey: k8s.ManagedByFleetshardValue,
+				tenantIDLabelKey:      remoteCentral.Id,
```
SimonBaeumer, line 126: WDYTH about setting the tenant-id label also on the namespace?

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
0x656b694d: Well, ok, I can see a potential use case to find pods in multi-tenant setup one day.
This ID is going to modify central pods behavior. I'll need to update the backend, once we agree on the naming.
@SimonBaeumer, please expand how the similar namespace label is used, why cannot the name be the same?

fleetshard/pkg/central/reconciler/reconciler.go
stehessel: But currently it's just an empty label, I don't really get it. Couldn't you still query all namespaces via `kubectl get namespaces -l rhacs.redhat.com/tenant`, even if each label value would be the actual id? Can you maybe post the query that you have in mind?

fleetshard/pkg/central/reconciler/reconciler.go
stehessel: I would rather set the `rhacs.redhat.com/tenant` on the namespace to the actual id.

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: > But currently it's just an empty label, I don't really get it.

Empty labels and annotation are often used in Kubernetes to identify something. The existence of the label is equal to `rhacs.redhat.com/tenant=true`.

> Couldn't you still query all namespaces via kubectl get namespaces -l rhacs.redhat.com/tenant, even if each label value would be the actual id?

I am honestly not sure if referencing a label without the value works in this scenario, never tried it. 

>  Can you maybe post the query that you have in mind?

The query you mentioned is what I have in mind: `kubectl get namespaces -l rhacs.redhat.com/tenant`
As discussed in a thread above with @0x656b694d: https://github.com/stackrox/acs-fleet-manager/pull/627#discussion_r1039659277

SimonBaeumer: Referencing the label name with values works as well:

```
❯ k get ns -l "kubernetes.io/metadata.name"
NAME              STATUS   AGE
default           Active   12d
kube-system       Active   12d
kube-public       Active   12d
kube-node-lease   Active   12d
stackrox          Active   11d
test1             Active   21s
❯ k get ns -l "kubernetes.io/metadata.name=test1"
NAME    STATUS   AGE
test1   Active   23s
```

fleetshard/pkg/central/reconciler/reconciler.go
stehessel: But you don't have to specify the value to query for the existence of the label.

SimonBaeumer: >But you don't have to specify the value to query for the existence of the label.

That's correct.

@stehessel Are you worried about the number of labels on an k8s object? 🤔

stehessel: In this case yes (well not really worried, just a bit baffled). I see no reason for two tenant identifier labels.

SimonBaeumer: Agree, let's use `rhacs.redhat.com/tenant` then.

SimonBaeumer (dismissed):
Please add the tenant ID to the namespace labels and ready to 🚢

0x656b694d (approved):
LGTM

PR #636, opened 2022-12-05
ROX-13227: Fix negative DNS caching

kovayur:
## Description
Fixing negative DNS caching issue: Do not expose URLs until corresponding DNS records are created in Route53.

## Test manual
1. CI green
2. Test with UI

```
# To run tests locally run:
```


kovayur:
> This solves the problem that customers sees available links before DNS is available but if he goes to the URL he can suggest based on the central ID it will still fail so we might consider changing our SOA config to lesser retry time: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html. WDYTH?

Yeah, it could be. I think it's enough not to advertise for the user to access the link.

PR #638, opened 2022-12-06
ROX-13756: Add managed DB parameters to the Helm chart

kovayur:
## Description
This PR adds Managed DB specific parameters and enables them for Staging.

## Test manual
```
export HELM_PRINT_ONLY=true

./dp-terraform/helm/rhacs-terraform/terraform_cluster.sh stage acs-stage-dp-02
./dp-terraform/helm/rhacs-terraform/terraform_cluster.sh prod acs-prod-dp-01
```

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
+++ b/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
@@ -35,7 +35,6 @@ case $ENVIRONMENT in
     # Get the first non-merge commit, starting with HEAD.
     # On main this should be HEAD
     FLEETSHARD_SYNC_TAG="$(git rev-list --no-merges --max-count 1 --abbrev-commit --abbrev=7 HEAD)"
-    "${SCRIPT_DIR}/check_image_exists.sh" "${FLEETSHARD_SYNC_TAG}"
     ;;
 
   prod)
@@ -58,6 +57,12 @@ if [[ $CLUSTER_ENVIRONMENT != "$ENVIRONMENT" ]]; then
     exit 2
 fi
 
+if [[ "${HELM_PRINT_ONLY:-}" == "true" ]]; then
+    HELM_DEBUG_FLAGS="--debug --dry-run"
+else
+    "${SCRIPT_DIR}/check_image_exists.sh" "${FLEETSHARD_SYNC_TAG}"
+fi
+
 load_external_config "cluster-${CLUSTER_NAME}" CLUSTER_
 oc login --token="${CLUSTER_ROBOT_OC_TOKEN}" --server="$CLUSTER_URL"
 
@@ -75,8 +80,8 @@ if [[ "${OPERATOR_USE_UPSTREAM}" == "true" ]]; then
     OPERATOR_SOURCE="rhacs-operators"
 fi
 
-# helm template --debug ... to debug changes
-helm upgrade rhacs-terraform "${SCRIPT_DIR}" \
+# shellcheck disable=SC2086
+helm upgrade rhacs-terraform "${SCRIPT_DIR}" ${HELM_DEBUG_FLAGS:-} \
   --install \
   --namespace rhacs \
   --create-namespace \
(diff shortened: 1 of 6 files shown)


dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -58,10 +64,12 @@ if [[ $CLUSTER_ENVIRONMENT != "$ENVIRONMENT" ]]; then
     exit 2
 fi
 
-# Get the first non-merge commit, starting with HEAD.
-# On main this should be HEAD, on production, the latest merged main commit.
-FLEETSHARD_SYNC_TAG="$(git rev-list --no-merges --max-count 1 --abbrev-commit --abbrev=7 HEAD)"
-"${SCRIPT_DIR}/check_image_exists.sh" "${FLEETSHARD_SYNC_TAG}"
+if [[ -z ${FLEETSHARD_SYNC_TAG:-} ]]; then
```
porridge, line 67: If I'm interpreting this correctly, you'd like to drive the image tag from parameter store?
I'm not sure this is a good idea, for two reasons:
- even harder to see what revision was rolled out for a given commit - a concern that @ebensh had
- from change process PoV, since it allows software version to be changed without a review.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -44,6 +46,10 @@ case $ENVIRONMENT in
 
     OPERATOR_USE_UPSTREAM=false
     OPERATOR_VERSION="v3.72.0"
+
+    FLEETSHARD_SYNC_MANAGED_DB_ENABLED=false
+    FLEETSHARD_SYNC_MANAGED_DB_SUBNET_GROUP=""
+    FLEETSHARD_SYNC_MANAGED_DB_SECURITY_GROUP=""
```
porridge, line 52: I see production and staging is diverging more and more. This means we cannot trust that what worked in stage will work in production. What's the plan here?

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -79,8 +87,16 @@ if [[ "${OPERATOR_USE_UPSTREAM}" == "true" ]]; then
     OPERATOR_SOURCE="rhacs-operators"
 fi
 
-# helm template --debug ... to debug changes
-helm upgrade rhacs-terraform "${SCRIPT_DIR}" \
+helm_upgrade() {
```
porridge, line 90: Why do we need a function? Why not just something like the following (note lack of quotes):
```
if [[ "${DEBUG:-}" == "true" ]]; then
    HELM_DEBUG_FLAGS="--debug --dry-run"
fi 
...
helm upgrade ... ${HELM_DEBUG_FLAGS:-} \
```

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -79,8 +87,16 @@ if [[ "${OPERATOR_USE_UPSTREAM}" == "true" ]]; then
...
+    declare -a HELM_EXTRA_FLAGS
+    if [[ "${DEBUG:-}" == "true" ]]; then
```
porridge, line 92: `$DEBUG` is very generic and I wouldn't expect it to imply dry-run too. How about something like:

```suggestion
    if [[ "${HELM_PRINT_ONLY:-}" == "true" ]]; then
```

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
kovayur: No, I wanted to test it locally. Currently it is not possible to do this from the branch.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
vladbologa: It's temporary. We want to test Managed DB on staging, before the (planned) release to production next week.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: Alright, perhaps then instead ignore image check failure in the dry-run mode?

kovayur: Done.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: Change here does not mean we're promoting yet. Promotion happens through the merge to the `production` branch. I'd much rather do changes in sync here than forget to do something before the merge to production, like we (almost?) did [here](https://github.com/stackrox/acs-fleet-manager/pull/639/files#diff-9bef9a0b8f2c4df1b38bc82e7f9dc526d5edffb14eb8d97f595ec940fbd15430R41) for example.

vladbologa: Oh we'll surely not forget to bring these in sync, and enable RDS in prod.

I would not enable it here in prod, and then risk that it gets enabled by mistake by somebody doing a release, before we see that the feature works in staging.

Trying it in staging is the only way to prove it works unfortunately, the dev and CI environments are different from what we have in prod when it comes to VPCs and VPC peering configuration.

porridge: Perhaps this is not in scope of this PR (I don't know the background) but I also had in mind these two lines that differ from stage:
```
    OPERATOR_USE_UPSTREAM=false
    OPERATOR_VERSION="v3.72.0"
```

porridge (approved):

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
kovayur: This is covered here: #641

kovayur: Diverging eliminated as suggested by @porridge

vladbologa (approved):

PR #653, opened 2022-12-08
ROX-13747: first stab at running terraforming in dry-run mode on PRs

porridge:
## Description

Unfortunately it is not as useful as it could be, but perhaps a superficial smoke test is better than nothing.

The reason is that the image is only pushed to the app-sre org on merges to `main`, so it does not exist for the tested PRs. Therefore I need to ignore the outcome of the check script (and avoid useless retries) with something like:

```
    "${SCRIPT_DIR}/check_image_exists.sh" "${FLEETSHARD_SYNC_TAG}" 0 || echo >&2 "Ignoring failed image check in dry-run mode."
```

Perhaps it would make sense to parametrize `check_image_exists.sh` to look in another org, where we (I assume) push on every commit. But this adds yet another variable and increases complexity so I'm not sure it's worth it.

## Test manual

Relying on CI.

### DIFF
--- a/.github/workflows/deploy-stage.yaml
+++ b/.github/workflows/deploy-stage.yaml
@@ -1,5 +1,7 @@
 name: Deploy Stage Env
 
+# Please keep dry-run-stage.yaml in sync as far as possible when modifying this file.
+
 concurrency: stage
 
 on:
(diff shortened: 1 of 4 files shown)


porridge:
When sending this PR I realized this will probably not work, because PR branches are not authorized to access the GH "environment" object which provides (or at least should provide) access to credentials. We'll need some further thought on how to make this work without giving the public the ability to access our staging environment at will by creating a PR. :thinking:

porridge:
superseded by https://github.com/stackrox/acs-fleet-manager/pull/915

PR #667, opened 2022-12-13
ROX-12921: Introduce timeout for ACSCS requests

ivan-degtiarenko:
## Description
Alternative approach to https://github.com/stackrox/acs-fleet-manager/pull/618 as suggested by Evan.
Instead of creating a new worker to check for a timeout in this PR existing workers check for timeout for instances in corresponding status.

There are a few caveats to the implementation:
* `accepted_centrals_mgr` has a separate timeout for a certain kind of transient errors
* `preparing_dinosaurs_mgr` has separate timeout for server errors, and fail request immediately on client errors. The only value timeout brings here is that the timeout is checked before `PrepareDinosaurRequest` is executed.
* `provisioning_dinosaurs_mgr` did not own transition from `Provisioning` state before. There is a race condition now between updates coming from `fleetshard-sync` and timeout logic. In worst case scenario, for a short moment instance will become `Ready` before switching to `Failed`. That, of course, is behaviour only possible when timeout was just hit. 

## Test manual

1. Deployed fleet-manager locally
2. Logged in to the database and saw `failed` status and that `failed_reason` is `Creation time went over the 
timeout. Interrupting central initialization.`
3. Accessed `localhost:8080/metrics` and saw individual records per timed out request:
```
# HELP acs_fleet_manager_central_timeout_total_count number of total timed-out centrals
# TYPE acs_fleet_manager_central_timeout_total_count counter
acs_fleet_manager_central_timeout_total_count{cluster_id="20fusbtpjsm0t7fabl144ptnt9e2iaig",id="cedjb207t2un94jubmrg"} 1
acs_fleet_manager_central_timeout_total_count{cluster_id="20fusbtpjsm0t7fabl144ptnt9e2iaig",id="cedje9g7t2un94jubmu0"} 1
acs_fleet_manager_central_timeout_total_count{cluster_id="20fusbtpjsm0t7fabl144ptnt9e2iaig",id="cedjfn87t2un94jubmvg"} 1
```
4. Verified that time passed before timeout is hit matches with configuration value

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -296,7 +296,7 @@
         "filename": "templates/service-template.yml",
         "hashed_secret": "4e199b4a1c40b497a95fcd1cd896351733849949",
         "is_verified": false,
-        "line_number": 663,
+        "line_number": 668,
         "is_secret": false
       }
     ],
@@ -321,5 +321,5 @@
       }
     ]
   },
-  "generated_at": "2022-11-24T16:16:02Z"
+  "generated_at": "2022-12-15T17:48:26Z"
 }
--- a/Makefile
+++ b/Makefile
@@ -305,7 +305,7 @@ install: verify lint
 .PHONY: install
 
 clean:
-	rm -f fleet-manager fleetshard-sync probe
+	rm -f fleet-manager fleetshard-sync probe/bin/probe
 .PHONY: clean
 
 # Runs the unit tests.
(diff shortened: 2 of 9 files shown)


stehessel:
Nice work! This does look like the simpler approach to me. Seems like it would make testing much simpler as well.

stehessel (approved):

PR #668, opened 2022-12-13
ROX-13396: Cleanup incomplete tools_venv dir for idempotence

porridge:
## Description

When [there are problems](https://issues.redhat.com/browse/ROX-13649) with completing the setup of the `tools_venv` directory, one is left with a non-working one, causing errors such as the following on the subsequent invocations of the terraforming script (or of whatever else tries to use it):

```
[mowsiany@mowsiany rhacs-terraform]$ HELM_PRINT_ONLY=true ./terraform_cluster.sh stage acs-stage-dp02
/home/mowsiany/go/src/github.com/stackrox/acs-fleet-manager/dp-terraform/helm/rhacs-terraform/../../../scripts/lib/external_config.sh: line 37: aws-saml.py: command not found
```

This change makes the make target remove the directory if it fails. Hoping the `ERR` trap is supported by the `bash` that other team members use :crossed_fingers:

- [x] Evaluated ~and added CHANGELOG.md entry if required~

## Test manual

1. removed the directory
2. introduced a change to the last line of the target that made it fail
3. ran the terraforming script, made sure the directory was missing, as well as on subsequent invocations
4. removed the artificial failure
5. ran the terraforming script, made sure the directory was present, and the did not happen on subsequent invocations


kovayur (approved):

Makefile
```diff
@@ -119,6 +119,7 @@ $(GINKGO_BIN): $(TOOLS_DIR)/go.mod $(TOOLS_DIR)/go.sum
 TOOLS_VENV_DIR := $(LOCAL_BIN_PATH)/tools_venv
 $(TOOLS_VENV_DIR):
 	@set -e; \
+	trap "rm -rf $(TOOLS_VENV_DIR)" ERR; \
```
stehessel, line 122: Make uses the default POSIX shell `/bin/sh`, which may or may not be bash. It seems like `trap .. ERR` is a bash-ism, but it can be made standard compliant (see https://stackoverflow.com/questions/29427364/non-bash-equivalent-to-trap-err).

porridge: > Make uses the default POSIX shell `/bin/sh`, which may or may not be bash

Our [Makefile makes it use bash](https://github.com/stackrox/acs-fleet-manager/blob/93c90c5806c816af36e2f9a622abe8a3671fde0f/Makefile#L7) AFAICT.

porridge:
Rebased to after #669 in an attempt to get tests to pass.

PR #672, opened 2022-12-13
ROX-13593: Add telemetry storage key to Centrals

stehessel:
## Description
Add telemetry storage options to managed Centrals. The plan is to use Segment as telemetry backend - this change adds the necessary config for that. The values are passed by fleetshard-sync to Central via its CRD. To be able to use the new telemetry CRD options, an update of the `stackrox/rox` module is included. 

## Test manual

No manual testing, just the unit test.


fleetshard/config/config.go
```diff
@@ -25,6 +25,8 @@ type Config struct {
 	CreateAuthProvider   bool          `env:"CREATE_AUTH_PROVIDER" envDefault:"false"`
 	MetricsAddress       string        `env:"FLEETSHARD_METRICS_ADDRESS" envDefault:":8080"`
 	EgressProxyImage     string        `env:"EGRESS_PROXY_IMAGE"`
+	TelemetryEndpoint    string        `env:"TELEMETRY_ENDPOINT"`
+	TelemetryStorageKey  string        `env:"TELEMETRY_STORAGE_KEY"`
```
dhaus67, line 29: We shall follow the example of spliting out our config knobs to a separate `Telemetry` struct to keep the ever-so-growing config struct a bit more comprehensable.

fleetshard/pkg/central/reconciler/telemetry.go
```diff
@@ -0,0 +1,12 @@
...
+const (
+	endpointVariable   = "ROX_TELEMETRY_ENDPOINT"
+	storageKeyVariable = "ROX_TELEMETRY_STORAGE_KEY_V1"
+)
+
+// TelemetryOptions defines parameters for pushing telemetry to a remote storage.
+type TelemetryOptions struct {
+	Endpoint   string
+	StorageKey string
+}
```
dhaus67, line 12: With the change mentioned above to move this out to a separate struct, we get away with not introducing the separate struct here but rather within the config (and also depend on it within the reconciler struct).

dhaus67 (approved):
We don't have upstream operators running in stage dataplane clusters, right?

Would be nice to actually go ahead and test this on stage once Michael's PRs have landed, but I suppose we are bound to the downstream release for now😞

PR #675, opened 2022-12-15
ROX-13106: Show hint to user when OCM user or org is not found

stehessel:
## Description
OCM does not load the user information from Red Hat SSO lazily. Instead, it requires that the user has logged into OCM at least once - this can be done by visiting https://console.redhat.com/openshift. See https://issues.redhat.com/browse/SDB-3194 for more context on this behavior.

Our approach here is to inform the user why the creation failed and how they can fix it themselves. 404 in `OrganisationNotFound` occurs when neither org nor user is registered with OCM. 404 in `FailedClusterAuthorization` occurs when the user is unknown to OCM, but they belong to an org that is registered with OCM.

## Test manual

Tested with local fleet manager and newly created RH accounts.


janisz (approved):
No single test needs to be changed?

stehessel:
> No single test needs to be changed?

No, as far I can tell we don't really test the error messages in this code base.

PR #680, opened 2022-12-15
ROX-13911: Add RDS provisioning integration test

vladbologa:
## Description

Adds an integration test that checks RDS data provisioning and deprovisioning logic, that:
* uses the ACSCS dev account to instantiate a temporary RDS DB
* checks that the instance & cluster are created and deleted on AWS
* runs on CI in a separate check, to save time
* is disabled by default when running the tests, because it requires extra configuration and takes about 10 minutes to run

## Test manual

Relying on the CI to run the tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,6 +60,9 @@ jobs:
   verify-test:
     name: "Verify & Test"
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
     environment: development
     services:
       postgres:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,6 +60,7 @@ jobs:
   verify-test:
     name: "Verify & Test"
     runs-on: ubuntu-latest
+    environment: dev
     services:
       postgres:
         image: postgres:11
@@ -100,7 +101,17 @@ jobs:
           echo -n "${{ secrets.POSTGRES_PASSWORD }}" > secrets/db.password
       - name: Run Migration Script
         run: make db/migrate
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          aws-region: ${{ secrets.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github-rds-integration-tests
       - name: Verify & Test
+        env:
+          RUN_RDS_TESTS: "true"
+          AWS_RDS_SECURITY_GROUP_ID: ${{ secrets.AWS_RDS_SECURITY_GROUP_ID }}
+          AWS_RDS_DB_SUBNET_GROUP_NAME: ${{ secrets.AWS_RDS_DB_SUBNET_GROUP_NAME }}
+          AWS_REGION: ${{ secrets.AWS_REGION }}
         run: |
           export GOPATH=$(go env GOPATH)
           export PATH=${PATH}:$GOPATH/bin
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,7 @@ jobs:
   verify-test:
     name: "Verify & Test"
     runs-on: ubuntu-latest
-    environment: dev
+    environment: development
     services:
       postgres:
         image: postgres:11
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,10 +60,6 @@ jobs:
   verify-test:
     name: "Verify & Test"
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write
-      contents: read
-    environment: development
     services:
       postgres:
         image: postgres:11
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
   verify-test:
     name: "Verify & Test"
     runs-on: ubuntu-latest
+    environment: dev
```
kovayur, line 63: nit:
```suggestion
    environment: development
```
to be aligned with another environment names.

fleetshard/pkg/central/cloudprovider/awsclient/rds_test.go
```diff
@@ -0,0 +1,108 @@
...
+	sess, err := session.NewSession(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create session, %w", err)
+	}
+
+	return rds.New(sess), nil
+}
+
+func TestRDSProvisioning(t *testing.T) {
+	if os.Getenv("RUN_RDS_TESTS") != "true" {
```
kovayur, line 46: Don't we want to make a separate test suite for RDS instead of introducing a skip flag? I think it will be cleaner + we could run RDS tests in parallel with the main suite in another action.

vladbologa: Yes, that was my original plan. I'd like to get this one pass first, and then I'll look into that.

vladbologa: Done.

.github/workflows/rds.yaml
```diff
@@ -0,0 +1,73 @@
...
+            ${{ runner.os }}-go-
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          aws-region: ${{ secrets.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
+      - name: Verify & Test
+        env:
+          AWS_RDS_SECURITY_GROUP_ID: ${{ secrets.AWS_RDS_SECURITY_GROUP_ID }}
+          AWS_RDS_DB_SUBNET_GROUP_NAME: ${{ secrets.AWS_RDS_DB_SUBNET_GROUP_NAME }}
```
vladbologa, line 67: @kovayur Is it possible to get these from the parameter store? I manually stored them as secrets in the GitHub development environment, for now.

kovayur: Yes, it is possible. In fact, they already should present in the Parameter Store.

.github/workflows/rds.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          aws-region: ${{ secrets.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
+      - name: Verify & Test
+        env:
+          AWS_REGION: ${{ secrets.AWS_REGION }}
+        run: |
+          source "scripts/lib/external_config.sh"
```
kovayur, line 68:
```suggestion
          set -euo pipefail
          source "scripts/lib/external_config.sh"
```

.github/workflows/rds.yaml
```diff
@@ -0,0 +1,76 @@
...
+          export AWS_AUTH_HELPER="none"
```
kovayur, line 69: nit: move to `env` section?

.github/workflows/rds.yaml
```diff
@@ -0,0 +1,76 @@
...
+          init_chamber
+          load_external_config fleetshard-sync AWS_RDS_
+
+          export GOPATH=$(go env GOPATH)
+          export PATH=${PATH}:$GOPATH/bin
```
kovayur, line 74: nit: I understand that it was inspired by `ci.yaml` but is it really necessary?

fleetshard/pkg/central/cloudprovider/awsclient/rds_test.go
kovayur: As far as I understand, you can add move `rds_test.go` under `/test` subdirectory to be excluded by `make test` target:
```
go list ./... | grep -v /test
```

kovayur (changes requested):
Good job @vladbologa, just a few minor remarks

.github/workflows/rds.yaml
vladbologa: Looks like it's not, removed.

.github/workflows/rds.yaml
vladbologa: Done.

fleetshard/pkg/central/cloudprovider/awsclient/rds_test.go
vladbologa: I'm also using unexported types from the `awsclient` package in these tests, so if I'm moving this file in a new directory (e.g. `awsclient/test/integration`) I won't have access to them.

So I'd rather keep this as is, unless you have another idea how to fix this.

kovayur: IMO an **integration** test should not use internal package types, this indicates that we are mixing test kinds (unit vs integration). Perhaps we could use the aws client to make sure the DB is (de)provisioned. It would have been a more "fair" test. Nevertheless, I don't want to block this PR, perhaps this should be left as food for thought.

kovayur (approved):

fleetshard/pkg/central/cloudprovider/awsclient/rds_test.go
vladbologa: I agree and I also thought about it, but the alternative is not perfect either... Basically I'd have to copy the `instanceExists` and `clusterExists` functions from awsclient into the test package.

PR #683, opened 2022-12-16
ROX-12344: Add organisation name to central request

stehessel:
## Description
The organisation name is added to the central request to make the cloud service more observable. The plan is to pass it together with the org id to the data plane, set it as label/annotation and then scrape it with Prometheus. Note that the organisation name is effectively immutable for users - it can only be changed in edge cases by Red Hat IT - however it is not unique. The org id is immutable and unique per account. The org name for existing tenants will be migrated using the OCM client.

* The organisation name is fetched from OCM during the request preparation phase.
* A migration is added to add a column for the org name to the database.
* `GetOrganisationIDFromExternalID` is changed to `GetOrganisationFromExternalID`, which now returns the complete organisation struct instead of just the id.
* Check for `nil` connection in OCM client. This is technically unrelated to this change, but got uncovered by the e2e tests. Previously fleet-manager would run into a nil pointer dereference when OCM was not setup properly - like previously in the e2e tests. Now we return `InvalidOCMConnection` error in this case.
* `org-name` is added to the private API to communicate the name to fleetshard-sync.

- [x] ~Added test description under `Test manual`

## Test manual
Tested with local fleet-manager and dev data-plane cluster.

### DIFF
--- a/internal/dinosaur/pkg/handlers/cloud_accounts_test.go
+++ b/internal/dinosaur/pkg/handlers/cloud_accounts_test.go
@@ -31,8 +31,9 @@ func TestGetSuccess(t *testing.T) {
 		Build()
 	assert.NoError(t, err)
 	c := ocm.ClientMock{
-		GetOrganisationIDFromExternalIDFunc: func(externalID string) (string, error) {
-			return "external-id", nil
+		GetOrganisationFromExternalIDFunc: func(externalId string) (*v1.Organization, error) {
+			org, _ := v1.NewOrganization().ID("external-id").Build()
+			return org, nil
 		},
 		GetCustomerCloudAccountsFunc: func(externalID string, quotaID []string) ([]*v1.CloudAccount, error) {
 			return []*v1.CloudAccount{
@@ -65,9 +66,10 @@ func TestGetSuccess(t *testing.T) {
 func TestGetNoOrgId(t *testing.T) {
 	timesClientCalled := 0
 	c := ocm.ClientMock{
-		GetOrganisationIDFromExternalIDFunc: func(externalID string) (string, error) {
+		GetOrganisationFromExternalIDFunc: func(externalId string) (*v1.Organization, error) {
 			timesClientCalled++
-			return "external-id", nil
+			org, _ := v1.NewOrganization().ID("external-id").Build()
+			return org, nil
 		},
 		GetCustomerCloudAccountsFunc: func(externalID string, quotaID []string) ([]*v1.CloudAccount, error) {
 			timesClientCalled++
@@ -100,8 +102,8 @@ func TestGetNoOrgId(t *testing.T) {
 func TestGetCannotGetOrganizationID(t *testing.T) {
 	timesClientCalled := 0
 	c := ocm.ClientMock{
-		GetOrganisationIDFromExternalIDFunc: func(externalID string) (string, error) {
-			return "", errors.New("test failure")
+		GetOrganisationFromExternalIDFunc: func(externalId string) (*v1.Organization, error) {
+			return nil, errors.New("test failure")
 		},
 		GetCustomerCloudAccountsFunc: func(externalID string, quotaID []string) ([]*v1.CloudAccount, error) {
 			timesClientCalled++
--- a/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
+++ b/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
@@ -58,13 +58,13 @@ func addOrganisationNameToCentralRequest() *gormigrate.Migration {
 		ID: "202212150000",
 		Migrate: func(tx *gorm.DB) error {
 			if err := tx.Migrator().AddColumn(&CentralRequest{}, "OrganisationName"); err != nil {
-				return fmt.Errorf("adding new column OrganisationName in migration 202212150000: %w", err)
+				return fmt.Errorf("adding column OrganisationName in migration 202212150000: %w", err)
 			}
 			return nil
 		},
 		Rollback: func(tx *gorm.DB) error {
 			if err := tx.Migrator().DropColumn(&CentralRequest{}, "OrganisationName"); err != nil {
-				return fmt.Errorf("rolling back new column OrganisationName in migration 202212150000: %w", err)
+				return fmt.Errorf("rolling back column OrganisationName in migration 202212150000: %w", err)
 			}
 			return nil
 		},
--- a/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
+++ b/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
@@ -93,7 +93,14 @@ func addOrganisationNameToCentralRequest(ocmConfig *ocm.OCMConfig) *gormigrate.M
 			}
 
 			rows, err := tx.Model(&CentralRequest{}).Where("organisation_name IS NULL").Rows()
-			defer rows.Close()
+			if err != nil {
+				return errors.Wrapf(err, "fetching rows where organisation_name is NULL in migration %s", id)
+			}
+			defer func() {
+				if err := rows.Close(); err != nil {
+					panic(errors.Wrapf(err, "closing rows in migration %s", id))
+				}
+			}()
 
 			for rows.Next() {
 				var central CentralRequest
--- a/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
+++ b/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
@@ -1,0 +1,72 @@
+package migrations
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"github.com/stackrox/acs-fleet-manager/pkg/api"
+	"gorm.io/gorm"
+)
+
+func addOrganisationNameToCentralRequest() *gormigrate.Migration {
+	type AuthConfig struct {
+		ClientID     string `json:"idp_client_id"`
+		ClientSecret string `json:"idp_client_secret"`
+		Issuer       string `json:"idp_issuer"`
+		ClientOrigin string `json:"client_origin"`
+	}
+
+	type CentralRequest struct {
+		api.Meta
+		Region           string   `json:"region"`
+		ClusterID        string   `json:"cluster_id" gorm:"index"`
+		CloudProvider    string   `json:"cloud_provider"`
+		CloudAccountID   string   `json:"cloud_account_id"`
+		MultiAZ          bool     `json:"multi_az"`
+		Name             string   `json:"name" gorm:"index"`
+		Status           string   `json:"status" gorm:"index"`
+		SubscriptionID   string   `json:"subscription_id"`
+		Owner            string   `json:"owner" gorm:"index"`
+		OwnerAccountID   string   `json:"owner_account_id"`
+		OwnerUserID      string   `json:"owner_user_id"`
+		Host             string   `json:"host"`
+		OrganisationID   string   `json:"organisation_id" gorm:"index"`
+		OrganisationName string   `json:"organisation_name"`
+		FailedReason     string   `json:"failed_reason"`
+		PlacementID      string   `json:"placement_id"`
+		Central          api.JSON `json:"central"`
+		Scanner          api.JSON `json:"scanner"`
+
+		DesiredCentralVersion         string     `json:"desired_central_version"`
+		ActualCentralVersion          string     `json:"actual_central_version"`
+		DesiredCentralOperatorVersion string     `json:"desired_central_operator_version"`
+		ActualCentralOperatorVersion  string     `json:"actual_central_operator_version"`
+		CentralUpgrading              bool       `json:"central_upgrading"`
+		CentralOperatorUpgrading      bool       `json:"central_operator_upgrading"`
+		InstanceType                  string     `json:"instance_type"`
+		QuotaType                     string     `json:"quota_type"`
+		Routes                        api.JSON   `json:"routes"`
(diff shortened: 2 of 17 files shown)


stehessel:
* main
  * **PR #683** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/683" target="_blank">Graphite</a>  👈
    * **PR #684** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/684" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/683?utm_source=stack-comment).

pkg/client/ocm/client.go
```diff
@@ -277,11 +317,19 @@ func (c client) CreateAddonWithParams(clusterID string, addonID string, params [
 
 // CreateAddon ...
 func (c client) CreateAddon(clusterID string, addonID string) (*clustersmgmtv1.AddOnInstallation, error) {
+	if c.connection == nil {
```
mtodor, line 320: We should add this check only in functions where we are actually using `c.connection`. This is one example where this check is not required because is not used within this function. I have noticed also `GetClusterDNS`, but I didn't check them all.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -328,12 +331,19 @@ func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralR
...
+		return errors.OrganisationNotFound(dinosaurRequest.OrganisationID, err)
+	}
+
 	// Update the fields of the CentralRequest record in the database.
 	updatedCentralRequest := &dbapi.CentralRequest{
 		Meta: api.Meta{
 			ID: dinosaurRequest.ID,
 		},
-		Status: dinosaurConstants.CentralRequestStatusProvisioning.String(),
+		OrganisationName: org.Name(),
```
mtodor, line 345: I have noticed that `Name()` can return an empty string. Is it important for us that name has to be populated, or can we ignore such cases?

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -38,14 +38,14 @@ var supportedAMSBillingModels = map[string]struct{}{
...
+	org, err := q.amsClient.GetOrganisationFromExternalID(dinosaur.OrganisationID)
 	if err != nil {
 		return false, errors.OrganisationNotFound(dinosaur.OrganisationID, err)
 	}
 
-	hasQuota, err := q.hasConfiguredQuotaCost(orgID, instanceType.GetQuotaType())
+	hasQuota, err := q.hasConfiguredQuotaCost(org.ID(), instanceType.GetQuotaType())
 	if err != nil {
-		return false, errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("failed to get assigned quota of type %v for organization with id %v", instanceType.GetQuotaType(), orgID))
+		return false, errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("failed to get assigned quota of type %v for organization with id %v", instanceType.GetQuotaType(), org.ID()))
```
mtodor, line 48: It would be good to add `dinosaur.OrganisationID` in this log message. Just to cover the case when `ID()` is empty. I guess that `hasConfiguredQuotaCost` will fail for an empty string.

The same applies to the logs below.

internal/dinosaur/pkg/handlers/cloud_accounts_test.go
```diff
@@ -31,8 +31,9 @@ func TestGetSuccess(t *testing.T) {
 		Build()
 	assert.NoError(t, err)
 	c := ocm.ClientMock{
-		GetOrganisationIDFromExternalIDFunc: func(externalID string) (string, error) {
-			return "external-id", nil
+		GetOrganisationFromExternalIDFunc: func(externalId string) (*v1.Organization, error) {
+			org, _ := v1.NewOrganization().ID("external-id").Build()
```
mtodor, line 35: consideration: Can we here use `return v1.NewOrganization().ID("external-id").Build()`? I have noticed that we have a lot of cases like this in tests where we ignore errors. Maybe it's ok for us just to return what we get from `Build`.

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
```diff
@@ -0,0 +1,72 @@
...
+		RoutesCreationID              string     `json:"routes_creation_id"`
+		DeletionTimestamp             *time.Time `json:"deletionTimestamp"`
+		AuthConfig
+	}
+
+	return &gormigrate.Migration{
+		ID: "202212150000",
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.Migrator().AddColumn(&CentralRequest{}, "OrganisationName"); err != nil {
+				return fmt.Errorf("adding new column OrganisationName in migration 202212150000: %w", err)
```
mtodor, line 61: nit-nitpick: I would suggest dropping "new" from the message. Then message will sound better:
```
adding column Abc
...
rolling back column Abc
```

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
```diff
@@ -0,0 +1,72 @@
...
+		Namespace                     string     `json:"namespace"`
+		RoutesCreationID              string     `json:"routes_creation_id"`
+		DeletionTimestamp             *time.Time `json:"deletionTimestamp"`
+		AuthConfig
+	}
+
+	return &gormigrate.Migration{
+		ID: "202212150000",
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.Migrator().AddColumn(&CentralRequest{}, "OrganisationName"); err != nil {
```
mtodor, line 60: Should we also populate values for the new column? As I see, we are fetching it on creating a new central instance.

mtodor (commented):
Looks good! The only open question is whether we should set the organization name during migration.

internal/dinosaur/pkg/handlers/cloud_accounts_test.go
stehessel: I usually prefer a local variable because you can see its value during debugging.

internal/dinosaur/pkg/migrations/migrations.go
```diff
@@ -37,6 +37,7 @@ var migrations = []*gormigrate.Migration{
 	addClientOriginToCentralRequest(),
 	changeCentralClientOrigin(),
 	addCloudAccountIDToCentralRequest(),
+	addOrganisationNameToCentralRequest(),
```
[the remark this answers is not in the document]
stehessel: I'd see this as outside the scope of this PR - I think I followed the existing convention.

internal/dinosaur/pkg/services/dinosaur.go
stehessel: Added a check for empty org names.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
stehessel: Maybe I'm blind, but what other logs are you referring to? I think this is the only such case where `org.ID()` is used in an error/log message.

stehessel:
> I see in the description that "org name for existing tenants will be empty". But maybe we can improve that and set the organization name during migration.

The concern I have is that to retrieve the org name, we will have to make a request to an external entity (OCM) during the migration. We would also have to pass the OCM client / service account credentials to the migration code. Tbh I'm not sure that's worth it. On the other hand, it would save us from manually patching the org name.

internal/dinosaur/pkg/migrations/migrations.go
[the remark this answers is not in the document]
mtodor: Fair enough. I have created a ticket to change that. https://issues.redhat.com/browse/ROX-14103

internal/dinosaur/pkg/services/quota/ams_quota_service.go
mtodor: You are right.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -328,12 +331,23 @@ func (k *dinosaurService) PrepareDinosaurRequest(dinosaurRequest *dbapi.CentralR
...
 	}
 
+	// Obtain organisation name from AMS to store in central request.
+	org, err := k.amsClient.GetOrganisationFromExternalID(dinosaurRequest.OrganisationID)
+	if err != nil {
+		return errors.OrganisationNotFound(dinosaurRequest.OrganisationID, err)
+	}
+	orgName := org.Name()
+	if orgName == "" {
+		return errors.OrganisationNameInvalid(dinosaurRequest.OrganisationID, org.Name())
```
mtodor, line 341: nitpick: We can use the variable here instead of calling the function again.
```suggestion
		return errors.OrganisationNameInvalid(dinosaurRequest.OrganisationID, orgName)
```

pkg/errors/errors.go
```diff
@@ -747,6 +753,12 @@ func OrganisationNotFound(externalID string, err error) *ServiceError {
 	return NewWithCause(svcErr.Code, svcErr, reason, externalID)
 }
 
+// OrganisationNameInvalid indicates that OCM organisation was found, but its name is invalid.
+func OrganisationNameInvalid(externalID string, name string) *ServiceError {
+	reason := "organisation with external id %s has invalid name %q"
+	return New(ErrorGeneral, reason, externalID, name)
```
mtodor, line 759: nit-nitpick: I would suggest removing the variable and placing the string inline. It will improve readability a bit. And I would suggest that we also quote ID to easier distinguish variables from the rest of the message.
```suggestion
	return New(ErrorGeneral, "organisation with external id %q has invalid name %q", externalID, name)
```

mtodor (commented):
It looks good. From my point of view, it can be merged when we clarify what we want to do with updating of organization name for current instances.

stehessel:
@mtodor Let's discuss the DB update of existing tenants next year.

stehessel:
@mtodor I added a migration of existing DB records using the OCM client.

stehessel:
rebased in the hopes that it fixes the e2e failure.

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
```diff
@@ -0,0 +1,127 @@
...
+				if err := rows.Close(); err != nil {
+					panic(errors.Wrapf(err, "closing rows in migration %s", id))
+				}
+			}()
+
+			for rows.Next() {
+				var central CentralRequest
+				tx.ScanRows(rows, &central)
+
+				orgName, err := fetchOrgName(amsClient, central.OrganisationID)
```
mtodor, line 109: As I can see from the code, there is still a chance that we will have an empty organization name. I don't have enough context to understand if that represents some issue for us. (i.e., bad-rendered UI or something similar)

If an empty org name is not a problem and we can keep it like that in DB, then we don't have to add any additional error handling. If not, and we should ensure that some non-empty string is set, we can think about using orgID to populate the organization name.

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
```diff
@@ -0,0 +1,127 @@
...
+		RoutesCreationID              string     `json:"routes_creation_id"`
+		DeletionTimestamp             *time.Time `json:"deletionTimestamp"`
+		AuthConfig
+	}
+
+	id := "202212150000"
+	return &gormigrate.Migration{
+		ID: id,
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.Migrator().AddColumn(&CentralRequest{}, "OrganisationName"); err != nil {
```
mtodor, line 86: When gorm creates a column will it be empty (value is NULL), or will it set an empty string? (I think it's an empty string)? I didn't test this, and from a glance at grom docs, I cannot find an answer.

A quick look at the implementation: https://github.com/go-gorm/gorm/blob/4b768c8aff4335eec41b0b393a7978bda1e6194d/migrator/migrator.go#L342 - seems that it will be NULL. Maybe you already know the answer. If not, it would be good to double check it.

stehessel: Should be NULL.

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
```diff
@@ -0,0 +1,127 @@
...
+			}
+			defer func() {
+				if err := rows.Close(); err != nil {
+					panic(errors.Wrapf(err, "closing rows in migration %s", id))
+				}
+			}()
+
+			for rows.Next() {
+				var central CentralRequest
+				tx.ScanRows(rows, &central)
```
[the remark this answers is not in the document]
stehessel: I'm not sure I get this point, how can we read a DB record without primary key? I don't think postgres allows that.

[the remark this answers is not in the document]
stehessel: But yeah I can add error handling to `ScanRows`, that makes sense.

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
stehessel: We do verify that orgs exist on creation, so I'm reasonably confident that all orgs still exist and have a name in AMS. However it could be that an org is deleted before we run the migration. It's very unlikely, but could in principle happen.

mtodor: Let's go with the current solution. If the org does not exist in AMS, but we still have central running with it, then it's also a good thing to discover.

mtodor (approved):

PR #684, opened 2022-12-16
ROX-12344: Add org-id label and org-name annotation

stehessel:
## Description
Add organisation meta data to the central tenant. The org name is set as an annotation, because it does not comply with Kubernetes validation rules.

We will scrape the label and annotations via Prometheus and pass them via metric labels to Grafana dashboards. This will be added as a follow up.

```
relabel_configs:
  - action: labelmap
    regex: __meta_kubernetes_pod_label_(.+)
  - action: labelmap
    regex: __meta_kubernetes_pod_annotation_(.+)
```

## Checklist (Definition of Done)

- [x] Unit and integration tests added
- [x] Added test description under `Test manual`
- [ ] ~Evaluated and added CHANGELOG.md entry if required~
- [ ] ~Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~
- [x] CI and all relevant tests are passing
- [x] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [ ] ~Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.~

## Test manual
Tested with local fleet-manager and dev data-plane cluster.

PR #690, opened 2022-12-17
ROX-12344: Use dev rhacs org_id in static token

stehessel:
## Description

This change is in preparation for looking up the org name via OCM in fleet manager.
We need the org in the static token to be real for the lookup to succeed. As the real org
I have chosen the RHACS dev organisation.

Note that the static token is currently used in the e2e tests and is injected via the vault
instance. We therefore have a chicken and egg problem:
* The e2e test of this PR will fail until we change the static token secret in the vault.
* If we change the vault secret without merging this PR, the e2e tests in other PRs will fail.

I suggest to change the secret during a quiet time, wait for a successful build and merge the PR.

- [ ] ~CI and all relevant tests are passing~

### DIFF
--- a/config/static-token-payload.json
+++ b/config/static-token-payload.json
@@ -28,7 +28,7 @@
   "is_org_admin": true,
   "account_id": "12345678",
   "user_id": "12345678",
-  "org_id": "12345678",
+  "org_id": "16155304",
   "first_name": "Test",
   "email": "rhacs-ms-test@redhat.com",
   "username": "rhacs-ms-test@redhat.com"
--- a/dev/env/manifests/shared/03-configmap-config.yaml
+++ b/dev/env/manifests/shared/03-configmap-config.yaml
@@ -26,7 +26,7 @@ data:
     # ACS RH SSO Org Prod
     - "16134752"
     # org_id for the static token
-    - "12345678"
+    - "16155304"
   fleetshard-authz-org-ids-prod.yaml: |-
     # ACS RH SSO Org Prod
     - "16134752"
@@ -205,7 +205,7 @@ data:
         max_allowed_instances: 3
         registered_users: []
         # Static token's org_id, see config/static-token-payload.json
-      - id: 12345678
+      - id: 16155304
         any_user: true
         max_allowed_instances: 100
         registered_users: []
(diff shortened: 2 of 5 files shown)


stehessel:
* main
  * **PR #690** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/690" target="_blank">Graphite</a>  👈
    * **PR #683** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/683" target="_blank">Graphite</a> 
      * **PR #684** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/684" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/690?utm_source=stack-comment).

config/static-token-payload.json
```diff
@@ -28,7 +28,7 @@
   "is_org_admin": true,
   "account_id": "12345678",
   "user_id": "12345678",
-  "org_id": "12345678",
+  "org_id": "16155304",
```
ebensh, line 31: This may be a stupid question now, but if this is a real account, should we be including its full bearer token in the public repo? I haven't traced through so this is /probably/ silly, but just to confirm: you couldn't use this token to hit the real (Stage or Prod) fleet manager, correct?

dev/env/manifests/shared/03-configmap-config.yaml
```diff
@@ -26,7 +26,7 @@ data:
     # ACS RH SSO Org Prod
     - "16134752"
     # org_id for the static token
```
ebensh, line 28: I'd appreciate clarifying "for the static token" to something more specific - there are lots of static tokens, what does this one do (I know but it'd be good to add)

kovayur (approved):

config/static-token-payload.json
stehessel: To answer your question more explicitly: this is only the token payload. While it is fake (see my comment about the `user_id`), you would be able to use this token payload to query fleet manager. However, to convert the payload into a valid bearer token accepted by fleet manager, the token must be signed by the correct JWKS. The private key used to generate the bearer value of the static token is secret, so one cannot simply create a bearer token from the payload.

PR #692, opened 2022-12-19
Fix: avoid panic on nil pointer

0x656b694d:
Debugging incident-acscs-dec-18.

### DIFF
--- a/internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
+++ b/internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
@@ -67,16 +67,16 @@ func (k *DinosaurRoutesCNAMEManager) Reconcile() []error {
 				if changeOutput == nil || changeOutput.ChangeInfo == nil || changeOutput.ChangeInfo.Id == nil || changeOutput.ChangeInfo.Status == nil {
 					switch {
 					case changeOutput == nil:
-						glog.Infof("ChangeDinosaurCNAMErecords returned nil value")
+						glog.Infof("ChangeDinosaurCNAMErecords returned nil value for central %s", dinosaur.ID)
 					case changeOutput.ChangeInfo == nil:
-						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo value")
+						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo value for central %s", dinosaur.ID)
 					case changeOutput.ChangeInfo.Id == nil:
-						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo.Id value")
+						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo.Id value for central %s", dinosaur.ID)
 					case changeOutput.ChangeInfo.Status == nil:
-						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo.Status value")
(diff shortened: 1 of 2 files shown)


kurlov (approved):

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
```diff
@@ -64,6 +64,21 @@ func (k *DinosaurRoutesCNAMEManager) Reconcile() []error {
...
+						glog.Infof("ChangeDinosaurCNAMErecords returned nil value")
+					case changeOutput.ChangeInfo == nil:
+						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo value")
+					case changeOutput.ChangeInfo.Id == nil:
+						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo.Id value")
+					case changeOutput.ChangeInfo.Status == nil:
+						glog.Infof("ChangeDinosaurCNAMErecords returned nil ChangeInfo.Status value")
+					}
+
+					errs = append(errs, errors.New("empty change output"))
```
kurlov, line 79: Could you please add central ID or Name here? Or both

porridge (approved):

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
porridge: It is logged above at the top of the for loop.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
```diff
@@ -65,6 +65,11 @@ func (k *DinosaurRoutesCNAMEManager) Reconcile() []error {
 					continue
 				}
 
+				if changeOutput == nil {
```
kovayur, line 68: nit: I'd also check `changeOutput.ChangeInfo` for `nil` to be 100% sure.

0x656b694d: And then the Status and the Id? Well, alright, I'll bring these checks back. It's safer to be safe.

pkg/client/aws/client.go
```diff
@@ -139,25 +138,9 @@ func (client *awsClient) ChangeResourceRecordSets(dnsName string, recordChangeBa
 
 	recordSetsOutput, err := client.route53Client.ChangeResourceRecordSets(recordChanges)
 
+	err = wrapAWSError(err, "Failed to get DNS zone.")
```
kovayur, line 141: I believe we don't need to `wrapAWSError` anymore. But TBH I have a hard time understanding why this was done.

kovayur (approved):

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
```diff
@@ -65,6 +65,15 @@ func (k *DinosaurRoutesCNAMEManager) Reconcile() []error {
 					continue
 				}
 
+				switch {
```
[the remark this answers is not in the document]
ebensh: I understand if we want to leave it where it is to get the fix in though.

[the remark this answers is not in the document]
0x656b694d: Let's leave it for now, as this should normally not be executed after the fix in the aws client.

PR #718, opened 2023-01-09
Revert "Bump stage Central version to nightly-2023-01-09 build (#717)"

stehessel:
This reverts commit 1553ab728f9c74b30b283764abaaf135c030382b.

## Description
The ACS operator install of `rhacs-operator.v3.73.0-nightly-20230109` fails with

```
bundle unpacking failed. Reason: DeadlineExceeded, and Message: Job was active longer than specified deadline
```

Revert to restore stage environment. Will need to investigate why this fails even when uninstalling/installing the ACS operator.


stehessel:
ACS operator has been recovered, no need to revert.

PR #720, opened 2023-01-11
ROX-12921: initialize counter metrics with 0 value

ivan-degtiarenko:
## Description
The focus of this PR is to introduce a 0 value as a default value for "slow" counter metrics.
"Slow" here means that the counter metric is not actively updating and can take a long time before being incremented for the first time.
The reason to do that is that to correctly use the Prometheus `rate` function in the alert, a zero value should be defined. `rate` function is used to determine when the counter metric goes up.

This PR does not solve the described issue for the central timeout metric. The reason for that is that the solution within this PR works only for metrics without dynamically changing labels.  The issue for the timeout metric will be fixed on the alert side.

## Test manual

1. Deploy local fleet-manager
2. Open `http://localhost:8080/metrics` and see worker counter metrics and operations counter metrics:
```
# HELP acs_fleet_manager_reconciler_failure_count count of failed operations of the backgroup reconcilers
# TYPE acs_fleet_manager_reconciler_failure_count counter
acs_fleet_manager_reconciler_failure_count{worker_type="accepted_dinosaur"} 0
acs_fleet_manager_reconciler_failure_count{worker_type="central_auth_config"} 0
acs_fleet_manager_reconciler_failure_count{worker_type="deleting_dinosaur"} 0
acs_fleet_manager_reconciler_failure_count{worker_type="dinosaur_dns"} 0
acs_fleet_manager_reconciler_failure_count{worker_type="preparing_dinosaur"} 0
acs_fleet_manager_reconciler_failure_count{worker_type="provisioning_dinosaur"} 0
acs_fleet_manager_reconciler_failure_count{worker_type="ready_dinosaur"} 0
```
and
```
# HELP acs_fleet_manager_central_operations_success_count number of successful central operations
# TYPE acs_fleet_manager_central_operations_success_count counter
acs_fleet_manager_central_operations_success_count{operation="create"} 0
acs_fleet_manager_central_operations_success_count{operation="delete"} 0
# HELP acs_fleet_manager_central_operations_total_count number of total central operations
# TYPE acs_fleet_manager_central_operations_total_count counter
acs_fleet_manager_central_operations_total_count{operation="create"} 0
acs_fleet_manager_central_operations_total_count{operation="delete"} 0
```


pkg/metrics/metrics.go
```diff
@@ -787,4 +787,28 @@ func Reset() {
 
 	databaseRequestCountMetric.Reset()
 	databaseQueryDurationMetric.Reset()
+
+	InitCounterMetricsWithZeroEmptyLabels()
```
stehessel, line 791: Don't we have to re-init `InitReconcilerMetricsForType` here as well?

ivan-degtiarenko: So the idea behind `InitReconcilerMetricsForType` is that I don't want  `pkg/metrics` to depend on `internal/dinosaur/pkg/workers/dinosaurmgrs/`. I'm pretty sure that would make it a cyclic dependency.

An alternative would be to avoid dependency and just duplicate all the worker type names within `pkg/metrics`, but that would introduce one more place where developers would need to add worker type if they want to create a new one.
Overall, it seems fair to me that workers are responsible for initializing their worker type with Prometheus. WDYT?

stehessel: Yeah that makes sense, however if we don't set the counters to zero after a reset, we could run into the unset counter metric issue after a new leader is elected. If I understand the current solution correctly, it would only init the worker type counters once on start up, but not on every reset. Do you know if `.Reset()` would lead to a zero value or unset value of the counter metric?

pkg/metrics/metrics.go
```diff
@@ -788,3 +792,30 @@ func Reset() {
...
+	reconcilerSuccessCountMetric.With(labels).Add(0)
+	reconcilerErrorsCountMetric.With(labels).Add(0)
+}
+
+// InitOperationMetricsWithZero initializes counter "operation" metrics with 0 value and all possible labels.
+// We do not initialize central timeout count metric as it has dynamic labels.
+// We do not initialize observatorium request count metric as it is unused.
+// We do not initialize database request count metric as it grows quickly and would not suffer from being undefined.
+// We initialize reconciler metrics in InitReconcilerMetricsForType.
+func InitOperationMetricsWithZero() {
```
stehessel, line 811: I think we can leave out the operations metrics completely for the same reason as the database request count. Then we don't need the comment, which seems like it could go stale very fast - e.g. once we remove the Observatorium code, or decide to use it after all. So I would propose to just do the minimum here, and only init the reconciler metrics.

stehessel (approved):

PR #722, opened 2023-01-13
ROX-13666: Refactor DB connection string

vladbologa:
## Description

In preparation for supporting multiple DB users (i.e. the initial `rds_superuser`, and a non-privileged user for Centrals) , this PR refactors the RDS provisioning client so that it returns the connection parameters as a struct, instead of a string. 

This will make it easier to generate connection strings for different users.

## Test manual

Tested locally by verifying that a DB was provisioned, and that the Central was able to connect:
`./scripts/create-central.sh`

To test locally, I have to make the DB publicly accessible. I added a new VPC, security group and DB subnet group in dev on AWS for this purpose.

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/central/postgres/postgres.go
+++ b/fleetshard/pkg/central/postgres/postgres.go
@@ -7,27 +7,37 @@ import (
 
 // DBConnection stores the data necessary to connect to a PostgreSQL server
 type DBConnection struct {
-	Host     string
-	Port     int
-	User     string
-	Database string
+	host     string
+	port     int
+	user     string
+	database string
 }
 
-// CreatePostgresConnectionString generates a connection string from a DBConnection struct
-func CreatePostgresConnectionString(connection DBConnection) (string, error) {
-	if connection.Host == "" {
-		return "", fmt.Errorf("host parameter cannot be empty")
+// NewDBConnection constructs a new DBConnection struct
+func NewDBConnection(host string, port int, user, database string) (DBConnection, error) {
+	if host == "" {
+		return DBConnection{}, fmt.Errorf("host parameter cannot be empty")
 	}
-	if connection.Port == 0 {
-		return "", fmt.Errorf("port parameter cannot be 0")
+	if port == 0 {
+		return DBConnection{}, fmt.Errorf("port parameter cannot be 0")
 	}
-	if connection.User == "" {
-		return "", fmt.Errorf("user parameter cannot be empty")
+	if user == "" {
+		return DBConnection{}, fmt.Errorf("user parameter cannot be empty")
 	}
-	if connection.Database == "" {
-		return "", fmt.Errorf("database parameter cannot be empty")
+	if database == "" {
+		return DBConnection{}, fmt.Errorf("database parameter cannot be empty")
 	}
 
+	return DBConnection{
+		host:     host,
+		port:     port,
+		user:     user,
+		database: database,
+	}, nil
+}
+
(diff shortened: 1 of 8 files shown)


fleetshard/pkg/central/postgres/postgres_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Port:     14543,
+		User:     "test-user",
+		Database: "postgresdb",
+	}
+
+	dbConnectionString, err := CreatePostgresConnectionString(dbConnection)
+	require.NoError(t, err)
+	require.Equal(t, dbConnectionString, "host=localhost port=14543 user=test-user dbname=postgresdb sslmode=require")
+
+	dbConnection.Host = ""
```
kovayur, line 21: nit: it is common to break such tests into several small test cases with descriptive names. Check [endpoints_test.go](https://github.com/stackrox/stackrox/blob/master/tests/endpoints_test.go#L364) for example

fleetshard/pkg/central/postgres/postgres_test.go
```diff
@@ -0,0 +1,39 @@
...
+	_, err = CreatePostgresConnectionString(dbConnection)
+	require.Error(t, err)
```
kovayur, line 23: validate error message? (here and below)

fleetshard/pkg/central/postgres/postgres.go
```diff
@@ -0,0 +1,33 @@
...
+// DBConnection stores the data necessary to connect to a PostgreSQL server
+type DBConnection struct {
+	Host     string
+	Port     int
+	User     string
+	Database string
+}
+
+// CreatePostgresConnectionString generates a connection string from a DBConnection struct
+func CreatePostgresConnectionString(connection DBConnection) (string, error) {
```
kovayur, line 17: 1) I think that `EnsureDBProvisioned` should already return a valid `DBConnection` instance or error if he can't create it. Therefore I'd move these checks to constructor;
2) `CreatePostgresConnectionString` func is a good candidate for an instance method.


```
type DBConnection struct {
	host     string
	port     int
	user     string
	database string
}

func NewDBConnection(host string, port int, user string, database string) (DBConnection, error) {
	if host == "" {
		return DBConnection{}, fmt.Errorf("host parameter cannot be empty")
	}
	if port == 0 {
		return DBConnection{}, fmt.Errorf("port parameter cannot be 0")
	}
	if user == "" {
		return DBConnection{}, fmt.Errorf("user parameter cannot be empty")
	}
	if database == "" {
		return DBConnection{}, fmt.Errorf("database parameter cannot be empty")
	}

	return DBConnection{
		host:     host,
		port:     port,
		user:     user,
		database: database,
	}, nil
}

func (c DBConnection) asConnectionString() string {
	return fmt.Sprintf("host=%s port=%d user=%s dbname=%s sslmode=require", c.host, c.port, c.user, c.database)
}
```

kovayur (changes requested):

fleetshard/pkg/central/postgres/postgres_test.go
vladbologa: Done

fleetshard/pkg/central/postgres/postgres.go
vladbologa: Done

kovayur (approved):
Thanks, LGTM 🚀

PR #723, opened 2023-01-13
ROX-16883: Increase resource limits

connorgorman:
## Description

Up for debating the increases in requests, but the limits should be more near to what we expect customers to run in production. The result of this is that it could lead to more evictions, but also should give customers much faster API times. I'm seeing timeouts in the metrics

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/internal/dinosaur/pkg/defaults/default_scanner_spec.go
+++ b/internal/dinosaur/pkg/defaults/default_scanner_spec.go
@@ -13,7 +13,7 @@ type AnalyzerDefaults struct {
 	MemoryRequest resource.Quantity `env:"MEMORY_REQUEST" envDefault:"100M"`
 	CPURequest    resource.Quantity `env:"CPU_REQUEST"    envDefault:"5m"`
 	MemoryLimit   resource.Quantity `env:"MEMORY_LIMIT"   envDefault:"2500M"`
-	CPULimit      resource.Quantity `env:"CPU_LIMIT"      envDefault:"250m"`
+	CPULimit      resource.Quantity `env:"CPU_LIMIT"      envDefault:"2"`
 	AutoScaling   string            `env:"AUTOSCALING"    envDefault:"Enabled"`
 	MinReplicas   int32             `env:"MIN_REPLICAS"   envDefault:"1"`
 	Replicas      int32             `env:"REPLICAS"       envDefault:"1"`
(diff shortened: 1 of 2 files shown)


internal/dinosaur/pkg/defaults/default_scanner_spec.go
```diff
@@ -11,9 +11,9 @@ import (
 // AnalyzerDefaults ...
 type AnalyzerDefaults struct {
 	MemoryRequest resource.Quantity `env:"MEMORY_REQUEST" envDefault:"100M"`
-	CPURequest    resource.Quantity `env:"CPU_REQUEST"    envDefault:"5m"`
+	CPURequest    resource.Quantity `env:"CPU_REQUEST"    envDefault:"100m"`
```
SimonBaeumer, line 14: The requests for CPU are so low because it reserves CPU resources. We hit a problem when too many instances are deployed to the cluster that the cluster run out of resources, even though the CPU usage was < 40%.
For now the idea is that instances with higher resource requirements are scaled manually vertically.

SimonBaeumer (changes requested):
@connorgorman Do you have any concerns about limiting the resources to lower than recommended?
So far we did not hit any issues, i.e. with hack-fest deploying >70 instances.
The secured clusters were very small in these scenarios.

connorgorman:
@SimonBaeumer I am seeing quite a few timeout: context deadline exceeded errors when contacting the database. Also, there seems to be a decent amount of throttling. I've removed the requests, but increased the limits so that customers can at least get a core of resources

kylape:
You could take care of https://issues.redhat.com/browse/ROX-16657 and bump memory request to `1Gi`.  We'll need to make sure we are ready to increase node count on our clusters in the likely event that ACS instances will not be able to be scheduled due to lack of allocatable memory.

connorgorman:
@kylape I had initially adjusted the requests in this PR, but there were concerns that we weren't utilizing the cluster enough

kylape:
I think when customers start using the cloud service for production
workloads, the average central will likely use more resources.

SimonBaeumer:
We can increase the resources but need an additional alert for memory and CPU requests exceeding cluster capacity. 
Added a ticket for this: https://issues.redhat.com/browse/ROX-16884

kylape:
let me try to sum up all the things

* Need to bump memory and cpu limits because things are slow
* This increases risk of evictions, so bump up cluster node count
* Evictions trigger incidents - this should be fixed by modifying the alert rule
* Evictions can also be reduced by increasing memory request, but this risks scheduling issues, especially at the point when requests are raised across the fleet

kylape (approved):

internal/dinosaur/pkg/defaults/default_scanner_spec.go
porridge: FTR, we had [an alert today](https://redhat-internal.slack.com/archives/C055YVB53M5) where the 0.005 core request might have been at least a contributing factor.

PR #733, opened 2023-01-16
Rox 13593: Add telemetry service

stehessel:
## Description
This adds a telemetry boot service to fleet-manager, which emits tenant grouping events. The telemetry client uses Segment as the backend.

The org name can be added to the group properties after #683 is merged.

## Test manual

Tested connection to Segment with local fleet-manager.


internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -0,0 +1,46 @@
...
+	if t != nil && t.config.Enabled() {
+		return
+	}
+	if telemeter := t.config.Telemeter(); telemeter != nil {
+		props := map[string]any{
+			"cloudAccount":   central.CloudAccountID,
+			"cloudProvider":  central.CloudProvider,
+			"instanceType":   central.InstanceType,
+			"organisationID": central.OrganisationID,
+			"tenantID":       central.ID,
```
0x656b694d, line 29: Nit: central reports properties in Title Case (e.g. "Main Image"), that'll be visible on Amplitude. We may decide to agree on naming convention for consistency. Or not.

internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// NewTelemetry creates a new telemetry service instance.
+func NewTelemetry(config *telemetry.TelemetryConfig) *Telemetry {
+	return &Telemetry{config: config}
+}
+
+// RegisterTenant emits a group event that captures meta data of the input central instance.
+func (t *Telemetry) RegisterTenant(central *dbapi.CentralRequest) {
+	if t != nil && t.config.Enabled() {
```
0x656b694d, line 20:
```suggestion
	if t == nil || t.config == nil || !t.config.Enabled() {
```

internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -0,0 +1,46 @@
...
+		return
+	}
+	if telemeter := t.config.Telemeter(); telemeter != nil {
```
0x656b694d, line 23: `telemeter` should normally be not nil here, not need to check. Anyway `Group()` called on `nil` is no-op.

pkg/client/telemetry/telemetry.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// TelemetryConfig is a wrapper for the telemetry configuration.
+type TelemetryConfig struct {
+	phonehome.Config
+}
+
+// NewTelemetryConfig creates a new telemetry configuration.
+func NewTelemetryConfig() *TelemetryConfig {
+	return &TelemetryConfig{}
```
0x656b694d, line 16: Nit:
This would add some context to all messages issued from this client:
```suggestion
	return &TelemetryConfig{
		ClientID:   getFleetManagerID(),
		ClientName: "ACS Fleet Manager",
	}
```

internal/dinosaur/pkg/services/telemetry.go
stehessel: It can be nil, at least I got some nil-pointer dereference in some tests. I think the `nilTelemeter` does not actually allow to call `.Stop()` on it.

pkg/client/telemetry/telemetry.go
stehessel: Do we need an `Identify` call as well for fleet manager?

0x656b694d (commented):
May we also Track an event like "Central Creation Requested" (with proper wording), to support tracking of the cluster creation journey?

pkg/client/telemetry/telemetry.go
0x656b694d: If only there are any fleet manager specific properties (like version) that could be of our interest. Note that any Identity properties will have effect only on the following Track events on Amplitude.

internal/dinosaur/pkg/services/telemetry.go
0x656b694d: That should have been due to nil `config`. See [the previous comment](https://github.com/stackrox/acs-fleet-manager/pull/733#discussion_r1073379187)

internal/dinosaur/pkg/services/telemetry.go
stehessel: `t.config.Enabled()` should already return `false` if `t.config` is `nil`. But indeed the conditions got messed up when I refactored the conditions.

pkg/client/telemetry/telemetry.go
stehessel: We don't expose the build info in fleet manager at the moment. Let's add this later on if need be.

internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -0,0 +1,54 @@
...
+		"Instance Type":   central.InstanceType,
+		"Organisation ID": central.OrganisationID,
+		"Region":          central.Region,
+		"Tenant ID":       central.ID,
+	}
+	t.config.Telemeter().Group(central.ID, central.ID, props)
+}
+
+func (t *Telemetry) TrackInstanceCreation(central *dbapi.CentralRequest, error string) {
+	if t == nil || t.config == nil || !t.config.Enabled() {
```
0x656b694d, line 35: What about such helper?
```
func (t *Telemetry) enabled() bool {
	return t != nil && t.config != nil && t.config.Enabled()
}
```

pkg/client/telemetry/telemetry.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// TelemetryConfig is a wrapper for the telemetry configuration.
+type TelemetryConfig struct {
+	phonehome.Config
+}
+
+// NewTelemetryConfig creates a new telemetry configuration.
+func NewTelemetryConfig() *TelemetryConfig {
+	clientID := getEnv("HOSTNAME", "fleet-manager")
```
dhaus67, line 18: When deploying the fleet manager, what will be the equivalent of the `HOSTNAME` variable set?

Do we want this to be something more deterministic (e.g. always defaulting to fleet-manager), or something like the pod name?

dhaus67 (commented):
Do we also want to track the deprovisioning of central instances?
While the original ask is about "organization customer experience" within ACSCS, it might be worthwhile to add the deletion as well?

pkg/client/telemetry/telemetry.go
stehessel: On K8s `HOSTNAME` is set to the pod name. The default is for local execution.

0x656b694d (approved):

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -235,12 +235,12 @@ func (k *dinosaurService) RegisterDinosaurJob(dinosaurRequest *dbapi.CentralRequ
 
 	if hasCapacity, err := k.HasAvailableCapacityInRegion(dinosaurRequest); err != nil {
 		errorMsg := "failed to create central request"
-		k.telemetry.TrackInstanceRequested(dinosaurRequest, errorMsg)
+		k.telemetry.TrackCreationRequested(dinosaurRequest, errorMsg)
```
0x656b694d, line 238: Can we move the tracking upstack to the request handlers? Same for the deletion request.

stehessel: Yeah I think so, let me see.

stehessel:
Moved the telemetry calls up to the request handlers, and rebased onto current master.

internal/dinosaur/pkg/handlers/dinosaur.go
```diff
@@ -79,9 +82,11 @@ func (h dinosaurHandler) Create(w http.ResponseWriter, r *http.Request) {
 		},
 		Action: func() (interface{}, *errors.ServiceError) {
 			svcErr := h.service.RegisterDinosaurJob(convDinosaur)
+			h.telemetry.TrackCreationRequested(convDinosaur.OrganisationID, svcErr.Error())
```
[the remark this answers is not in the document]
stehessel: Ah good point, totally forgot about the admin API.

internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -0,0 +1,67 @@
...
+		t.config.Telemeter().Group(central.ID, central.ID, props)
+	}
+}
+
+// TrackCreationRequested emits a track event that signals the creation request of a Central instance.
+func (t *Telemetry) TrackCreationRequested(orgID string, error string) {
+	if t.enabled() {
+		props := map[string]any{
+			"Error":   error,
+			"Success": error == "",
```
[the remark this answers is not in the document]
stehessel: This will lead to a nil pointer dereference if `err == nil`. But we can do this with an additional nil check.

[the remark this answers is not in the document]
0x656b694d: I hoped anybody in a sane mind would implement an error interface allowing for calling on nil pointer. I noticed you fixed this problem in this repository for some errors, thank you for that.
Up to you with this change.

dhaus67 (approved):

PR #734, opened 2023-01-16
Increase OCM token margin for refresh

dhaus67:
## Description

We have seen in the probe service an error during provisioning:
```
Error: ... retrieving access token via OCM auth type: invalid_grant: Invalid refresh token
```

The OCM offline token was valid, so the assumption is currently the following:
In case the created access token (which expires after 15 minutes) is sligthly before hitting the 1 minute margin of refresh _and_ the request takes a long amount of time to be processed (e.g. retrying due to network timeouts, latency), it might be that the token will be invalid once the request reaches the 
server.

While this may be an unusual event with low probablility, it doesn't hurt to increase the margin to 5 minutes, as the token itself is valid for 15 minutes.

### DIFF
--- a/pkg/client/fleetmanager/auth_ocm.go
+++ b/pkg/client/fleetmanager/auth_ocm.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	ocmTokenExpirationMargin = time.Minute
+	ocmTokenExpirationMargin = 5 * time.Minute
 	ocmClientID              = "cloud-services"
 	ocmAuthName              = "OCM"
 )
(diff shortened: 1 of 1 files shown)


dhaus67:
* main
  * **PR #734** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/734" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/734?utm_source=stack-comment).

stehessel (approved):
Might not be the problem, but can't hurt

PR #740, opened 2023-01-17
Revert "Revert "ROX-12344: Add organisation name to central request""

dhaus67:
Reverts stackrox/acs-fleet-manager#736

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -481,7 +481,7 @@ db/start:
 .PHONY: db/start
 
 db/migrate:
-	OCM_ENV=integration $(GO) run ./cmd/fleet-manager migrate
+	$(GO) run ./cmd/fleet-manager migrate
 .PHONY: db/migrate
 
 db/teardown:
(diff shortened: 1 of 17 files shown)


stehessel:
Plan discussed with @mtodor @dhaus67 and @ivan-degtiarenko:

1. Execute migration rollback on stage fleet-manager to remove database `organisation_name` column added in https://github.com/stackrox/acs-fleet-manager/pull/736. Ideally such that the rollback is idempotent.
2. Re-migrate the database schema to add the `organisation_column` again.

The whole issue came up because of the need to cherry-pick a commit *after* the initial database migration had already been performed on stage - with the intention to delay the same migration for prod until a later release. We weren't sure how to handle database rollbacks correctly, so this is considered an exercise to practice the migrations, even though the added column should not cause a problem even without schema rollback/re-migration dance.

stehessel (approved):

stehessel:
After unsuccessful attempt to perform rollback migration, we decided to improve the migration tooling. Merging this to unblock further work, the rollback should be optional here.

PR #745, opened 2023-01-21
Make org name migration more robust

stehessel:
## Description
This makes the org name migration more robust, in particular when run in dev environments.

* Don't try to parse ocm files if mock is enabled.
* Check for presence of column before attempting to create/delete it, to make the migration idempotent if an error occurs.
* Only build ocm client if records actually need to be migrated.

- [ ] ~CI and all relevant tests are passing~

## Test manual

Error scenarios that might occur during migrations, like missing credentials.

### DIFF
--- a/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
+++ b/internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
@@ -11,18 +11,17 @@ import (
 )
 
 func newAMSClient(ocmConfig *ocm.OCMConfig) (ocm.Client, error) {
-	err := ocmConfig.ReadFiles()
-	if err != nil {
+	if ocmConfig.EnableMock {
+		return nil, nil
+	}
+	if err := ocmConfig.ReadFiles(); err != nil {
 		return nil, errors.Wrap(err, "reading OCM config files")
 	}
-	if !ocmConfig.EnableMock {
-		conn, _, err := ocm.NewOCMConnection(ocmConfig, ocmConfig.AmsURL)
-		if err != nil {
-			return nil, errors.Wrap(err, "creating OCM connection")
-		}
-		return ocm.NewClient(conn), nil
+	conn, _, err := ocm.NewOCMConnection(ocmConfig, ocmConfig.AmsURL)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
```diff
@@ -80,16 +79,14 @@ func addOrganisationNameToCentralRequest(ocmConfig *ocm.OCMConfig) *gormigrate.M
...
 		ID: id,
 		Migrate: func(tx *gorm.DB) error {
-			if err := tx.Migrator().AddColumn(&CentralRequest{}, "OrganisationName"); err != nil {
-				return errors.Wrapf(err, "adding column OrganisationName in migration %s", id)
-			}
-
-			amsClient, err := newAMSClient(ocmConfig)
-			if err != nil {
-				return errors.Wrapf(err, "creating AMS client in migration %s", id)
+			if !tx.Migrator().HasColumn(&CentralRequest{}, colName) {
```
SimonBaeumer, line 86: I searched too long for this method on Friday afternoon 🤦

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
```diff
@@ -119,8 +125,10 @@ func addOrganisationNameToCentralRequest(ocmConfig *ocm.OCMConfig) *gormigrate.M
 			return nil
 		},
 		Rollback: func(tx *gorm.DB) error {
-			if err := tx.Migrator().DropColumn(&CentralRequest{}, "OrganisationName"); err != nil {
-				return errors.Wrapf(err, "rolling back column OrganisationName in migration %s", id)
+			if tx.Migrator().HasColumn(&CentralRequest{}, colName) {
```
SimonBaeumer, line 128: Should we delete the drop column altogether as it is never applied?

SimonBaeumer (approved):

internal/dinosaur/pkg/migrations/202212150000_add_organisation_name.go
stehessel: It would be applied when one calls `fleet-manager migrate rollback-{all,last}` and the column in still in the schema.

PR #748, opened 2023-01-23
ROX-13666: Create non-privileged RDS DB user

vladbologa:
## Description

At the moment, ACSCS Central instances connect to their databases using an account with `rds_superuser` privileges. While this is not a root account (the `rdsadmin`'s account credentials are not shared by AWS), it's good practice to create an account that has a least privileges required for operation.

Existing Centrals retrieve the Central-DB password from the `central-db-password` secret. This PR adds an extra annotation to this secret (`"platform.stackrox.io/user-type"`), which it then uses to determine what kind of password is stored there (the master password, or the non-privileged user password).

If the secret contains the master password (which is the case for new Centrals, or pre-existing Centrals that were created by a previous version of `fleetshard-sync`), then a database initialisation is needed, which:
- creates the non-privileged DB user
- creates the `central_active` DB and creates the `pg_stat_statements` extension into it (this is something that Central is normally doing, but it can't with a non superuser Postgres account)

The initialisation is designed to not fail if some parts have already been performed. Once it is completed, the password of the `rds_superuser` account will be replaced, and **won't be stored anywhere**. If we ever need it again, it can be recreated using the AWS RDS API, provided the necessary credentials are available.

### New Central provisioning
Tested locally by verifying that a DB was provisioned, and that the Central was able to connect:
`./scripts/create-central.sh`

I also checked the Central logs for DB related errors.

### Existing Centrals upgrade
Using the steps above, I created a central from the `main` branch.
Afterwards, I checked out the branch of this PR and executed `DOCKER_CONFIG=~/.docker make deploy/dev-fast`. This rebuilds and restarts `fleet-manager` and `fleetshard-sync`.

After that I verified that the user by created (inspected the `central-db-password`, and verified that the DB user exists, and that the Central is running correctly).

```
# To run tests locally run:
```


fleetshard/pkg/central/postgres/dbinit.go
```diff
@@ -0,0 +1,152 @@
...
+
+func installExtensions(ctx context.Context, con DBConnection) error {
+	db, err := sql.Open("postgres", con.AsConnectionStringWithPassword())
+	if err != nil {
+		return fmt.Errorf("opening DB: %w", err)
+	}
+
+	defer db.Close()
+
+	_, err = db.ExecContext(ctx, "CREATE EXTENSION IF NOT EXISTS pg_stat_statements")
```
vladbologa, line 146: With a non-privileged account, Central tries to install the `pg_stat_statements` extension but fails.

As a workaround, I'm creating the `central_active` DB right after AWS provisions the DB (see the function above), and I'm using the `rds_superuser` account to install the extension.

We generally don't want to use to use `rds_superuser`, as it's bad practice. 

@dashrews78 @c-du do you have any objections to this workaround? One problem is that hardcoding the DB name (central_active) is fragile.

fleetshard/pkg/central/postgres/dbinit.go
```diff
@@ -0,0 +1,152 @@
...
+		case nil:
+			err = tx.Commit()
+		default:
+			tx.Rollback()
+		}
+	}()
+
+	// Central needs to be able to create databases
+	_, err = tx.ExecContext(ctx, "CREATE USER "+userName+
+		" WITH NOSUPERUSER INHERIT CREATEDB NOREPLICATION VALID UNTIL 'infinity' PASSWORD '"+userPassword+"'")
```
vladbologa, line 86: @dashrews78 @c-du Does Central still need `CREATEDB`, if we create `central_active` before providing the DB to Central?

From glancing at the source code, I think it might still need to create `central_backup` and `central_restore` DBs. Are those actually created in ACSCS? AWS is also taking daily snapshots of our DBs, btw.

dashrews78: It still needs it for upgrade/rollback.  in an upgrade it will create things like `central_previous` and `temp` potentially.  Backup and restore are currently hiddeen for ACSCS

fleetshard/pkg/central/postgres/dbinit.go
dashrews78: That particular extension is at the server level and reports on all databases.  So I think if you installed that on the `postgres` database before creating the database and/or `central_active` that may solve the problem.  It will exist when central tries to mess with it upon getting the connection and thus not try to re-enable it.  In theory.

vladbologa: I tried installing it on `postgres`, but Central was still failing when creating the extension. The way I could get it work was to create `central_active` (with Central as owner) and then installing the extension into `central_active` using the superuser account.

fleetshard/pkg/central/postgres/dbinit.go
dashrews78: If that is something we want to handle differently for ACSCS in the future we can probably explore that, but for now we need that permission

fleetshard/pkg/central/postgres/dbinit.go
dashrews78: we may need a ticket to look at that from a central code perspective on when we do that extension install.  This could cause issues if customers try to us any external installation.

vladbologa: It would be best if Central does not attempt to install extensions in ACSCS at all. For `pg_stat_statements` we're good (due to `IF NOT EXISTS`, which is also used by Central). But if other extensions are added, they would cause failures in Central (unless they are added via `fleetshard-sync` as well).

dashrews78: This might require more thought and coordination.  We will have to think through what to do if we need a new extension and it is not installed in the target database for managed services and potentially for other external databases.  We are also going to have to investigate and analyze if we have access to the data in the extension if it was installed more globally.

vladbologa: I did some more tests, but it seems a user really needs to be a member of the `rds_superuser` role in order to create extensions, except for a small subset of "trusted" extensions:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.Extensions.Trusted

Even if we were to continue using `rds_superuser` in ACSCS, it's still not a guarantee that an extension created by Central on vanilla Postgres would work on RDS. Only some extensions are supported:
https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html#postgresql-extensions-13x

Regarding your last point, I found this in the [Postgres docs](https://www.postgresql.org/docs/current/pgstatstatements.html):
> When pg_stat_statements is active, it tracks statistics across all databases of the server. To access and manipulate these statistics, the module provides views pg_stat_statements and pg_stat_statements_info, and the utility functions pg_stat_statements_reset and pg_stat_statements. These are not available globally but can be enabled for a specific database with CREATE EXTENSION pg_stat_statements.

dashrews78: I took a quick look, and best I can tell we currently are not using those views.  I'm sure there is useful info we may want to include as metrics at some point but until then one could make the argument that we could eliminate that for ACSCS or perhaps put it behind a debug type flag in general.  @c-du @connorgorman thoughts?

vladbologa: Even if we remove this particular extension creation from Central, we should still decide who's responsible for creating them in general.

I'd argue that on ACSCS the responsibility of installing extensions should fall on `fleetshard-sync`, because it is the one that provisions and manages the DB, and has access to the superuser account. So I don't think it's wrong that `fleetshard-sync` creates the `central_active` DB and installs extensions. This approach would fit also with multi-tenancy, where `fleetshard-sync` would create separate accounts in the same DB for each Central.

dashrews78: Not going to disagree with that.  It makes a lot of sense, but I worry about the upgrade case considering how it currently works.  What happens when you copy a database that has extensions enabled that you didn't have permissions to enable?  Restore is different for ACSCS so I'm not too worried about that case.

vladbologa: That's a very good point, I'll have to try that.

dashrews78: yes I 100% agree with that statement.  I'm not entirely sure how to solve that, however, other than "knowing" that if I add an extension, I need to coordinate.  I'm open to suggestions.

vladbologa: An idea would be to have some CI tests in the stackrox repo, that would use a non-root user (ideally a user created using this code from acs-fleet-manager).

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -537,6 +590,48 @@ func (r *CentralReconciler) ensureCentralDBSecretExists(ctx context.Context, rem
...
+	secret.Annotations[dbUserTypeAnnotation] = userType
+
+	err = r.client.Update(ctx, secret)
+	if err != nil {
+		return fmt.Errorf("updating Central DB secret: %w", err)
+	}
+	return nil
+}
+
+func (r *CentralReconciler) noPrivilegedCentralDBUserExists(ctx context.Context, remoteCentralNamespace string) bool {
```
vladbologa, line 614: rename to `nonPrivilegedCentralDBUserExists`

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -45,6 +46,11 @@ const (
 	orgIDLabelKey             = "rhacs.redhat.com/org-id"
 	tenantIDLabelKey          = "rhacs.redhat.com/tenant"
 
+	dbUserTypeAnnotation    = "platform.stackrox.io/user-type"
+	dbUserTypeMaster        = "master"
+	dbUserTypeCentral       = "central"
+	dbNonPrivilegedUserName = "rhacs_central"
```
SimonBaeumer, line 52:
```suggestion
	dbCentralUserName = "rhacs_central"
```

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -537,6 +590,48 @@ func (r *CentralReconciler) ensureCentralDBSecretExists(ctx context.Context, rem
...
+			Name: centralDbSecretName,
+		},
+	}
+
+	err := r.client.Get(ctx, ctrlClient.ObjectKey{Namespace: remoteCentralNamespace, Name: centralDbSecretName}, secret)
+	if err != nil {
+		return false
+	}
+
+	dbUserType, exists := secret.Annotations[dbUserTypeAnnotation]
```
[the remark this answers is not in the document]
vladbologa: To the first question: legacy Centrals are migrated too, see the line below (if the annotation does not exist, it's assumed that this is a master password, and the new user will be created).

To the second question: in theory it's nil-safe (we always add the managed-by annotation), but I should have checked for nil nevertheless.

fleetshard/pkg/central/postgres/dbinit.go
```diff
@@ -0,0 +1,156 @@
...
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/lib/pq"
+)
+
+const centralDBName = "central_active"
```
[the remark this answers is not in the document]
vladbologa: If / when we have multi-tenancy, we'll have to adapt this code.

I suppose in that case we'll have to create a DB here in `fleetshard-sync`, create a user that can *only* access that DB and not the others, and pass it to Central.

But as long as there's no support for that in Central or ACSCS, I'd leave this as is.

fleetshard/pkg/central/postgres/dbconnection.go
```diff
@@ -31,13 +34,32 @@ func NewDBConnection(host string, port int, user, database string) (DBConnection
...
+func (c DBConnection) WithPassword(password string) DBConnection {
+	c.password = password // pragma: allowlist secret
+	return c
+}
+
+// AsConnectionString returns a string that can be used to connect to a PostgreSQL server. The password is omitted.
 func (c DBConnection) AsConnectionString() string {
-	return fmt.Sprintf("host=%s port=%d user=%s dbname=%s sslmode=require",
-		c.host, c.port, c.user, c.database)
+	return fmt.Sprintf("host=%s port=%d user=%s dbname=%s sslmode=%s",
```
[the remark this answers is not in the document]
vladbologa: `AsConnectionStringWithPassword()` is used to connect from `fleetshard-sync`
`AsConnectionString()` is used to create a connection string without the passport, in order to pass it to Central (via the Central CR).

This is why there are two versions of this function. I could unexport `asConnectionStringWithPassword()`, so that it's not mistakenly used outside of this package (and so leaking the password in plaintext in a connection string).

fleetshard/pkg/central/postgres/dbinit.go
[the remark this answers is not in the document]
SimonBaeumer: @vladbologa Wouldn't it just adding the UUID to the database names, i.e. `central_active_<UUID>`? It seems like low effort, saving later lots of migration work, no?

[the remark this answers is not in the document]
vladbologa: This is a very good idea, to simplify future migration to multi-tenancy.

But unfortunately I don't think it's low effort because:
- we need to implement support in Central, as a prerequisite
- we need to migrate the databases of legacy Centrals

Should open a new ticket for this.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -489,44 +482,165 @@ func (r *CentralReconciler) ensureNamespaceDeleted(ctx context.Context, name str
...
+
+	if !centralDBSecretExists {
+		dbMasterPassword, err := random.GenerateString(25, random.AlphanumericCharacters)
+		if err != nil {
+			return fmt.Errorf("generating Central DB master password: %w", err)
+		}
+		if err := r.ensureCentralDBSecretExists(ctx, remoteCentralNamespace, dbUserTypeMaster, dbMasterPassword); err != nil {
+			return fmt.Errorf("ensuring that DB secret exists: %w", err)
+		}
+	}
```
SimonBaeumer, line 523: This logic can be moved into `ensureCentralDBSecretExists`. If the secret does not exist a new secret is created, thus ensured the secret is there.

In L525 is looks like the previously generated password is read from the cluster again. This can also be omitted if `ensureCentralDBSecretExists` guarantees to persist the secret in k8s.

fleetshard/pkg/central/postgres/dbinit.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func initializeNonPrivilegedUser(ctx context.Context, db *sql.DB, userName, userPassword string) error {
+	err := createUser(ctx, db, userName, userPassword)
+	if err != nil {
```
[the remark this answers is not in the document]
vladbologa: Done.

fleetshard/pkg/central/postgres/dbinit.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+func createUser(ctx context.Context, db *sql.DB, userName, userPassword string) error {
```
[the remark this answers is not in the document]
vladbologa: Renamed to `createNonPrivilegedUser`

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -537,6 +590,48 @@ func (r *CentralReconciler) ensureCentralDBSecretExists(ctx context.Context, rem
...
+func (r *CentralReconciler) noPrivilegedCentralDBUserExists(ctx context.Context, remoteCentralNamespace string) bool {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: centralDbSecretName,
+		},
+	}
+
+	err := r.client.Get(ctx, ctrlClient.ObjectKey{Namespace: remoteCentralNamespace, Name: centralDbSecretName}, secret)
+	if err != nil {
+		return false
```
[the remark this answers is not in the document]
vladbologa: Done.

fleetshard/pkg/central/reconciler/reconciler.go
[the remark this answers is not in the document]
vladbologa: Added nil-check.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -499,6 +492,64 @@ func (r *CentralReconciler) ensureNamespaceDeleted(ctx context.Context, name str
...
+	dbConnection, err := r.managedDBProvisioningClient.GetDBConnection(remoteCentral.Id)
+	if err != nil {
+		return "", fmt.Errorf("getting RDS DB connection data: %w", err)
+	}
+	return dbConnection.GetConnectionForUser(dbNonPrivilegedUserName).AsConnectionString(), nil
+}
+
+func (r *CentralReconciler) ensureManagedCentralDBInitialized(ctx context.Context, remoteCentral private.ManagedCentral) error {
+	remoteCentralNamespace := remoteCentral.Metadata.Namespace
+	if err := r.ensureCentralDBSecretExists(ctx, remoteCentralNamespace); err != nil {
```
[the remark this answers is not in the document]
vladbologa: Done.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -499,6 +492,64 @@ func (r *CentralReconciler) ensureNamespaceDeleted(ctx context.Context, name str
...
+	}
+	err = r.managedDBInitFunc(ctx, dbConnection.WithPassword(dbMasterPassword), dbNonPrivilegedUserName, dbCentralPassword)
+	if err != nil {
+		return fmt.Errorf("creating non-privileged DB user: %w", err)
+	}
+
+	// Replace the password stored in the secret. This replaces the master password (the password of the
+	// rds_superuser account) with the password of the non-privileged account. Note that we don't store
+	// the master anywhere from this point on.
+	err = r.updateCentralDBSecret(ctx, remoteCentralNamespace, dbUserTypeCentral, dbCentralPassword)
```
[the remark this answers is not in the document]
vladbologa: Done.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -499,6 +492,64 @@ func (r *CentralReconciler) ensureNamespaceDeleted(ctx context.Context, name str
...
+	if err != nil {
+		return fmt.Errorf("provisioning RDS DB: %w", err)
+	}
+
+	dbConnection, err := r.managedDBProvisioningClient.GetDBConnection(remoteCentral.Id)
+	if err != nil {
+		return fmt.Errorf("getting RDS DB connection data: %w", err)
+	}
+
+	dbCentralPassword, err := random.GenerateString(25, random.AlphanumericCharacters)
```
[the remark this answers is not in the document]
vladbologa: Done, but `random.GenerateString(25, random.AlphanumericCharacters)` is still duplicated (once for master pass, once for Central pass). Should I make this line into a standalone function?

fleetshard/pkg/central/postgres/dbinit.go
```diff
@@ -0,0 +1,156 @@
...
+// - creates the central_active DB and installs extensions
+func InitializeDatabase(ctx context.Context, con DBConnection, userName, userPassword string) error {
+	db, err := sql.Open("postgres", con.AsConnectionStringWithPassword())
+	if err != nil {
+		return fmt.Errorf("opening DB: %w", err)
+	}
+
+	defer db.Close()
+
+	err = initializeNonPrivilegedUser(ctx, db, userName, userPassword)
```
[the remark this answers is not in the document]
vladbologa: Done

fleetshard/pkg/central/reconciler/reconciler.go

fleetshard/pkg/central/reconciler/reconciler.go
vladbologa: Consider the following case:
1. the master password is generated and stored in the secret
2. we use the AWS API to create the DB, with this password
3. this operation takes about ~10 minutes, let's assume that fleetshard restarts or the context expires during this time
4. when we reach this point again, if we generate a new password we won't know the DB password anymore (because we already used the previous password to start the DB provisioning)

If I were to move the logic into `ensureCentralDBSecretExists` then there would be situations where the password that we give as parameter would not be stored, which is not a good API in my opinion.

BTW, there's a difference between the logic for the master pass and the one for the central pass:
- master password is stored **once** before everything else, so we don't lock ourselves out of the DB
- central password is stored after everything is done; if storing fails, we still have the master password and can recreate another central password (so this step can be attempted multiple times)

fleetshard/pkg/central/postgres/dbinit.go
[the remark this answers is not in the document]
dashrews78: FWIW, this is my general thought on how we may handle the multi-tenancy when we get there.

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: That makes sense. I think it is not worth to abstract this function given the use-case here.

fleetshard/pkg/central/reconciler/reconciler.go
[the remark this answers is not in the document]
SimonBaeumer: Yes, please. The duplication here is the domain logic, the generation of the password itself to avoid.

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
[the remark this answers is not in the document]

SimonBaeumer (approved):

PR #763, opened 2023-01-25
ROX-13593: Use `user_id` and `rh-user-id` claims in telemetry events

stehessel:
## Description
Extract the `user_id` or `rh-user-id` (human user token vs service account token) and emit them as part of the telemetry events. Group the telemetry user together with the tenant group.

I tried to clean up the claim naming without changing too much. In particular, I tried to distinguish more clearly between the subject claim, account id claim and the new user id claims. However no changes to the Central request schema have been made.

The plan going forward is to deprecate and eventually remove the `OwnerAccountID` column all together, removing the need for the `account_id` token.

## Test manual

Manual test of the telemetry events.

### DIFF
--- a/internal/dinosaur/pkg/services/telemetry.go
+++ b/internal/dinosaur/pkg/services/telemetry.go
@@ -1,7 +1,12 @@
 package services
 
 import (
+	"context"
+
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/dbapi"
+	"github.com/stackrox/acs-fleet-manager/pkg/auth"
 	"github.com/stackrox/acs-fleet-manager/pkg/client/telemetry"
 )
 
@@ -19,9 +24,27 @@ func (t *Telemetry) enabled() bool {
 	return t != nil && t.config != nil && t.config.Enabled()
 }
 
+func getUserFromContext(ctx context.Context) (string, error) {
+	claims, err := auth.GetClaimsFromContext(ctx)
+	if err != nil {
+		return "", errors.Wrap(err, "cannot obtain claims from context")
+	}
+	user, err := claims.GetUserID()
+	if err != nil {
+		return "", errors.Wrap(err, "cannot obtain user ID from claims")
+	}
+	return user, nil
+}
+
 // RegisterTenant emits a group event that captures meta data of the input central instance.
-func (t *Telemetry) RegisterTenant(central *dbapi.CentralRequest) {
+// Adds the token user to the tenant group.
+func (t *Telemetry) RegisterTenant(ctx context.Context, central *dbapi.CentralRequest) {
 	if t.enabled() {
+		user, err := getUserFromContext(ctx)
+		if err != nil {
+			glog.Warning(errors.Wrap(err, "cannot get telemetry user from context claims"))
+			return
+		}
 		props := map[string]any{
 			"Cloud Account":   central.CloudAccountID,
 			"Cloud Provider":  central.CloudProvider,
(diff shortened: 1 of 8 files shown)


internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -30,19 +53,26 @@ func (t *Telemetry) RegisterTenant(central *dbapi.CentralRequest) {
 			"Region":          central.Region,
 			"Tenant ID":       central.ID,
 		}
-		t.config.Telemeter().Group(central.ID, central.ID, props)
+		t.config.Telemeter().Group(central.ID, user, props)
 	}
```
janisz, line 57: nit: we can save indentation

```suggestion
	if !t.enabled() {
		return
	}
	user, err := getUserFromContext(ctx)
        if err != nil {
	        glog.Warning(errors.Wrap(err, "cannot get telemetry user from context claims"))
	        return
        }
        props := map[string]any{
	        "Cloud Account":   central.CloudAccountID,
	        "Cloud Provider":  central.CloudProvider,
	        "Instance Type":   central.InstanceType,
	        "Organisation ID": central.OrganisationID,
	        "Region":          central.Region,
	        "Tenant ID":       central.ID,
        }
        t.config.Telemeter().Group(central.ID, user, props)
```

internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -52,14 +82,21 @@ func (t *Telemetry) TrackCreationRequested(id string, user string, isAdmin bool,
 }
 
 // TrackDeletionRequested emits a track event that signals the deletion request of a Central instance.
-func (t *Telemetry) TrackDeletionRequested(id string, user string, isAdmin bool, err error) {
+func (t *Telemetry) TrackDeletionRequested(ctx context.Context, tenantID string, isAdmin bool, requestErr error) {
 	if t.enabled() {
```
janisz, line 86: ditto

pkg/auth/acs_claims.go
```diff
@@ -30,10 +30,23 @@ func (c *ACSClaims) GetUsername() (string, error) {
 
 // GetAccountID ...
 func (c *ACSClaims) GetAccountID() (string, error) {
-	if accountID, ok := (*c)[tenantUserIDClaim].(string); ok {
+	if accountID, ok := (*c)[tenantAccountIDClaim].(string); ok {
```
janisz, line 33: I think I've never seen this construction. Maybe if we don't use pointer receiver functions it will be easier to read

pkg/auth/acs_claims.go
```diff
@@ -49,8 +62,8 @@ func (c *ACSClaims) GetOrgID() (string, error) {
 		tenantIDClaim, alternateTenantIDClaim)
 }
 
-// GetUserID ...
-func (c *ACSClaims) GetUserID() (string, error) {
+// GetSubject returns the subject claim of the token. It identifies the principal authenticated by the token.
+func (c *ACSClaims) GetSubject() (string, error) {
```
janisz, line 66: Probably it's too late for this change but it will be great if we have types for Subject and ID instead of strings.

janisz (approved):

pkg/auth/acs_claims.go
stehessel: Might be a bit unusual, but I think it's better to be consistent with the other usages in the existing code here.

internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -30,19 +53,26 @@ func (t *Telemetry) RegisterTenant(central *dbapi.CentralRequest) {
...
 		var errMsg string
+		if requestErr != nil {
+			errMsg = requestErr.Error()
+		}
+
+		user, err := getUserFromContext(ctx)
 		if err != nil {
-			errMsg = err.Error()
+			glog.Warning(errors.Wrap(err, "cannot get telemetry user from context claims"))
```
dhaus67, line 70: `glog.Warning` or `glog.Error`?

internal/dinosaur/pkg/services/telemetry.go
```diff
@@ -52,14 +82,21 @@ func (t *Telemetry) TrackCreationRequested(id string, user string, isAdmin bool,
...
```
dhaus67, line 94: Same notion as above.

dhaus67 (approved):

pkg/auth/acs_claims.go
stehessel: Indeed, I think this would be a bigger refactoring that is out of scope here.

PR #764, opened 2023-01-26
ROX-13924: provision failover RDS instance

vladbologa:
## Description

For high availability, a failover RDS DB instance is created, in a different Availability Zone (AZ).

This PR implements provisioning logic for the failover instance (also for existing Centrals), and makes sure that both instances are deleted when deprovisioning an ACSCS instance.

Note that during ACSCS provisioning, `fleetshard` does not wait for the failover instance to be completely created, as that would increase our provisioning time.

~~- [ ] Evaluated and added CHANGELOG.md entry if required~~

## Test manual

Tested in a local cluster. Verified in the AWS console that the DB was provisioned and that 2 instances were created, and that the Central was able to connect:
`./scripts/create-central.sh`

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/central/cloudprovider/awsclient/rds.go
+++ b/fleetshard/pkg/central/cloudprovider/awsclient/rds.go
@@ -27,6 +27,7 @@ const (
 	dbUser           = "rhacs_master"
 	dbPrefix         = "rhacs-"
 	dbInstanceSuffix = "-db-instance"
+	dbFailoverSuffix = "-db-failover"
 	dbClusterSuffix  = "-db-cluster"
 	awsRetrySeconds  = 30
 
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/central/cloudprovider/awsclient/rds.go
```diff
@@ -57,61 +58,39 @@ type RDS struct {
...
 		return fmt.Errorf("ensuring DB cluster %s exists: %w", clusterID, err)
 	}
 
+	instanceID := getInstanceID(databaseID)
 	if err := r.ensureDBInstanceCreated(instanceID, clusterID); err != nil {
 		return fmt.Errorf("ensuring DB instance %s exists in cluster %s: %w", instanceID, clusterID, err)
 	}
 
+	failoverID := getFailoverInstanceID(databaseID)
+	if err := r.ensureDBInstanceCreated(failoverID, clusterID); err != nil {
```
kovayur, line 71: I'm not 100 sure if it's a blocking operation, but if it is we may want to execute it in a separate goroutine to add some parallelism. Just food of thought for further improvements.

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
```diff
@@ -57,61 +58,39 @@ type RDS struct {
...
+		return fmt.Errorf("ensuring failover DB instance %s exists in cluster %s: %w", failoverID, clusterID, err)
+	}
+
 	return r.waitForInstanceToBeAvailable(ctx, instanceID, clusterID)
```
kovayur, line 75: Don't we need to wait for a failover instance to be available too?

kovayur (approved):

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
vladbologa: It's not blocking, they go in parallel.

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
vladbologa: We don't _need_ to, because Central only requires the primary one to start (the first one we create is a read/write instance, the second one is a read only instance that is not used except as failover, because Central doesn't support RO instances).

There's the very unlikely scenario that the primary one would fail in those few minutes before the failover one is available, but:
- I'd rather keep the provisioning time shorter, and
- there's not much to recover in fresh DB anyway :)

PR #786, opened 2023-02-02
update observability operator to v4.0.4

stehessel:
## Description

The update is only done in stage to make sure the update is safe.

The new version includes some nice features:
- version bumps of golang and bundled operators. This should fix some security issues.
- self-reported metrics, which allows us to define alerts on reconciliation failures of the observability operator.
- removed "Kafka" from default resources names.

The latter could break some of our alerting rules and documented links. However, it should only come into effect once we delete the existing "Kafka" resources in the cluster and re-create them with the new default. We can do this at a later stage.

## Test manual

Tested this version on a dev cluster, didn't notice any problems.


dhaus67 (approved):
b9c

PR #789, opened 2023-02-03
Add namespace to kubectl calls

SimonBaeumer:

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,7 @@ DOCKER ?= docker
 DOCKER_CONFIG ?= "${PWD}/.docker"
 
 # Default Variables
+ACSMS_NAMESPACE ?= acsms
 ENABLE_OCM_MOCK ?= false
 OCM_MOCK_MODE ?= emulate-server
 JWKS_URL ?= "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/certs"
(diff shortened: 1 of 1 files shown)


kurlov (approved):

Makefile
```diff
@@ -72,6 +72,7 @@ DOCKER ?= docker
 DOCKER_CONFIG ?= "${PWD}/.docker"
 
 # Default Variables
+ACSMS_NAMESPACE ?= acsms
```
vladbologa, line 75: this env var is already defined in `dev/env/scripts/lib.sh:84`, can we get it from there?

vladbologa (changes requested):

Makefile
SimonBaeumer: Hm, good question. The option I am seeing is moving the default variable from `lib.sh` to the Makefile, or a wrapper-script which sets all environment variables before executing the given script.
Imho setting the namespace here is okay, for the development environment I do not expect this to change - and if it does it does not cause big disruptions.
My suggestion is that the namespace is duplicated in the `Makefile`.

vladbologa: Yes, since it only relates to the dev env it's acceptable to have this duplication.

vladbologa (approved):

PR #804, opened 2023-02-09
Use new telemetry options

stehessel:
## Description
Use the new telemetry options interface to specify the group ID explicitly in track calls. We are hoping this will prevent Segment from wrongly attaching the last created group, which might belong to a different tenant.


0x656b694d (approved):
Thank you!

stehessel:
\retest

stehessel:
* main
  * **PR #805** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/805" target="_blank">Graphite</a> 
    * **PR #804** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/804" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/804?utm_source=stack-comment).

porridge (approved):
I'm probably not the best person to review this, but I see no issues :-)

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -232,7 +232,6 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 		}
 
 		central.Spec.Central.DB = &v1alpha1.CentralDBSpec{
-			IsEnabled:                v1alpha1.CentralDBEnabledPtr(v1alpha1.CentralDBEnabledTrue),
```
vladbologa, line 235: I have to look a bit what happened here, as I believe this might switch back to RocksDB in Prod.

vladbologa (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
porridge: Good point @vladbologa, we need to be careful to only use a recent operator (at or past https://github.com/stackrox/stackrox/commit/b3dac4224695eae8fd99f17e49dba9b6e08a75ea) before we deploy this! Perhaps instead it would be safer to leave this in for the time being until the dust has settled.

vladbologa: So if I understand correctly what's happening here, this is bringing in code that is post v3.74 (so it's code that will be in 4.0).

Meanwhile prod is in v3.73, which needs `IsEnabled` to be set to true. So does v3.74.

vladbologa: So I think `CentralDBEnabledPtr` was removed by mistake (cc @kurlov)

Adding it back here https://github.com/stackrox/stackrox/pull/4812

vladbologa: @porridge As long as we still have `IsEnabled` explicitly set to `Enabled`, this should work with older operators too, right?

I believe that if we merge https://github.com/stackrox/stackrox/pull/4812, then update the reference to `stackrox/stackrox` in this PR, and restore the deleted line above, we should be fine.

porridge: I agree with all of the above. :-)

vladbologa (approved):

PR #807, opened 2023-02-10
add group name to traits

stehessel:
## Description
`name` is a reserved trait for the Segment name (see https://segment.com/docs/connections/spec/group/#traits). The hope is that attaching the group name to the traits, will help with the group -> tenant association.

The tenant grouping of the first creation events got broken after #804. In my local testing specifying the group name explicitly seems to help. So with the last change + this, we are at "sometimes the grouping works, sometimes not, but if it works it's the correct group".


stehessel:
was a coincidence, no difference to the race condition

PR #821, opened 2023-02-13
update prod observability operator to v4.0.4

stehessel:
## Description

Update observability operator to `v4.0.4`. This version has been live in stage for some time, and brings security updates and a change to the resource naming (`kafka` -> `obs`).


ivan-degtiarenko (approved):
Looks good. Do we also need to change value here?
https://github.com/stackrox/acs-fleet-manager/blob/d1863a5eeff21c2f9a8c3ab5c5a253b34ce20494/dp-terraform/helm/rhacs-terraform/charts/observability/values.yaml#L7-L7

stehessel:
> Do we also need to change value here?

We don't need to, but probably not a bad idea in case anyone installs the Helm chart outside of `terraform_cluster.sh`.

PR #825, opened 2023-02-16
Use a released operator in stage, to test before ACSCS release.

porridge:
## Description

The nightly build currently deployed to staging points to not-yet-existing docs, so we do not want to push that to production.
So we revert the operator version in staging to the version that we intend to push to production, to be able to validate everything is OK before proceeding with ACS CS release.

Also (no impact) explicitly set `OPERATOR_USE_UPSTREAM="false"` in the `prod` case for now to make the distinction more clear.

I *suppose* the existing centrals in staging might not like this downgrade, but I think they are disposable. I plan to post an announcement about this when merging this change.

- [x] ~Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`~

## Test manual

None.

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
+++ b/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
@@ -35,7 +35,7 @@ case $ENVIRONMENT in
     OBSERVABILITY_OBSERVATORIUM_GATEWAY="https://observatorium-mst.api.stage.openshift.com"
     OBSERVABILITY_OPERATOR_VERSION="v4.0.4"
     OPERATOR_USE_UPSTREAM="true"
-    OPERATOR_VERSION="v3.74.0-nightly-20230213"
+    OPERATOR_VERSION="v3.73.2"
     ;;
 
   prod)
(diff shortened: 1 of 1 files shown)


dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -35,14 +35,15 @@ case $ENVIRONMENT in
     OBSERVABILITY_OBSERVATORIUM_GATEWAY="https://observatorium-mst.api.stage.openshift.com"
     OBSERVABILITY_OPERATOR_VERSION="v4.0.4"
     OPERATOR_USE_UPSTREAM="true"
```
stehessel, line 37: Shouldn't this also be `false` then?

stehessel (approved):
What tests are you looking to perform? Just see that nothing crashes? In any case we'd like to revert to the nightly version asap to not block the telemetry efforts.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: I thought we wanted to use upstream version also in prod? @SimonBaeumer ?
If yes, then this should be `true`, and when doing the release we will update the `prod` case to also be true.
If not, then I agree we should use `false` in both places.

PR #827, opened 2023-02-16
ROX-13913: Verify RDS certificates in Postgres connections

vladbologa:
## Description

This PR adds the AWS RDS CA certificates to the fleetshard-sync image, which are then:
* used verify the Postgres connection from fleetshard to RDS
* passed to Central as additionalCAs, so that also Central can verify its Postgres connections

## Test manual

Tested in a local cluster, by running:
INSTALL_OPERATOR=NO MANAGED_DB_ENABLED=TRUE ./dev/env/scripts/up.sh
./scripts/create-central.sh

And then waiting for a Central to start successfully.

```
# To run tests locally run:
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,9 @@ FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS build
 
 ENV GOFLAGS="-mod=mod"
 
+RUN mkdir /rds_ca
+ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem /rds_ca/aws-rds-ca-global-bundle.pem
+
 RUN mkdir /src
 WORKDIR /src
 RUN CGO_ENABLED=0 go install -ldflags "-s -w -extldflags '-static'" github.com/go-delve/delve/cmd/dlv@latest
@@ -16,15 +19,16 @@ FROM build as build-standard
 RUN make binary
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as debug
-ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem /usr/local/share/ca-certificates/aws-rds-ca-global-bundle.pem
 COPY --from=build-debug /go/bin/dlv /src/fleet-manager /src/fleetshard-sync /usr/local/bin/
 COPY --from=build-debug /src /src
+COPY --from=build /rds_ca /usr/local/share/ca-certificates
 EXPOSE 8000
 WORKDIR /
 ENTRYPOINT [ "/usr/local/bin/dlv" , "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "exec", "/usr/local/bin/fleet-manager", "serve"]
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as standard
 COPY --from=build-standard /src/fleet-manager /src/fleetshard-sync /usr/local/bin/
+COPY --from=build /rds_ca /usr/local/share/ca-certificates
 EXPOSE 8000
 WORKDIR /
 ENTRYPOINT ["/usr/local/bin/fleet-manager", "serve"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ FROM build as build-standard
 RUN make binary
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as debug
+ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem /usr/local/share/ca-certificates/aws-rds-ca-global-bundle.pem
 COPY --from=build-debug /go/bin/dlv /src/fleet-manager /src/fleetshard-sync /usr/local/bin/
 COPY --from=build-debug /src /src
 EXPOSE 8000
--- a/fleetshard/pkg/central/postgres/dbconnection.go
+++ b/fleetshard/pkg/central/postgres/dbconnection.go
@@ -16,21 +16,21 @@ type DBConnection struct {
 	password string
 }
 
-var once sync.Once
-var rdsCertificateData []byte
+var (
+	once               sync.Once
+	rdsCertificateData []byte
+)
 
 const (
-	// CentralRDSCertificateBaseName is the name of the additional CA that is passed to Central
-	CentralRDSCertificateBaseName = "rds-ca-bundle"
+	// CentralRDSCACertificateBaseName is the name of the additional CA that is passed to Central
+	CentralRDSCACertificateBaseName = "rds-ca-bundle"
 
 	sslMode = "verify-full"
-
-	caPath = "/usr/local/share/ca-certificates/"
-
-	rdsCertificatePath = caPath + "aws-rds-ca-global-bundle.pem"
-
-	// rdsCertificatePathCentral stores the location where the RDS CA bundle is mounted in the Central image
-	rdsCertificatePathCentral = caPath + "00-" + CentralRDSCertificateBaseName + ".crt"
+	caPath  = "/usr/local/share/ca-certificates/"
+	// rdsCACertificatePath stores the location where the RDS CA bundle is mounted in the fleetshard image
+	rdsCACertificatePath = caPath + "aws-rds-ca-global-bundle.pem"
+	// rdsCACertificatePathCentral stores the location where the RDS CA bundle is mounted in the Central image
+	rdsCACertificatePathCentral = caPath + "00-" + CentralRDSCACertificateBaseName + ".crt"
 )
 
 // NewDBConnection constructs a new DBConnection struct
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -16,6 +16,7 @@ FROM build as build-standard
 RUN make binary
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7 as debug
+ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem /usr/local/share/ca-certificates/aws-rds-ca-global-bundle.pem
```
kovayur, line 19: Either need to copy this line for `standard` or move it to `build` and `COPY --from build <cert-file>` for both `debug` and `standard`.

fleetshard/pkg/central/postgres/dbconnection.go
```diff
@@ -45,16 +56,17 @@ func (c DBConnection) WithPassword(password string) DBConnection {
...
+		c.host, c.port, c.user, c.database, sslMode, rdsCertificatePathCentral)
 }
 
-// asConnectionStringWithPassword returns a string that can be used to connect to a PostgreSQL server. This function
-// exposes the password in plain-text, so it should be used with care.
-func (c DBConnection) asConnectionStringWithPassword() string {
-	return c.AsConnectionString() + fmt.Sprintf(" password=%s", c.password)
+// asConnectionForFleetshard returns a string that can be used by fleetshard to connect to a PostgreSQL server. This function
+// exposes the password in plain-text, so its output should be used with care.
+func (c DBConnection) asConnectionForFleetshard() string {
```
kovayur, line 67: Do we need to specify pem here if we run `update-ca-trust` after adding a certificate in the `Dockerfile`?

kovayur (commented):
Thanks @vladbologa, this looks good. Just a couple of remarks.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -246,6 +247,20 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 				Name: centralDbSecretName,
 			},
 		}
+
+		additionalCa, err := os.ReadFile(postgres.RDSCertificatePath)
```
kovayur, line 251: How about caching it?

fleetshard/pkg/central/postgres/dbconnection.go
vladbologa: I'm not sure, I will try. But I think it's preferable like this, so that only RDS servers are accepted.

fleetshard/pkg/central/reconciler/reconciler.go
vladbologa: Makes sense, when there are many Centrals it'll keep reading this file.

Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS build
 
 ENV GOFLAGS="-mod=mod"
 
+RUN mkdir /rds_ca
+ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem /rds_ca/aws-rds-ca-global-bundle.pem
```
SimonBaeumer, line 6: Does this call fail when a 404 or 500 is returned?

fleetshard/pkg/central/postgres/dbconnection.go
```diff
@@ -14,7 +16,22 @@ type DBConnection struct {
 	password string
 }
 
-const sslMode = "require"
+var once sync.Once
+var rdsCertificateData []byte
```
SimonBaeumer, line 20:
```suggestion
var (
  once sync.Once
  rdsCertificateData []byte
)
```

fleetshard/pkg/central/postgres/dbconnection.go
```diff
@@ -14,7 +16,22 @@ type DBConnection struct {
...
+
+const (
+	// CentralRDSCertificateBaseName is the name of the additional CA that is passed to Central
+	CentralRDSCertificateBaseName = "rds-ca-bundle"
+
+	sslMode = "verify-full"
+
+	caPath = "/usr/local/share/ca-certificates/"
+
+	rdsCertificatePath = caPath + "aws-rds-ca-global-bundle.pem"
```
SimonBaeumer, line 30: Why so many spaces between the constant definitions?

fleetshard/pkg/central/postgres/dbconnection.go
```diff
@@ -45,16 +62,17 @@ func (c DBConnection) WithPassword(password string) DBConnection {
 	return c
 }
 
-// AsConnectionString returns a string that can be used to connect to a PostgreSQL server. The password is omitted.
-func (c DBConnection) AsConnectionString() string {
-	return fmt.Sprintf("host=%s port=%d user=%s dbname=%s sslmode=%s",
-		c.host, c.port, c.user, c.database, sslMode)
+// AsConnectionStringForCentral returns a string that can be used by Central to connect to the PostgreSQL server
+func (c DBConnection) AsConnectionStringForCentral() string {
```
SimonBaeumer, line 66: This is leaky abstraction. The caller of the function is exposed in the method signature. How should someone use this function if they don't use it from Central?
`AsConnectionStringWithCertificateAuth`

fleetshard/pkg/central/postgres/dbconnection.go
```diff
@@ -45,16 +62,17 @@ func (c DBConnection) WithPassword(password string) DBConnection {
...
+		c.host, c.port, c.user, c.database, sslMode, rdsCertificatePathCentral)
 }
 
-// asConnectionStringWithPassword returns a string that can be used to connect to a PostgreSQL server. This function
-// exposes the password in plain-text, so it should be used with care.
-func (c DBConnection) asConnectionStringWithPassword() string {
-	return c.AsConnectionString() + fmt.Sprintf(" password=%s", c.password)
+// asConnectionStringForFleetshard returns a string that can be used by fleetshard to connect to a PostgreSQL server. This function
+// exposes the password in plain-text, so its output should be used with care.
+func (c DBConnection) asConnectionStringForFleetshard() string {
```
SimonBaeumer, line 73:
```suggestion
func (c DBConnection) asConnectionStringWithPasswordAuth() string {
```

fleetshard/pkg/central/postgres/dbconnection.go
SimonBaeumer: Can you add information that this is used to trust AWS RDS instances? (not authentication)

```suggestion
	rdsCertificatePath = caPath + "aws-rds-ca-global-bundle.pem"
```

fleetshard/pkg/central/postgres/dbconnection.go
```diff
@@ -64,3 +82,17 @@ func (c DBConnection) GetConnectionForUser(userName string) DBConnection {
 
 	return nonPrivilegedConnection
 }
+
+// GetRDSCertificate returns the location where the RDS CA bundle is mounted in the fleetshard image
+func GetRDSCertificate() ([]byte, error) {
+	var err error
+	once.Do(func() {
```
SimonBaeumer, line 89: Certificates changing at runtime are not updated anymore and require a Pod restart to be read.
Is this safe to use here?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -246,6 +246,20 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 				Name: centralDbSecretName,
 			},
 		}
+
+		rdsCA, err := postgres.GetRDSCertificate()
+		if err != nil {
+			glog.Warningf("Could not read RDS CA bundle: %v", err)
```
SimonBaeumer, line 252: Can you rename the function t o`GetRDSCACertificates`? The error message is more descriptive than the function name.
If you find a better way to not have so many uppercases it would be cool as in my suggestion😆

SimonBaeumer (changes requested):

fleetshard/pkg/central/postgres/dbconnection.go
SimonBaeumer: not sure if my suggestion is valid go style

fleetshard/pkg/central/postgres/dbconnection.go
vladbologa: "Central" is not the caller of the function. This function generates a connection string that is to be used by Central, and you can't / shouldn't use it for something else (e.g. it includes the path where Central stores its CA certificates in the `sslrootcert` field)

fleetshard/pkg/central/postgres/dbconnection.go
vladbologa: Naming them `AsConnectionStringWithCertificateAuth` / `asConnectionStringWithPasswordAuth` implies to me that one uses passwords, and the other doesn't (which is not true). They both use passwords and certificates, except that in the case of Central we don't include the password (it's shared via a secret), in case of Fleetshard we put it in the connection string (because it's the same process).

fleetshard/pkg/central/postgres/dbconnection.go
vladbologa: These certificates cannot change at runtime, they are included in the image so they can only change when we deploy a new fleetshard image.

Dockerfile
vladbologa: Building of the image fails if this file can't be downloaded.

fleetshard/pkg/central/reconciler/reconciler.go
vladbologa: Oh that's how I also wanted to name it, and changed my mind because of all the caps. 😄

fleetshard/pkg/central/reconciler/reconciler.go

fleetshard/pkg/central/postgres/dbconnection.go
vladbologa: Done

fleetshard/pkg/central/postgres/dbconnection.go
vladbologa: Done

fleetshard/pkg/central/postgres/dbconnection.go
vladbologa: I guess the underlying problem is that I'm hardcoding the path of the `sslrootcert`. If I make the paths configurable I could make these functions more generic like before.

vladbologa: Working on cleaning this up a bit

SimonBaeumer (approved):

PR #840, opened 2023-02-21
Build and push probe separately, and on production branch too.

porridge:
## Description

Problem:
- `acs-fleet-manager` image is built and pushed by jenkins for all `main` and `production` commits.
- probe image is built and pushed by GHA only for `main` commits
- however deployment of both uses the same way to obtain image tag to push
- this obviously leads to image pull failures for prober when cherry-picked commits exist on `production` branch

To fix this quickly without refactoring the whole probe build/push process and repository configuration, this change moves the probe build+push to a separate workflow, and enables this workflow on both `main` and `production`. Once merged and working, this commit will need to be cherry-picked to the `production` branch.

- [x] ~Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`~

## Test manual

N/A

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -106,18 +106,12 @@ jobs:
           export PATH=${PATH}:$GOPATH/bin
           make verify binary test test/integration
         timeout-minutes: 14
-      - name: Build and publish image to quay.io
+      - name: Build and publish fleet* image to quay.io
         if: github.event_name == 'push'
         env:
           QUAY_USER: ${{ secrets.QUAY_RHACS_ENG_FM_RW_USERNAME }}
           QUAY_TOKEN: ${{ secrets.QUAY_RHACS_ENG_FM_RW_PASSWORD }}
           QUAY_IMAGE_REPOSITORY: rhacs-eng/fleet-manager
-
-          QUAY_PROBE_USER: ${{ secrets.QUAY_RHACS_ENG_PROBE_RW_USERNAME }}
-          QUAY_PROBE_TOKEN: ${{ secrets.QUAY_RHACS_ENG_PROBE_RW_PASSWORD }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -106,18 +106,12 @@ jobs:
           export PATH=${PATH}:$GOPATH/bin
           make verify binary test test/integration
         timeout-minutes: 14
-      - name: Build and publish image to quay.io
+      - name: Build and publish fleet* image to quay.io
```
stehessel, line 109: Don't we have the same issue with this fleet* image? There are also file filters in the workflow, so it won't run on every main commit either. Do you know where we use it, if we actually deploy the appSRE Jenkins image?

porridge: I don't know where the `rhacs-eng` fleet* image is being used. Perhaps the CI e2e tests :shrug:

.github/workflows/probe.yaml
```diff
@@ -0,0 +1,38 @@
+name: Build and push probe image.
```
mtodor, line 1: nitpick: I have noticed that other workflow names are without a dot at the end.
```suggestion
name: Build and push probe image
```

mtodor (approved):
There is one case that could be problematic. Since the probe is using fleet-manager API, versions of deployed probe and fleet-manager should be the same.

With this solution, we could run into a problem that deployed probe does not work with deployed fleet manager. Since we are deploying fleet manager defined here: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/acs-fleet-manager/cicd/saas.yaml
And probe from the `production` branch.

Another concern can be that the probe code in the production branch is different from the code in the main branch, and it can happen that probe deployed in production was never tested on staging.

I would suggest creating a follow-up ticket to solve this problem.

kurlov (approved):

PR #847, opened 2023-02-23
ROX-10918: use rhacs label selectors

stehessel:
## Description

We want to get rid of the default `strimzi` selector label, and replace it with `rhacs`. The observability operator allows such customizations via the `Observability` custom resource. We need to make sure the operator does not auto create one with default settings.

The CR is created with the current default settings, except:
* Set the cluster id to the cluster name - it's currently a UUID of the OpenShift cluster. The cluster is set by the operator as an external metric label.
* Disable logging explicitly. This shouldn't make a difference in most cases, because the operator ignores logging if the logging operator already exists in the cluster.
* Overwrite the selector labels and set them to `rhacs`.

## Test manual

Tested the Helm chart on dev cluster.

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-02-no-init.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-02-no-init.yaml
@@ -1,0 +1,6 @@
+# This config map prevents the operator from creating a Observability custom resoruce on start up.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: observability-operator-no-init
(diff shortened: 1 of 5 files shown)


janisz (approved):

stehessel:
So I've been thinking about how to roll out this change, and I've decided to do it in incremental steps. So instead of replacing the label selector, I have added both `rhacs` and `strimzi` as possible values. After this change has been rolled out to stage and production, I'll merge https://github.com/stackrox/rhacs-observability-resources/pull/59. Finally, we can remove the `strimzi` label selector again.

PR #848, opened 2023-02-27
ROX-14696: Add terraforming and probe deployment to acs-stage-eu-01.

ebensh:
## Description
Adds acs-stage-eu-01 terraforming to the CD pipeline.

https://issues.redhat.com/browse/ROX-14696

- [ ] ~~CI and all relevant tests are passing~~

## Test manual

I could manually CD the cluster but I want to ensure adding it to CD is a supported way to do the initial bootstrapping. Per https://github.com/stackrox/acs-fleet-manager/blob/main/.github/workflows/deploy-stage.yaml#L5-L8 that means we need to merge to main and fix it if it's broken.

### DIFF
--- a/.github/workflows/deploy-stage.yaml
+++ b/.github/workflows/deploy-stage.yaml
@@ -7,6 +7,12 @@ on:
     branches:
       - main
 
+env:
+  # Credentials are populated by explicit `configure-aws-credentials` jobs in
+  # the workflow, so loading additional credentials in the terraform_cluster.sh
+  # script is not necessary.
+  AWS_AUTH_HELPER: none
+
 jobs:
   terraform:
     name: Re-terraform stage clusters
@@ -31,15 +37,11 @@ jobs:
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
       - name: Run terraforming on acs-stage-dp-02
         working-directory: ./dp-terraform/helm/rhacs-terraform
-        env:
-          AWS_AUTH_HELPER: none # credentials are populated by the above action
         run: |
           set -euo pipefail
           ./terraform_cluster.sh stage acs-stage-dp-02
       - name: Run terraforming on acs-stage-eu-01
         working-directory: ./dp-terraform/helm/rhacs-terraform
-        env:
-          AWS_AUTH_HELPER: none # credentials are populated by the above action
         run: |
           set -euo pipefail
           ./terraform_cluster.sh stage acs-stage-eu-01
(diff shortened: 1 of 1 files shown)


ebensh:
* main
  * **PR #848** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/848" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/848?utm_source=stack-comment).

.github/workflows/deploy-stage.yaml
```diff
@@ -29,13 +29,20 @@ jobs:
         with:
           aws-region: ${{ secrets.AWS_REGION }}
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
-      - name: Run terraforming on THE stage cluster
+      - name: Run terraforming on acs-stage-dp-02
         working-directory: ./dp-terraform/helm/rhacs-terraform
         env:
           AWS_AUTH_HELPER: none # credentials are populated by the above action
```
porridge, line 35: Perhaps it's possible to move this up a level to prevent repetition? :thinking:

ebensh: I thought about it - my concern was then that if a single cluster terraform fails, all of them would fail. We want to have `set -euo pipefail` so that, if I understand it correctly, individual commands _inside_ the terraform script failing would cause an overall failure; but we don't want the overall command to fail if any individual terraform failed.

What are your thoughts?

ebensh (commented):
I'm concerned about a higher level problem regarding admin password creation that I will address separately in Slack. This can still be reviewed, but I won't merge until that discussion is done too.

porridge:
> I thought about it - my concern was then that if a single cluster terraform fails, all of them would fail.

I think we're talking about different things.
I'm just thinking about moving the `env` block to the parent so that we don't need the same block in 3 places.

.github/workflows/deploy-stage.yaml
```diff
@@ -29,13 +29,20 @@ jobs:
         with:
           aws-region: ${{ secrets.AWS_REGION }}
           role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/github
-      - name: Run terraforming on THE stage cluster
+      - name: Run terraforming on acs-stage-dp-02
```
stehessel, line 32: Maybe use https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs?

porridge: We probably want to run these in sequence and stop after first failure. Not sure if this is possible with matrix.

ebensh:
> I'm just thinking about moving the `env` block to the parent so that we don't need the same block in 3 places.

I've made the change, but it looks less explicit and more magical to me having the env at the top.

PR #871, opened 2023-03-08
ROX-15155: Add ACS Operator install/upgrade from Fleetshard-sync

kurlov:
## Description

Install or upgrade ACS Operator from the Fleetshard-sync service using helm chart. The Operator upgrade runs before the Central instances reconciler loop.

It is implemented in the scope of Canary upgrade for Central instances feature. The next steps are:
- Run multiple ACS Operator version simultaneously
- Manage from Fleet Manager which Operator version should reconcile Central instance

~- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~

## Test manual

First of all, the OLM based ACS Operator should be deleted without removing an existing CRDs
1. List and delete OLM subscription for the ACS Operator

```
kubectl get subscription -n stackrox-operator
kubectl delete subscription <subscription> -n stackrox-operator
```

2. Delete the Operator’s ClusterServiceVersion

```
kubectl get clusterserviceversion -n stackrox-operator
kubectl delete clusterserviceversion rhacs-operator.<version> -n stackrox-operator
```

3. Check that there is no running ACS Operator

```
kubectl get pods -n stackrox-operator
NAME                                                              READY   STATUS      RESTARTS      AGE
```

Then it's time to run and verify
4. Build and deploy

```
make deploy/dev
```

5.  Check that a new ACS Operator pod is running and neccessary CRDs exist

```
kubectl get pods -n stackrox-operator
NAME                                                              READY   STATUS      RESTARTS      AGE
rhacs-operator-controller-manager-66b6557d46-vfmbs                2/2     Running     1 (24h ago)   29h
...

kubectl describe crd securedclusters.platform.stackrox.io
Name:         securedclusters.platform.stackrox.io
Namespace:
Labels:       <none>
Annotations:  controller-gen.kubebuilder.io/version: v0.10.0
API Version:  apiextensions.k8s.io/v1
Kind:         CustomResourceDefinition
....

kubectl describe crd centrals.platform.stackrox.io
Name:         centrals.platform.stackrox.io
Namespace:
Labels:       <none>
Annotations:  controller-gen.kubebuilder.io/version: v0.10.0
API Version:  apiextensions.k8s.io/v1
Kind:         CustomResourceDefinition
...
```


fleetshard/pkg/runtime/runtime.go
```diff
@@ -110,6 +116,13 @@ func (r *Runtime) Start() error {
 		Environment:       r.config.Environment,
 	}
 
+	err := r.upgradeOperator()
```
johannes94, line 119: Do I understand correctly that this should only be executed once during fleetshard-sync startup?

After we merge this the fleetshard-sync in this version will be deployed to stage and it's going to try install the operator this way. Are we at a point where we want to replace the old way of installing the operator, if so we need to:

- Uninstall the old OLM resources on the stage data plane cluster
- Change the terraforming script for data plane so that it won't reinstall the operator through OLM
- Do the same to prod on the next release

If we don't want that yet, one option would be to put this behind a feature flag (e. g. through an ENV variable) and work on improving it until we are at a point where we want to replace the old installation method.

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	u.resourcesChart = charts.MustGetChart("rhacs-operator")
+	objs, err := charts.RenderToObjects(releaseName, operatorNamespace, u.resourcesChart, chartVals)
+	if err != nil {
+		return fmt.Errorf("installing operator chart: %w", err)
+	}
+
+	for _, obj := range objs {
+		if obj.GetNamespace() == "" {
+			obj.SetNamespace(operatorNamespace)
```
johannes94, line 44: I'm not sure about this. I don't think every resource should get a namespace attribute. For example the CRDs that are also part of the referenced helm chart are not namespace scoped.

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+	"helm.sh/helm/v3/pkg/chartutil"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	operatorNamespace = "stackrox-operator"
+	releaseName       = "rhacs-operator"
+	operatorImage     = "quay.io/rhacs-eng/stackrox-operator:3.73.1"
```
johannes94, line 20: In case we bring this to prod, I'd like this to be configurable so that we don't have to build a new fleetshard-sync image if we need to change the operator image.

Also current operator version deployed to prod is `3.74.0` so this should be the default.

johannes94 (changes requested):

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// ACSOperatorManager ...
+type ACSOperatorManager struct {
+	client         ctrlClient.Client
+	resourcesChart *chart.Chart
+}
+
+// Upgrade ...
+func (u *ACSOperatorManager) Upgrade(ctx context.Context) error {
```
SimonBaeumer, line 30: Can you refactor this with Malte's implementation in the reconciler?

fleetshard/pkg/central/operator/upgrade_test.go
```diff
@@ -0,0 +1,114 @@
...
+	assert.NotEmpty(t, CentralCRD.Object["spec"])
+
+	// check ServiceAccount exists
+	err = fakeClient.Get(context.TODO(), client.ObjectKey{Namespace: operatorNamespace, Name: ServiceAccount.GetName()}, ServiceAccount)
+	require.NoError(t, err)
+	assert.Equal(t, k8sAPIVersion, CentralCRD.GetAPIVersion())
+	assert.NotEmpty(t, CentralCRD.Object["metadata"])
+	assert.NotEmpty(t, CentralCRD.Object["spec"])
+
+	// check Operator Deployment exists
```
SimonBaeumer, line 101: Can you validate the image reference in the deployment?
This is missing but crucial for the successful deployment.

SimonBaeumer (changes requested):
Reviewed it, please add a feature flag like @johannes94 suggested.
I am still testing it locally, but need to fix my setup first.

fleetshard/pkg/central/charts/data/rhacs-operator/templates/platform.stackrox.io_centrals.yaml
```diff
@@ -0,0 +1,947 @@
+apiVersion: apiextensions.k8s.io/v1
```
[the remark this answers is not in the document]
kurlov: I think we decided that it's fair trade considering that CRDs don't change often

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-service-account.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+imagePullSecrets:
+  - name: quay-ips
+kind: ServiceAccount
+metadata:
+  name: rhacs-operator-controller-manager
```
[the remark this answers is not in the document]
kurlov: `ServiceAccount` is assigned in the [deployment yaml](https://github.com/stackrox/acs-fleet-manager/blob/c71e5b3947313e5b62add131bda5b341394589ed/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml#L121)

johannes94 (approved):
LGTM, except the RoleBinding, but we can add this if we need it further down the line. I'll approve for now since this is feature flagged anyways.

PR #882, opened 2023-03-14
expose instance type as label on Central pod

stehessel:
The label will be scraped by Prometheus as a source for dashboards. This will help increase the observability around eval and standard instances.


rukletsov (approved):
I wonder if we want to do more than just forwarding fleet manager's `eval` and `standard`. What I think would be useful is actually the relevant SKUs that are associated with customer's account. I know that AMS hides this information on purpose but I actually think this would be helpful for debugging given how many issues we've seen so far.

I wonder if there is a way we can convey the SKUs somehow. Maybe via having different instance count for each SKU and those be power of 2?

stehessel:
> I wonder if there is a way we can convey the SKUs somehow. Maybe via having different instance count for each SKU and those be power of 2?

Can you explain what you mean by this? I'd be careful to add things for debugging that don't generalise to edge cases, e.g. if customers have multiple SKUs that grant ACS quota.

PR #886, opened 2023-03-17
Set ACS Operator CPU request and limit to 2

ebensh:
## Description
Set explicit ACS Operator CPU request and limit to 2 (via OLM config).

- ~~[ ] Added test description under `Test manual`~~
- ~~[ ] CI and all relevant tests are passing~~

## Test manual

Will test in stage using CD to ensure the config is okay, then push to prod to see if it addresses the issue.

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
+++ b/dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
@@ -17,6 +17,13 @@ spec:
   source: {{ .Values.acsOperator.source }}
   sourceNamespace: {{ .Values.acsOperator.sourceNamespace }}
   startingCSV: rhacs-operator.{{ .Values.acsOperator.version }}
+  config:
+    resources:
+      requests:
+        cpu: 2
+      limits:
+        cpu: 2
+
 ---
 {{- if .Values.acsOperator.upstream }}
 apiVersion: operators.coreos.com/v1alpha1


porridge (approved):
I dream of a ticket that we could link to from PR description that would provide the rationale...

vladbologa (commented):
The deployment managed by the operator has 2 containers, each with its own resource settings:
- manager (the actual operator)
- kube-rbac-proxy

I wonder what exactly this config will change. Hopefully just the manager, but I'm not sure.

vladbologa (approved):

stehessel:
Why did you set the request equal to the limit? Isn't increasing the limit enough to avoid throttling?

porridge:
> Why did you set the request equal to the limit? Isn't increasing the limit enough to avoid throttling?

Only if the node has enough resources, I think. Setting the request helps bin packing make sure it does.

stehessel:
> Only if the node has enough resources, I think. Setting the request helps bin packing make sure it does.

Won't a high request just lead to the pod being unscheduled if not enough resources are available - i.e. what seemed to have happened here? Maybe we should assign a higher pod priority to the operator pod to make sure it gets scheduled.

porridge:
> > Only if the node has enough resources, I think. Setting the request helps bin packing make sure it does.
> 
> Won't a high request just lead to the pod being unscheduled

Yes, this is a risk, but I definitely would not use the word "just" :-)
It might be a good compromise if we're reasonably sure CPU spikes will not be coordinated across multiple pods (which is not obvious). However I think unschedulability issues could mean that using smaller `requests` would provide a false sense of security in case of a resource crunch. 

> Maybe we should assign a higher pod priority to the operator pod to make sure it gets scheduled.

I'm not sure we want to preempt other control plane components :thinking:

PR #901, opened 2023-03-22
Minor fixes to script output when using Secrets Manager

kovayur:
## Description
- Fix incorrect backend flag in `osd-cluster-idp-setup.sh`
- This is the correct situation when there are no parameters in each of the stores. A warning should be raised if both are empty.

## Test manual
Tested terraforming locally

### DIFF
--- a/scripts/lib/external_config.sh
+++ b/scripts/lib/external_config.sh
@@ -109,8 +109,7 @@ load_external_config() {
     local parameter_store_output
     local secrets_manager_output
     parameter_store_output=$(run_chamber env "$service")
-    [[ -z "$parameter_store_output" ]] && echo "WARNING: no parameters found under '/$service' in AWS Parameter Store of this environment"
     secrets_manager_output=$(run_chamber env "$service" -b secretsmanager)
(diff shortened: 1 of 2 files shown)


scripts/lib/external_config.sh
```diff
@@ -109,8 +109,7 @@ load_external_config() {
     local parameter_store_output
     local secrets_manager_output
     parameter_store_output=$(run_chamber env "$service")
-    [[ -z "$parameter_store_output" ]] && echo "WARNING: no parameters found under '/$service' in AWS Parameter Store of this environment"
     secrets_manager_output=$(run_chamber env "$service" -b secretsmanager)
-    [[ -z "$secrets_manager_output" ]] && echo "WARNING: no parameters found under '/$service' in AWS Secrets Manager of this environment"
+    [[ -z "$parameter_store_output" && -z "$secrets_manager_output" ]] && echo "WARNING: no parameters found under '/$service' of this environment"
```
vladbologa, line 113: Did I understand correctly that we still use Parameter Store for some non-sensitive params? If that's case shouldn't we have both warnings as before?

vladbologa (approved):

PR #903, opened 2023-03-23
ROX-16108: Remove Targeted Cluster Placement Strategy

ebensh:
## Description
Removes the TargetedClusterPlacementStrategy.

This placement strategy was a breakglass mechanism that was added before in ServicePreview to enable forcing cluster selection to a single specific cluster. It is not currently used and will immediately be replaced by disabling Schedulable field for all non-desired clusters once https://github.com/stackrox/acs-fleet-manager/pull/904 is merged.

## Checklist (Definition of Done)

- ~~[ ] Unit and integration tests added~~
- ~~[ ] Added test description under `Test manual`~~
- ~~[ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~~
- [ ] CI and all relevant tests are passing
- [x] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- ~~[ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.~~

## Test manual

Ran the local tests per instructions below, all passing.

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup OCM_OFFLINE_TOKEN=<ocm-offline-token> OCM_ENV=development
make verify lint binary test test/integration
```

PR #904, opened 2023-03-23
ROX-16108: Replace Skip Scheduling with Schedulable

ebensh:

## Test manual

```
# To run tests locally run:
```


ebensh:
* main
  * **PR #903** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/903" target="_blank">Graphite</a> 
    * **PR #904** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/904" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/904?utm_source=stack-comment).

internal/dinosaur/pkg/services/cluster_placement_strategy.go
```diff
@@ -33,18 +31,23 @@ type FirstReadyPlacementStrategy struct {
 
 // FindCluster ...
 func (d FirstReadyPlacementStrategy) FindCluster(central *dbapi.CentralRequest) (*api.Cluster, error) {
-	clusters, err := d.clusterService.FindAllClusters(FindClusterCriteria{Status: api.ClusterReady})
+	clusters, err := d.clusterService.FindAllClusters(FindClusterCriteria{
```
machi1990, line 34: you could extend the `FindClusterCriteria` to also accept `Schedulable` as the criteria and only filter clusters that are schedulable direct from the DB.
```
Schedulable: true
```

machi1990: Thank you for the reply @ebensh 
> I actually tried this at one point, then reverted 

I am glad this was considered and thanks for the additional details on why it was rejected. 

A few reply inline..

> My reasoning is that it would force a filtering by Schedulable on a function that is fairly general (FindAllClusters), even if you don't care about using it for scheduling purposes

Yes, the function `FindAllClusters()` is general but it accepts a criteria object which is `FindClusterCriteria`. The presence of criteria changes the meaning o the function and it is not "find all clusters" but "find all cluster that matches the criteria".

> I also thought about making the parameters in the FindClusterCriteria each optional and adding a filter for each specified field - this would be reasonable, but seemed like a lot of work for the change I want to make

I do not think this would've been needed, at least not for this change.

> My last consideration was that, in general, the Cluster table is never going to be large. If I have to do some filtering on <10 rows in memory, it's a reasonable thing to do.

This is a fair point and acceptable. It is a tradeoff between an in memory filter or a DB filter. I'd have gone for the later since it is considerable the same amount of work codewise, just because when we scale up (to more than 10 clusters and ooh I really look up for this...;-) ), then perf issue on the cluster placement part is never a big consideration.

internal/dinosaur/pkg/migrations/202303220000_drop_skip_scheduling_from_clusters.go
```diff
@@ -0,0 +1,61 @@
...
+		Region                           string   `json:"region"`
+		Status                           string   `json:"status" gorm:"index"`
+		StatusDetails                    string   `json:"status_details" gorm:"-"`
+		IdentityProviderID               string   `json:"identity_provider_id"`
+		ClusterDNS                       string   `json:"cluster_dns"`
+		ProviderType                     string   `json:"provider_type"`
+		ProviderSpec                     string   `json:"provider_spec"`
+		ClusterSpec                      string   `json:"cluster_spec"`
+		AvailableCentralOperatorVersions api.JSON `json:"available_central_operator_versions"`
+		SupportedInstanceType            string   `json:"supported_instance_type"`
```
machi1990, line 34:
```suggestion
```

These fields are not needed

internal/dinosaur/pkg/migrations/202303230000_add_schedulable_to_clusters.go
```diff
@@ -0,0 +1,61 @@
...
+		Region                           string   `json:"region"`
+		Status                           string   `json:"status" gorm:"index"`
+		StatusDetails                    string   `json:"status_details" gorm:"-"`
+		IdentityProviderID               string   `json:"identity_provider_id"`
+		ClusterDNS                       string   `json:"cluster_dns"`
+		ProviderType                     string   `json:"provider_type"`
+		ProviderSpec                     string   `json:"provider_spec"`
+		ClusterSpec                      string   `json:"cluster_spec"`
+		AvailableCentralOperatorVersions api.JSON `json:"available_central_operator_versions"`
+		SupportedInstanceType            string   `json:"supported_instance_type"`
```
machi1990, line 34:
```suggestion
```
these fields are not needed

internal/dinosaur/pkg/migrations/202303230000_add_schedulable_to_clusters.go
```diff
@@ -0,0 +1,61 @@
...
+		Schedulable                      bool     `json:"schedulable"` // To be added
+	}
+
+	id := "202303231200"
+	colName := "Schedulable"
```
machi1990, line 39:
```suggestion
	colName := "schedulable"
```
I think it has to be lowercase

internal/dinosaur/pkg/migrations/202303220000_drop_skip_scheduling_from_clusters.go
```diff
@@ -0,0 +1,61 @@
...
+		SkipScheduling                   bool     `json:"skip_scheduling" gorm:"default:false"` // To be dropped
+	}
+
+	id := "202303221200"
+	colName := "SkipScheduling"
```
machi1990, line 39:
```suggestion
	colName := "skip_scheduling"
```
?

internal/dinosaur/pkg/services/clusters.go
```diff
@@ -115,6 +113,7 @@ func (c clusterService) Create(cluster *api.Cluster) (*api.Cluster, *apiErrors.S
 	cluster.ClusterID = clusterSpec.InternalID
 	cluster.ExternalID = clusterSpec.ExternalID
 	cluster.Status = clusterSpec.Status
+	cluster.Schedulable = true
```
machi1990, line 116: I understand this as, all the clusters are schedulable by default when they are created. That's the intention write?

internal/dinosaur/pkg/migrations/202303230000_add_schedulable_to_clusters.go
ebensh: The rest of the migrations in the directory have CamelCase style naming based on the Go field - but this may cause confusion if the initial AutoMigrate() call uses snake_case() for all column names. We could use the move to limited availability as a good time to change for consistency.

I haven't changed it here to remain consistent with existing migrations.

internal/dinosaur/pkg/migrations/202303220000_drop_skip_scheduling_from_clusters.go
ebensh: See above, specifically https://github.com/stackrox/acs-fleet-manager/blob/evan%2FROX-16108-remove-skip-scheduling/internal/dinosaur/pkg/migrations/20220916000000_add_skip_scheduling_to_clusters.go#L47-L47

Though now you have me worrying that the AutoMigrate() will create snake_case and the Rollback will actually attempt to drop the CamelCase name, in which case all of these migrations' `Rollback`s are broken.

internal/dinosaur/pkg/services/clusters.go
ebensh: Yes, though this is one part I felt unsure about.

I couldn't find where the Clusters in the DataplaneClusterConfig, which is where we specify all of our clusters, are merged into the actual Clusters service as database entries. I need to look more deeply to understand where that merging happens, when, and how. This would help me understand if the DataplaneClusterConfig is treated as a sort of Upsert on whatever is already in the database.

In terms of the Cluster API, maybe this is wrong, and we should just take the value as it's provided in the input. What wasn't clear to me was where the Cluster is ever actually created or called. We don't have any dynamic cluster creation, we do everything through the DataplaneClusterConfig, and what's important there is that the value set in the config is passed as the Schedulable field of the Cluster.

I'm realizing as I write this that maybe this is going to override the value from the DataplaneClusterConfig? The interaction here isn't clear to me.

I also thought about putting this logic into the database as a default field (like SkipScheduling had) but this felt wrong to me. The database shouldn't be where that logic is located; the database just knows that it can be schedulable or not. The actual Cluster service is where we have that additional logic saying "When we create a new Cluster in the database, it should be Schedulable unless you explicitly tell me it is not." I'm open to suggestions there though.

internal/dinosaur/pkg/migrations/202303230000_add_schedulable_to_clusters.go
ebensh: Done.

internal/dinosaur/pkg/migrations/202303220000_drop_skip_scheduling_from_clusters.go
ebensh: Done.

internal/dinosaur/pkg/services/clusters.go
machi1990: > I couldn't find where the Clusters in the DataplaneClusterConfig, which is where we specify all of our clusters, are merged into the actual Clusters service as database entries. I need to look more deeply to understand where that merging happens, when, and how. This would help me understand if the DataplaneClusterConfig is treated as a sort of Upsert on whatever is already in the database.

This is where this happens; https://github.com/stackrox/acs-fleet-manager/blob/192bf07acdfe01943c9a48a63c504b9f67b76677/internal/dinosaur/pkg/workers/clusters_mgr.go#L649

That's only create the missing entries though i.e a new entry.
For an already existing entry, that can go here: https://github.com/stackrox/acs-fleet-manager/blob/192bf07acdfe01943c9a48a63c504b9f67b76677/internal/dinosaur/pkg/workers/clusters_mgr.go#L673

It is to be noted though that this method call: https://github.com/stackrox/acs-fleet-manager/blob/192bf07acdfe01943c9a48a63c504b9f67b76677/internal/dinosaur/pkg/workers/clusters_mgr.go#L700 won't update the the zero-field of a type. Specifically, when the cluster is marked as not schedulable (i.e set to `false`) then it won't be updated in the database using that method call. This is the inherited gorm behaviour and it is documented in https://github.com/stackrox/acs-fleet-manager/blob/192bf07acdfe01943c9a48a63c504b9f67b76677/internal/dinosaur/pkg/services/clusters.go#L32-L33

To update zero fields, a method that calls or make use of https://gorm.io/docs/update.html#Updates-multiple-columns needs to be implemented. This is what is there for example for the dinosaur.go file: https://github.com/stackrox/acs-fleet-manager/blob/192bf07acdfe01943c9a48a63c504b9f67b76677/internal/dinosaur/pkg/services/dinosaur.go#L91-L94

> In terms of the Cluster API, maybe this is wrong, and we should just take the value as it's provided in the input. What wasn't clear to me was where the Cluster is ever actually created or called. We don't have any dynamic cluster creation, we do everything through the DataplaneClusterConfig, and what's important there is that the value set in the config is passed as the Schedulable field of the Cluster.
> I'm realizing as I write this that maybe this is going to override the value from the DataplaneClusterConfig? The interaction here isn't clear to me.

Yes, this method is only called for dynamically created clusters. If you are only using the manual list, then this won't be needed for now. In my opinion, the change in this line can be removed.

> I also thought about putting this logic into the database as a default field (like SkipScheduling had) but this felt wrong to me. The database shouldn't be where that logic is located; the database just knows that it can be schedulable or not. The actual Cluster service is where we have that additional logic saying "When we create a new Cluster in the database, it should be Schedulable unless you explicitly tell me it is not." I'm open to suggestions there though.

I agree, we should have an explicit applicative logic that controls the default value of the cluster when it is initially inserted. Since this value comes from the configuration, I'm thinking maybe we can use the value set in that config when the cluster is initially inserted over here: https://github.com/stackrox/acs-fleet-manager/blob/192bf07acdfe01943c9a48a63c504b9f67b76677/internal/dinosaur/pkg/workers/clusters_mgr.go#L649

internal/dinosaur/pkg/services/cluster_placement_strategy.go
johannes94: +1 for Evans solution. Performance wise it doesn't really matter and GORM handles default values in SELECT statement the same as for [updates](https://gorm.io/docs/update.html#Updates-multiple-columns), which was already linked in another comment. In other words it wouldn't include a filter for bools set to false when using select with a struct. IMO because of this it is simpler to just have a in memory filter.

johannes94 (approved):

internal/dinosaur/pkg/services/cluster_placement_strategy.go
machi1990: I am fine with the implemented solution, I just wanted to be sure that alternatives have been considered and why they are rejected. 

> Performance wise it doesn't really matter

This is not strictly true: it does matter but it is an acceptable tradeoff for small data - this is not always the case at scale with large data sets. My understanding is that this is an acceptable solution for now because of the scale of data plane that we've, also considering the fact that we've already filtered the data planes per cloud provider and region.

> GORM handles default values in SELECT statement the same as for [updates](https://gorm.io/docs/update.html#Updates-multiple-columns), which was already linked in another comment. In other words it wouldn't include a filter for bools set to false when using select with a struct. IMO because of this it is simpler to just have a in memory filter.

This is true, and there are ways to fix it which I do not think are hard.

johannes94: > This is not strictly true: it does matter but it is an acceptable tradeoff for small data - this is not always the case at scale with large data sets. My understanding is that this is an acceptable solution for now because of the scale of data plane that we've, also considering the fact that we've already filtered the data planes per cloud provider and region.
To be clear, my statement was solely referring to this specific case (currently talking about 2 clusters in ACSCS), it was not meant to be about DB vs in memory filtering in general.

> This is true, and there are ways to fix it which I do not think are hard.
I don't think it's hard either, but I personally prefer the in memory filter in this case.

machi1990: > To be clear, my statement was solely referring to this specific case (currently talking about 2 clusters in ACSCS), it was not meant to be about DB vs in memory filter in general.

Noted. 

> GORM handles default values in SELECT statement the same as for [updates](https://gorm.io/docs/update.html#Updates-multiple-columns), which was already linked in another comment. In other words it wouldn't include a filter for bools set to false when using select with a struct. IMO because of this it is simpler to just have a in memory filter.

> This is true, and there are ways to fix it which I do not think are hard.
> I don't think it's hard either, but I personally prefer the in memory filter in this case.

Also worth noting that, for this specific case, this limitation doesn't apply since we'll always want to place workload on clusters that are schedulable; thus `Schedulable:true`.

machi1990 (approved):
My comments have been addressed/considered, I am approving on the basis of that.
There are related test failures which might need to be addressed.

pkg/api/cluster_types.go
```diff
@@ -154,7 +154,6 @@ type Cluster struct {
 	StatusDetails      string        `json:"status_details" gorm:"-"`
 	IdentityProviderID string        `json:"identity_provider_id"`
 	ClusterDNS         string        `json:"cluster_dns"`
-	SkipScheduling     bool          `json:"skip_scheduling"`
 	// the provider type for the cluster, e.g. OCM, AWS, GCP, Standalone etc
 	ProviderType ClusterProviderType `json:"provider_type"`
 	// store the provider-specific information that can be used to managed the openshift/k8s cluster
```
ludydoo, line 159: not your typo but:
```suggestion
	// store the provider-specific information that can be used to manage the openshift/k8s cluster
```

ebensh:
> My comments have been addressed/considered, thanks: I am approving on the basis of that.
> There are related test failures which might need to be addressed.

Thank you for the review! It really helps to have those questions asked to ensure we're thinking about them and coming to the right solution. And also thanks for the detailed code pointers on the merging of the dataplane cluster configuration, looks like that helped Vlad fix the holes in my PR :)

Thanks Johannes for the review and Vlad for the fixes!

internal/dinosaur/pkg/workers/clusters_mgr.go
```diff
@@ -684,7 +685,7 @@ func (c *ClusterManager) reconcileClusterWithManualConfig() []error {
 		newCluster.ProviderType = manualCluster.ProviderType
 		newCluster.ClusterDNS = manualCluster.ClusterDNS
 		newCluster.SupportedInstanceType = manualCluster.SupportedInstanceType
-		newCluster.SkipScheduling = false
+		newCluster.Schedulable = manualCluster.Schedulable
```
machi1990, line 688: when `manualCluster.Schedulable` is `false` it won't be updated in the database. I mentioned that in the https://github.com/stackrox/acs-fleet-manager/pull/904#discussion_r1148079961 due to a gorm behaviour.
The `clusterService.Update()` has to be revisited to allow updating zero values fields: https://gorm.io/docs/update.html#Updates-multiple-columns.

machi1990: Thank you for double checking. Yes, I was specifically talking about the database not being updated.

vladbologa: According to the gorm docs, using a map instead of a struct would fix this behavior

machi1990: > According to the gorm docs, using a map instead of a struct would fix this behavior

Yes, that's correct. It's also what's used in the `dinosaur.go#Updates()` method.

vladbologa: Maybe I simpler would be to replace `dbConn.Updates(cluster)` with `dbConn.Select("*").Updates(cluster)`. I'll have to check if it actually works.

ebensh: I've pushed a commit to try the values map but haven't had time to test. Will loop back in a bit to test.

vladbologa: Added a new commit with the missing moq file.

Tested your changes locally. Added the cluster then checked the DB:
```
serviceapitests=# select id, schedulable from clusters;
          id          | schedulable
----------------------+-------------
 cgi4eci401g4r7mbniq0 | t
 ```
 
 Then changed `schedulable` to `false` in the dataplane config yaml, restarted fleet-manager and checked the DB again:
 ```
 serviceapitests=# select id, schedulable from clusters;
          id          | schedulable
----------------------+-------------
 cgi4eci401g4r7mbniq0 | f
 ```

machi1990: thanks @ebensh @vladbologa . lgtm

ebensh: Thanks Manyanda! BTW while you're here, do you have a suggestion for replacing the diff call (for golang struct) with something for map of values (or JSON)? The diff library is used only in this one place as far as I can tell, if we wanted we could do the diff on the values map itself (confirm an update is required, then pretty print the diff to the logs).

machi1990: +1 I agree on that suggested approach. Or I think `reflect.DeepEqual` could be used as well.

PR #905, opened 2023-03-23
ROX-15806: Remove pause reconcile annotation before Central deletion

vladbologa:
## Description

If a Central has the pause-reconcile annotation set, and its deletion is requested, it currently remains stuck in a deprovisioning state. The Central itself is not deleted by the Operator, however fleetshard-sync deletes all its other resources, making it unusable.

This PR takes the approach of setting the pause-reconcile annotation to false (if it's present). This way the Central is deleted, and the UI is not in an inconsistent state.

An alternative would be to have also fleetshard-sync not reconcile a Central that has the annotation set.

## Test manual

Tested locally.

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -724,16 +724,10 @@ func (r *CentralReconciler) ensureCentralCRDeleted(ctx context.Context, central 
 		return false, errors.Wrapf(err, "delete central CR %s/%s", central.GetNamespace(), central.GetName())
 	}
 
-	if central.Annotations != nil {
-		// avoid being stuck in a deprovisioning state due to the pause reconcile annotation
-		if _, exists := central.Annotations[pauseReconcileAnnotation]; exists {
-			central.Annotations[pauseReconcileAnnotation] = "false"
-			err = r.client.Update(ctx, central)
-			if err != nil {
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -722,6 +723,19 @@ func (r *CentralReconciler) ensureCentralCRDeleted(ctx context.Context, central
 
 		return false, errors.Wrapf(err, "delete central CR %s/%s", central.GetNamespace(), central.GetName())
 	}
+
+	if central.Annotations != nil {
+		// avoid being stuck in a deprovisioning state due to the pause reconcile annotation
+		if _, exists := central.Annotations[pauseReconcileAnnotation]; exists {
```
kovayur, line 729: Maybe check that it's set to true and not just exist? This will help avoid unnecessary updates in the reconciliation loop before Central is actually deleted.

kovayur (approved):
Looks good, only one small improvement

kovayur:
Unit test?

vladbologa:
> Unit test?

I'm trying to write a test, but it's proving hard to write a meaningful test with the fake client.

`ensureCentralCRDeleted` simply deletes the Central CR in the test environment, so I can't check afterwards if it set the annotation to false.

Any ideas? I could go for a e2e test for example.

PR #907, opened 2023-03-28
ROX-12188: api target central version

johannes94:
## Description
This PR fixes the `/api/rhacs/v1/admin/centrals/{id}` PATCH endpoint to validate and update the central_version for a central if specified in the request body.

## Test manual

```
# To run tests locally run:
```


openapi/fleet-manager-private-admin.yaml
```diff
@@ -435,9 +435,9 @@ components:
             namespace:
               type: string
             central:
-              $ref: "#/components/schemas/CentralSpec"
+              $ref: "fleet-manager.yaml#/components/schemas/CentralSpec"
```
kylape, line 438: Can you explain what this change does?  The `CentralSpec` definitions between the two files doen't seem to be different, but I may not understand some openapi stuff here.

kurlov (approved):

openapi/fleet-manager-private-admin.yaml
johannes94: The reference without this change was wrong because there is no CentralSpec component defined in this File. Somehow the generation of Go Code did still work properly but when using `make run/docs` the swagger UI was throwing some errors because of this reference. Also my IDE was complaining about this unresolved Reference. I still don't understand why the Go code generation was not complaining about this.

internal/dinosaur/pkg/handlers/validation.go
```diff
@@ -257,13 +257,31 @@ func ValidateResourceName(name string) (corev1.ResourceName, bool) {
...
-func ValidateCentralVersion(version *private.CentralDefaultVersion) handlers.Validate {
+// ValidateUpdateCentralVersion validates the CentralVersion field of a CentralUpdateRequest
+func ValidateUpdateCentralVersion(updateReq *private.CentralUpdateRequest) handlers.Validate {
 	return func() *errors.ServiceError {
-		if err := services.ValidateCentralVersionString(version.Version); err != nil {
+		if updateReq.CentralVersion == "" {
+			return nil
+		}
+
+		return validateCentralVersionString(updateReq.CentralVersion)()
```
SimonBaeumer, line 267: This looks not like a validation for the Central Version, it validates a set of registries. I would prefer we call it what it is - a full image reference in this case, and not only the version translated into an image tag.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -320,6 +320,8 @@ func updateCentralRequest(request *dbapi.CentralRequest, updateRequest *private.
 		return fmt.Errorf("updating ScannerSpec within CentralRequest: %w", err)
 	}
 
+	new.DesiredCentralVersion = updateRequest.CentralVersion
```
SimonBaeumer, line 323: Why does the `updateRequest` translate the field into the `DesiredCentralVersion`? 
I expected an API request which sets the `DesiredCentralVersion` directly.

internal/dinosaur/pkg/handlers/validation.go
johannes94: Is you're suggestion to change the wording "version" to "image reference" throughout the whole stack? Because it is probably going to be a full reference everywhere. So we would have to change everything from DB to API to fit that wording. Also I called it generic and not specific to the "set of registries" use case because this way we don't have to change function name in order to implement additional validation if we need to.

PR #913, opened 2023-04-03
ROX-16317: Extract a reusable data plane deployment workflow.

porridge:
## Description

Another copy of the rollout workflow (for the PR check) would be too much.
So I'm extracting the common code into a reusable workflow.

One unclear thing is whether we should be setting `concurrency` in the reusable or in the caller workflow.
One data point is that the linter complained about using `${{ inputs... }}` in the reusable workflow name, but does not complain when I use it in the `concurrency` field, so I'm hopeful it will work as intended.

Also adding a GHA linter workflow while at it, stolen from `stackrox/stackrox` but using `latest` rather than a pinned version, and running only when files in the `.github` dir change.

## Test manual

Ran actionlint.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,24 @@
+name: Lint GitHub Actions workflows
+on:
+  push:
+    branches:
+      - master
```
kovayur, line 5: This line is not needed, right?

.github/workflows/lint.yml
kovayur, line 1: Nice. It would be even better to add it to our [pre-commit](https://github.com/rhysd/actionlint/blob/main/docs/usage.md#pre-commit) [configuration](https://github.com/stackrox/acs-fleet-manager/blob/0a01d2ff4baacd7fa34383a565f6fad677ebb5eb/.pre-commit-config.yaml). Then the extra GHA wouldn't be needed and we could validate the files locally.

kovayur (changes requested):
Nice refactoring 👍. It would be even nicer to see the github actions linter as a pre-commit hook, but not critical.

.github/workflows/lint.yml
porridge: Done.

porridge:
```
Executable `actionlint` not found
```
Ha, it's not as easy as I thought. Should I revert the latest commit, or do we want to fetch the tool in the precommit action? WDYT @kovayur ?

kovayur (approved):

PR #928, opened 2023-04-05
import rpm gpg key for CentOS

stehessel:
## Description

This is a workaround to the issue https://bugzilla.redhat.com/show_bug.cgi?id=2184640. This should fix the e2e tests in CI.

### DIFF
--- a/.openshift-ci/e2e-runtime/Dockerfile
+++ b/.openshift-ci/e2e-runtime/Dockerfile
@@ -1,5 +1,8 @@
 FROM quay.io/centos/centos:stream9
 
+# Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2184640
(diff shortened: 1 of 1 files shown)


.openshift-ci/e2e-runtime/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM quay.io/centos/centos:stream9
 
+# Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2184640
+rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
```
kovayur, line 4:
```suggestion
RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
```

kovayur (changes requested):

PR #951, opened 2023-04-13
ROX-15941: Run multiple operators from FSS

kurlov:
## Description
Make FSS able to run multiple ACS Operator versions.
Each operator version has its own deployment.
The metrics service and ServiceAccount can be shared between multiple versions.
For the sake of simplicity, there are two versions installed on startup before reconciler run.

The next steps are:
- fetch operators versions from FM
- installed desired versions on ticker base manner (similar to centrals reconcile)
- cleanup redundant operator versions

## Test manual

First of all, the OLM based ACS Operator should be deleted without removing an existing CRDs
1. List and delete OLM subscription for the ACS Operator

```
kubectl get subscription -n stackrox-operator
kubectl delete subscription <subscription> -n stackrox-operator
```

2. Delete the Operator’s ClusterServiceVersion

```
kubectl get clusterserviceversion -n stackrox-operator
kubectl delete clusterserviceversion rhacs-operator.<version> -n stackrox-operator
```

3. Check that there is no running ACS Operator

```
kubectl get pods -n stackrox-operator
NAME                                                              READY   STATUS      RESTARTS      AGE
```

Then it's time to run and verify
4. Build and deploy

```
make deploy/dev
```

5.  Check that a two ACS Operator versions are running

```
kubectl get pods -n stackrox-operator
NAME                                                              READY   STATUS      RESTARTS       AGE
c8dae07dab84cfc9348095952222435b18a827136292688d44a0b210f6nf8dl   0/1     Completed   0              118d
rhacs-operator-controller-manager-3.74.0-5fb9cbcdbb-56nqv         2/2     Running     0              13s
rhacs-operator-controller-manager-3.74.1-5765676ffc-l9bpp         2/2     Running     0              13s
...
```

5.  Check deployments

```
kubectl get deployments -n stackrox-operator
NAME                                       READY   UP-TO-DATE   AVAILABLE   AGE
rhacs-operator-controller-manager-3.74.0   1/1     1            1           27s
rhacs-operator-controller-manager-3.74.1   1/1     1            1           27s
```


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -777,27 +777,9 @@ func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remot
...
-			continue
-		}
-		if !apiErrors.IsNotFound(err) {
-			return fmt.Errorf("failed to retrieve object %s/%s of type %v: %w", key.Namespace, key.Name, obj.GroupVersionKind(), err)
-		}
-		err = r.client.Create(ctx, obj)
-		glog.V(10).Infof("Creating object %s/%s", obj.GetNamespace(), obj.GetName())
-		if err != nil && !apiErrors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create object %s/%s of type %v: %w", key.Namespace, key.Name, obj.GroupVersionKind(), err)
+		err := charts.InstallOrUpdateChart(ctx, obj, r.client)
```
SimonBaeumer, line 780: Can you extract this refactoring into a separate PR?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,59 @@ package operator
...
+// Image represents operator image
+type Image struct {
+	repository string
+	tag        string
+}
+
+// ImageFromString creates Image from image string
+func ImageFromString(image string) Image {
+	s := strings.Split(image, ":")
+	return Image{s[0], s[1]}
```
SimonBaeumer, line 30: The images are later received as user input. Can you add validation here?

SimonBaeumer (changes requested):
Nice, the implementation looks overall very good to me!
Some minor changes and the PR can be merged 🚢

fleetshard/pkg/central/charts/charts.go
```diff
@@ -65,3 +71,29 @@ func MustGetChart(name string) *chart.Chart {
 	}
 	return chrt
 }
+
+// InstallOrUpdateChart installs a new object from helm chart or update an existing object with the same Name, Namespace and Kind
+func InstallOrUpdateChart(ctx context.Context, obj *unstructured.Unstructured, client ctrlClient.Client) error {
```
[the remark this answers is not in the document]
kurlov: Yes, I would like to do it in the separate PR

fleetshard/pkg/runtime/runtime.go
```diff
@@ -209,12 +209,18 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
...
+	// TODO: Leave Operator installation before reconciler run until migration
+	operatorImages := [2]string{"quay.io/rhacs-eng/stackrox-operator:3.74.0", "quay.io/rhacs-eng/stackrox-operator:3.74.1"}
+	for _, image := range operatorImages {
+		glog.Infof("Install Operator: %s", image)
+		err := r.operatorManager.InstallOrUpgrade(ctx, operator.ImageFromString(image))
+		if err != nil {
+			return fmt.Errorf("operator upgrade: %w", err)
+		}
 	}
+	// TODO: delete redundant operators
```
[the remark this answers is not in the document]
kurlov: The versions which no longer needed. There is no ACS instance which needs this version so we should drop such operator

fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
```diff
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-operator:
-  # TODO: Split image into registry and image name + tag
-  image: quay.io/rhacs-eng/stackrox-operator:3.74.0
+operatorImage:
```
ludydoo, line 5: Would it make sense to keep the `operator` key, so that we could put all configuration for the operator deployment under this key?
For example
```
operator:
  tag: ...
  repository: ...
  resources: ...
  centralLabelSelector: ...
```

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,59 @@ package operator
...
 	objs, err := charts.RenderToObjects(releaseName, operatorNamespace, u.resourcesChart, chartVals)
 	if err != nil {
-		return fmt.Errorf("installing operator chart: %w", err)
+		return fmt.Errorf("failed rendering operator chart: %w", err)
 	}
 
-	// TODO(ROX-16338): handle namespace assigning with refactoring of chart deployment
 	for _, obj := range objs {
-		if obj.GetNamespace() == "" {
+		if obj.GetNamespace() == "" && obj.GetKind() != crdKind {
```
ludydoo, line 54: Is this to ensure that all objects have a namespace? If yes, there might be some object kinds that are not namespaced scoped, such as Namespace, ClusterRole, ClusterRoleBinding, PersistentVolume.

fleetshard/pkg/central/charts/charts.go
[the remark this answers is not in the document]
kurlov: probably worth to add in https://github.com/stackrox/acs-fleet-manager/pull/954

fleetshard/pkg/runtime/runtime.go
```diff
@@ -209,12 +209,18 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
...
 func (r *Runtime) upgradeOperator() error {
-	glog.Infof("Start Operator upgrade")
 	ctx := context.Background()
-	err := r.operatorManager.InstallOrUpgrade(ctx)
-	if err != nil {
-		return fmt.Errorf("operator upgrade: %w", err)
+	// TODO: gather desired operator versions from fleet-manager and update operators based on ticker
+	// TODO: Leave Operator installation before reconciler run until migration
+	operatorImages := [2]string{"quay.io/rhacs-eng/stackrox-operator:3.74.0", "quay.io/rhacs-eng/stackrox-operator:3.74.1"}
+	for _, image := range operatorImages {
```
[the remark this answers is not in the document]
kurlov: Sounds great!

fleetshard/pkg/central/reconciler/reconciler.go
kurlov: opened https://github.com/stackrox/acs-fleet-manager/pull/954

fleetshard/pkg/central/operator/upgrade.go
kurlov: I've change the implementation and added a check if there is no tags at all

fleetshard/pkg/central/operator/upgrade.go
kurlov: It's a naive approach. If object does not have namespace then it will be ignored and will take no affect

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,97 @@ package operator
...
+			s := strings.Split(img, ":")
+			val.repository = s[0]
+			val.tags = append(val.tags, s[1])
+		} else {
+			glog.Errorf("failed to parse image %s", img)
+		}
+	}
+	if len(val.tags) == 0 {
+		return nil, fmt.Errorf("zero tags parsed from images %s", strings.Join(images, ", "))
+	}
```
ludydoo, line 44: Should this be an error? e.g. for new clusters where there are no Centrals installed

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,97 @@ package operator
...
+		if obj.GetNamespace() == "" && obj.GetKind() != crdKind {
 			obj.SetNamespace(operatorNamespace)
 		}
 		key := ctrlClient.ObjectKey{Namespace: obj.GetNamespace(), Name: obj.GetName()}
 		var out unstructured.Unstructured
 		out.SetGroupVersionKind(obj.GroupVersionKind())
 		err := u.client.Get(ctx, key, &out)
 		if err == nil {
-			glog.V(10).Infof("Updating ACS Operator %s/%s", obj.GetNamespace(), obj.GetName())
+			glog.V(10).Infof("Updating object %s/%s", obj.GetNamespace(), obj.GetName())
```
ludydoo, line 83:
```suggestion
			glog.V(10).Infof("Updating %s %s/%s", obj.GetKind(), obj.GetNamespace(), obj.GetName())
```

fleetshard/pkg/runtime/runtime.go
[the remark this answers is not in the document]
SimonBaeumer:
```suggestion
	// TODO: delete unused operator versions
```

SimonBaeumer (approved):

johannes94 (changes requested):
LGTM, except the concern with the repository value.

porridge (changes requested):
I will look at the code in a moment, but I think this piece of description is not accurate:
> ~Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~

Please accurately document the rollout and rollback procedure from the release engineer's PoV.

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
```diff
@@ -34,8 +38,6 @@ spec:
               value: quay.io/rhacs-eng
             - name: ROX_OPERATOR_COLLECTOR_REGISTRY
               value: quay.io/rhacs-eng
-            - name: OPERATOR_CONDITION_NAME
-              value: rhacs-operator.v3.74.0
           image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0
```
porridge, line 41: Shouldn't this version be a parameter? Or do you assume it will be OK to upgrade this in all running operators when the newest one bumps the version?

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
porridge, line 1: Overall, what is the plan for keeping this file in sync with the upstream operator releases? Who, when and how copies the changes across the repos?

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
```diff
@@ -1,10 +1,14 @@
...
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
     app: rhacs-operator
     control-plane: controller-manager
-  name: rhacs-operator-controller-manager
+  ## Name field must contain up to 63 characters
+  ## https://www.rfc-editor.org/rfc/rfc1123
+  name: rhacs-operator-controller-manager-{{ $tag | trunc 29 }}
```
porridge, line 11: I think we should abort if `$tag` is longer. Silent truncation means the versions would clash if we cut off the trailing part (which is the part most likely to actually differ between versions).

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,97 @@ package operator
...
+
+func parseRepositoryAndTags(images []string) (*chartValue, error) {
+	val := chartValue{}
+	for _, img := range images {
+		if strings.Contains(img, ":") {
+			s := strings.Split(img, ":")
+			val.repository = s[0]
+			val.tags = append(val.tags, s[1])
+		} else {
+			glog.Errorf("failed to parse image %s", img)
```
porridge, line 39: Shouldn't this cause an error to be returned?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,97 @@ package operator
...
 // ACSOperatorManager keeps data necessary for managing ACS Operator
 type ACSOperatorManager struct {
 	client         ctrlClient.Client
 	resourcesChart *chart.Chart
 }
 
 // InstallOrUpgrade provisions or upgrades an existing ACS Operator from helm chart template
-func (u *ACSOperatorManager) InstallOrUpgrade(ctx context.Context) error {
+func (u *ACSOperatorManager) InstallOrUpgrade(ctx context.Context, images []string) error {
+	val, err := parseRepositoryAndTags(images)
```
porridge, line 57: `val` is not terribly informative... how about:
```suggestion
	operatorSettings, err := parseRepositoryAndTags(images)
```

fleetshard/pkg/central/operator/upgrade.go
porridge: Why single out the CRD kind, then?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,97 @@ package operator
...
 		out.SetGroupVersionKind(obj.GroupVersionKind())
 		err := u.client.Get(ctx, key, &out)
 		if err == nil {
-			glog.V(10).Infof("Updating ACS Operator %s/%s", obj.GetNamespace(), obj.GetName())
+			glog.V(10).Infof("Updating object %s/%s", obj.GetNamespace(), obj.GetName())
 			obj.SetResourceVersion(out.GetResourceVersion())
 			err := u.client.Update(ctx, obj)
 			if err != nil {
-				return fmt.Errorf("failed to update ACS Operator %s/%s of type %v %s", key.Namespace, key.Name, obj.GroupVersionKind(), err)
+				return fmt.Errorf("failed to update object %s/%s of type %s %w", key.Namespace, key.Name, obj.GetKind(), err)
```
porridge, line 87: Perhaps not in scope of this PR, but the kubectl convention is to separate the resource type and name with a slash (`pods/web-pod-13je7`), rather than namespace and name :thinking:

fleetshard/pkg/runtime/runtime.go
```diff
@@ -209,12 +210,17 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
 }
 
 func (r *Runtime) upgradeOperator() error {
-	glog.Infof("Start Operator upgrade")
 	ctx := context.Background()
-	err := r.operatorManager.InstallOrUpgrade(ctx)
+	// TODO: gather desired operator versions from fleet-manager and update operators based on ticker
+	// TODO: Leave Operator installation before reconciler run until migration
+	operatorImages := []string{"quay.io/rhacs-eng/stackrox-operator:3.74.0", "quay.io/rhacs-eng/stackrox-operator:3.74.1"}
+	glog.Infof("Install Operators: %s", strings.Join(operatorImages, ", "))
```
porridge, line 217:
```suggestion
	glog.Infof("Installing Operators: %s", strings.Join(operatorImages, ", "))
```

porridge (changes requested):

fleetshard/pkg/central/operator/upgrade.go
kurlov: Yes, it's about the case when we passed some tags and none of them were able to be parsed and applied

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
kurlov: Good point. I added a TODO for this

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
kurlov: CRDs will be downloaded directly from the main stackrox repository.
But I don't have a good solution for deployment. There will be an alert for failed operator start which could help to spot if deployment is incorrect for particular version

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
kurlov: Added random string to the end of deployment name

fleetshard/pkg/central/operator/upgrade.go
kurlov: I am not sure, I think it's better to keep parsing instead of canceling whole process

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,97 @@ package operator
...
+type chartValue struct {
+	repository string
+	tags       []string
+}
+
+func parseRepositoryAndTags(images []string) (*chartValue, error) {
+	val := chartValue{}
+	for _, img := range images {
+		if strings.Contains(img, ":") {
+			s := strings.Split(img, ":")
```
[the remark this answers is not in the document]
kurlov: yes, you are right. I thought that we don't need such feature and all images relieved from FM should have the same repository. But it's not complex to handle different repositories

fleetshard/pkg/central/operator/upgrade.go
kurlov: Only CRDs should not have namespace. Deployment, Service and ServiceAccount should be in the operator namespace

fleetshard/README.md
```diff
@@ -81,3 +81,92 @@ STATIC_TOKEN=<generated value | bitwarden value> \
 AUTH_TYPE=STATIC_TOKEN \
 ./dev/env/scripts/exec_fleetshard_sync.sh
 ```
+
+### Manage ACS Operator(s)
+
+Fleetshard-sync service is able to manager installation/update
```
ludydoo, line 87: typo
```suggestion
Fleetshard-sync service is able to manage installation/update
```

fleetshard/README.md
```diff
@@ -81,3 +81,92 @@ STATIC_TOKEN=<generated value | bitwarden value> \
...
+OLM reinstalls a new version of the operator even if the operator’s CSV was deleted earlier.
+In effect, you must tell OLM that you do not want new versions of the operator to be installed by deleting the ACS Operator subscription
+```
+kubectl get subscription -n <operator_namespace>
+kubectl delete subscription <subscription> -n <operator_namespace>
+```
+
+2. Delete the Operator’s ClusterServiceVersion.
+The ClusterServiceVersion contains all the information that OLM needs to manage an operator,
+and it effectively represents an operator that is installed on cluster
```
ludydoo, line 104: minor:
```suggestion
and it effectively represents an operator that is installed on the cluster
```

fleetshard/README.md
```diff
@@ -81,3 +81,92 @@ STATIC_TOKEN=<generated value | bitwarden value> \
...
+
+```
+kubectl get deployments -n <operator_namespace>
+NAME                                       READY   UP-TO-DATE   AVAILABLE   AGE
+rhacs-operator-controller-manager-3.74.1   1/1     1            1           27s
+```
+
+#### Rollback installation/update of ACS Operator:
+
+1. Redeploy Fleetshard-syn with disabled `FEATURE_FLAG_UPGRADE_OPERATOR_ENABLED` environment variable
```
ludydoo, line 141: minor:
```suggestion
1. Redeploy Fleetshard-sync with `FEATURE_FLAG_UPGRADE_OPERATOR_ENABLED=false` environment variable
```

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
```diff
@@ -1,10 +1,14 @@
...
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
     app: rhacs-operator
     control-plane: controller-manager
-  name: rhacs-operator-controller-manager
+  ## Name field must contain up to 63 characters
+  ## https://www.rfc-editor.org/rfc/rfc1123
+  name: rhacs-operator-controller-manager-{{ .tag | trunc 10 }}-{{ randAlphaNum 10 | lower }}
```
ludydoo, line 11: minor:
```suggestion
  name: rhacs-operator-controller-manager-{{ .tag | trunc 10 | lower }}-{{ randAlphaNum 10 | lower }}
```

porridge: Just a quick thought - doesn't using random strings cause all deployments to be replaced on every run? I don't think we want that. Why not just fail on longer tags? I don't think we'll ever need them :thinking:

ludydoo: @porridge I agree. Also since the "tag" is our "unique" key to install operator versions, we don't need a random suffix

fleetshard/pkg/central/operator/upgrade.go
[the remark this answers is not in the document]
johannes94: Ok, I think we'll need it to be able to switch between downstream/upstream images. Would you mind adding it in this PR or are you going to create a follow up PR. If so please add a comment so that we don't forgot about this.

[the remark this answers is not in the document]
kurlov: It's added, now we can use operators from different registries

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
kurlov: I've changed it in a way that long tags will be ignored. My only concern is that we probably should not return an error if there is one too long tag version but instead keep parsing the next image. I am not 100% sure about it though.

Let's assume that we have received two operator images for deployment. One is fine and another is too long to be the suffix for deployment. Should we cancel both deployment or only one with long version?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,98 @@ package operator
...
+	// maximum number of characters for operator deployment suffix
+	// See rhacs-operator-deployment.yaml for more details
+	maxOperatorDeploymentSuffixLength = 29
 )
 
+func parseOperatorImages(images []string) ([]chartutil.Values, error) {
+	var operatorImages []chartutil.Values
+	for _, img := range images {
+		if strings.Contains(img, ":") {
+			s := strings.Split(img, ":")
```
johannes94, line 32: I'd suggest to add a check that there actually are 2 elements in the slice after you run strings.Split() before accessing the array (otherwise we might get a panic if the string is malformed). This usually won't happen but better safe than sorry. I'll approve and leave it to you if you want to add this.

johannes94 (approved):

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
porridge: I don't think we should be silently ignoring requests to install a problematic operator. These cases always make me think of a compiler which just ignores the statements that it didn't understand, because it wants to be robust and produce a program that works "mostly" :-)

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,98 @@ package operator
...
+		if obj.GetNamespace() == "" && obj.GetKind() != crdKind {
 			obj.SetNamespace(operatorNamespace)
 		}
 		key := ctrlClient.ObjectKey{Namespace: obj.GetNamespace(), Name: obj.GetName()}
 		var out unstructured.Unstructured
 		out.SetGroupVersionKind(obj.GroupVersionKind())
 		err := u.client.Get(ctx, key, &out)
 		if err == nil {
-			glog.V(10).Infof("Updating ACS Operator %s/%s", obj.GetNamespace(), obj.GetName())
+			glog.V(10).Infof("Updating %s/%s", obj.GetKind(), obj.GetName())
```
porridge, line 84: Let's keep the namespace in the message, it might be extremely useful for debugging. Similar below.

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,98 @@ package operator
...
 )
 
+func parseOperatorImages(images []string) ([]chartutil.Values, error) {
+	var operatorImages []chartutil.Values
+	for _, img := range images {
+		if strings.Contains(img, ":") {
+			s := strings.Split(img, ":")
+			tag := s[1]
+			if len(tag) > maxOperatorDeploymentSuffixLength {
+				glog.Errorf("tag version %s contains more than %d characters and cannot be used as a deployment suffix", tag, maxOperatorDeploymentSuffixLength)
```
porridge, line 35:
```suggestion
				glog.Errorf("Tag version %s contains more than %d characters and cannot be used as a deployment suffix.", tag, maxOperatorDeploymentSuffixLength)
```

fleetshard/pkg/central/operator/upgrade.go
porridge: I don't think we should be silently discarding malformed image specs. Better to fail early and loud (but perhaps collect all errors from the parsing and return them together).

fleetshard/pkg/runtime/runtime.go
```diff
@@ -210,12 +211,17 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
...
-	glog.Infof("Start Operator upgrade")
 	ctx := context.Background()
-	err := r.operatorManager.InstallOrUpgrade(ctx)
+	// TODO: gather desired operator versions from fleet-manager and update operators based on ticker
+	// TODO: Leave Operator installation before reconciler run until migration
+	operatorImages := []string{"quay.io/rhacs-eng/stackrox-operator:3.74.0", "quay.io/rhacs-eng/stackrox-operator:3.74.1"}
+	glog.Infof("Installing Operators: %s", strings.Join(operatorImages, ", "))
+	err := r.operatorManager.InstallOrUpgrade(ctx, operatorImages)
 	if err != nil {
 		return fmt.Errorf("operator upgrade: %w", err)
```
porridge, line 221:
```suggestion
		return fmt.Errorf("ensuring initial operator installation failed: %w", err)
```

fleetshard/README.md
```diff
@@ -81,3 +81,92 @@ STATIC_TOKEN=<generated value | bitwarden value> \
...
+3. Check that there is no running ACS Operator pod(s)
+
+```
+kubectl get pods -n <operator_namespace>
+NAME
+...
+```
+
+4. install ACS Operator
+In Openshift console go to `Operators -> OperatorHub -> Advanced Cluster Security for Kubernetes`
```
porridge, line 158: Isn't this something that the terraforming helm chart should be doing?
Speaking of this, shouldn't this PR also remove the `acs-operator.yaml` from that chart? Otherwise there will be a conflict on rollout, no?

porridge (changes requested):

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
kurlov: I added fail fast on parsing images

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,67 +4,98 @@ package operator
...
 	operatorNamespace = "stackrox-operator"
 	releaseName       = "rhacs-operator"
-	operatorImage     = "quay.io/rhacs-eng/stackrox-operator:3.74.0"
+	crdKind           = "CustomResourceDefinition"
+	// maximum number of characters for operator deployment suffix
+	// See rhacs-operator-deployment.yaml for more details
+	maxOperatorDeploymentSuffixLength = 29
 )
 
+func parseOperatorImages(images []string) ([]chartutil.Values, error) {
```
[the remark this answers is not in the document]
kurlov: Sounds nice! Implemented parsing in a way that it ignores duplicates

fleetshard/README.md
```diff
@@ -81,3 +81,92 @@ STATIC_TOKEN=<generated value | bitwarden value> \
...
+```
+kubectl get deployments -n <operator_namespace>
+NAME                                       READY   UP-TO-DATE   AVAILABLE   AGE
+rhacs-operator-controller-manager-3.74.1   1/1     1            1           27s
+```
+
+#### Rollback installation/update of ACS Operator:
+
+1. Redeploy Fleetshard-sync with disabled `FEATURE_FLAG_UPGRADE_OPERATOR_ENABLED=false` environment variable
+2. Delete existing ACS Operator deployment(s)
```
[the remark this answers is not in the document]
kurlov: Good point

fleetshard/README.md
kurlov: I aded an alternative approach with helm chart.
I also added a TODO to disable operator subscription installation. There is a feature flag so we should remove `acs-operator.yaml` only when `FEATURE_FLAG_UPGRADE_OPERATOR_ENABLED` is turned on. That's clumsy 😞

dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
```diff
@@ -1,3 +1,4 @@
+## TODO disable operator installation when FSS runs with FEATURE_FLAG_UPGRADE_OPERATOR_ENABLED=true
```
porridge, line 1: Can you file a task for the flag flip and change the format to `TODO(ROX-NNNN): `? Also in other such places.

fleetshard/README.md
```diff
@@ -81,3 +81,98 @@ STATIC_TOKEN=<generated value | bitwarden value> \
...
+1. Make sure that OLM ACS Operator subscription is deleted.
+OLM uses the subscription resource to subscribe to the latest version of an operator.
+OLM reinstalls a new version of the operator even if the operator’s CSV was deleted earlier.
+In effect, you must tell OLM that you do not want new versions of the operator to be installed by deleting the ACS Operator subscription
+```
+kubectl get subscription -n <operator_namespace>
+kubectl delete subscription <subscription> -n <operator_namespace>
+```
+
+2. Delete the Operator’s ClusterServiceVersion.
```
porridge, line 102: There is also the `OperatorGroup`. Perhaps just advise setting `acsOperator.enabled` to `false`? But now that you pointed out this functionality is behind a feature flag, it's OK to delay the rollout/rollback considerations to the PR that flips the flag.

fleetshard/README.md
porridge: Um, I don't think a simple `helm template -s templates/acs-operator.yaml --values values.yaml` will work. We depend on the values `--set` by the terraforming script, so I would just advise to re-terraform with `acsOperator.enabled`.

fleetshard/pkg/central/operator/upgrade.go
porridge: > Similar below.

The cases below were not addressed AFAICT?

fleetshard/pkg/central/operator/upgrade.go
porridge: So what's up with this one?

porridge (changes requested):

fleetshard/README.md
kurlov: Added re-terraform step. `acsOperator.enabled` is enabled by default

dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
kurlov: I think we should keep `acs-operator.yaml` for some time after enabling FSS operator installation. So I added two TODOs. One is for disabling operator installation in terraform script. Another is to drop `acs-operator.yaml` completely

fleetshard/README.md
kurlov: `OperatorGroup` looks generic. I am not sure we should delete it. Also, OLM documentation for removing operator does not mention `OperatorGroup` neither

dp-terraform/helm/rhacs-terraform/templates/acs-operator.yaml
```diff
@@ -1,3 +1,4 @@
+## TODO ROX-16646: drop acs-operator.yaml template
```
porridge, line 1:
```suggestion
## TODO(ROX-16646): drop acs-operator.yaml template
```

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -100,6 +100,7 @@ if [[ "${OPERATOR_USE_UPSTREAM}" == "true" ]]; then
     OPERATOR_SOURCE="rhacs-operators"
 fi
 
+# TODO ROX-16645: set acsOperator.enabled to false
```
porridge, line 103:
```suggestion
# TODO(ROX-16645): set acsOperator.enabled to false
```

dp-terraform/helm/rhacs-terraform/test.sh
```diff
@@ -5,6 +5,7 @@ CLUSTER_ID="test-clusterId"
 FM_ENDPOINT="127.0.0.1:443"
 OCM_TOKEN="example-token"
 
+# TODO ROX-16645: set acsOperator.enabled to false
```
porridge, line 8:
```suggestion
# TODO(ROX-16645): set acsOperator.enabled to false
```

fleetshard/README.md
porridge: Not sure what you mean by "generic"... This is an `OperatorGroup` specific to our installation of the ACS operator via OLM. If we are getting rid of the `Subscription` then we absolutely should also clean up the `OperatorGroup`.

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
```diff
@@ -78,13 +82,11 @@ spec:
                   containerName: manager
                   resource: limits.memory
                   divisor: '0'
-            - name: OPERATOR_CONDITION_NAME
-              value: rhacs-operator.v3.74.0
-            {{ if .Values.operator.centralLabelSelector -}}
+            {{- if $.Values.operator.centralLabelSelector -}}
             - name: CENTRAL_LABEL_SELECTOR
-              value: {{ .Values.operator.centralLabelSelector | quote }}
+              value: {{ $.Values.operator.centralLabelSelector }}
```
porridge, line 87: I must be missing something... we're deploying multiple controllers but they all share the same central label selector? :thinking:
Also, I don't see it being set anywhere. Is this dead code?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
-	apiErrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
 	operatorNamespace = "stackrox-operator"
 	releaseName       = "rhacs-operator"
-	operatorImage     = "quay.io/rhacs-eng/stackrox-operator:3.74.0"
+	crdKind           = "CustomResourceDefinition"
```
porridge, line 22: Delete unused const?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
+	// maximum number of characters for operator deployment suffix
+	// See rhacs-operator-deployment.yaml for more details
+	maxOperatorDeploymentSuffixLength = 29
```
porridge, line 25: I'm a little concerned that if someone changes the prefix length in the helm chart, they will not know to update this magic number here. How about passing the whole name from go code into the chart instead?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
+func parseOperatorImages(images []string) ([]chartutil.Values, error) {
+	var operatorImages []chartutil.Values
+	uniqueTags := make(map[string]bool)
+	for _, img := range images {
+		if strings.Contains(img, ":") {
+			strs := strings.Split(img, ":")
+			if len(strs) != 2 {
+				return nil, fmt.Errorf("failed to split image and tag from %s", img)
+			}
+			tag := strs[1]
```
porridge, line 37:
```suggestion
			repo, tag := strs[0], strs[1]
```
Then using `repo` below instead of `strs[0]` is a bit neater.

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
+	// maximum number of characters for operator deployment suffix
+	// See rhacs-operator-deployment.yaml for more details
+	maxOperatorDeploymentSuffixLength = 29
 )
 
+func parseOperatorImages(images []string) ([]chartutil.Values, error) {
+	var operatorImages []chartutil.Values
+	uniqueTags := make(map[string]bool)
+	for _, img := range images {
+		if strings.Contains(img, ":") {
```
porridge, line 32: It would be a bit more idiomatic to:
```suggestion
		if !strings.Contains(img, ":") {
					return nil, fmt.Errorf("failed to parse image %s", img)
		}
```
and de-dent the other branch. You might even be able to drop this in favor of just the `Split` call which should DTRT in case of missing ":"?

porridge (changes requested):
I think we're getting there!

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
ludydoo: This probably needs to be something like
```suggestion
              value: "rhacs.redhat.com/version={{ .tag | lower }}"
```

ludydoo (commented):
Just as a general question:

I see that you modified the helm chart so that we install multiple operators from the same helm release. Did you consider that each operator could also be its own helm release?

I'm wondering what would be the pros and cons of both approaches. 

Approach 1 (this PR)
- Fleetshard sync aggregates all desired installed operator versions
- Passes this list to the helm values in the `.operator.images`
- Helms installs multiple operator deployments from the same helm release through a loop

Approach 2
- Fleetshard sync aggregates all desired installed operator versions
- Would create 1 helm release for each operator version
- Helm installs 1 operator deployment for each helm release


Approach 1 Pros: 
- Simpler logic
- Probably that helm would remove superfluous versions itself
- Can use shared resources (e.g. metrics service?)

Approach 1 Cons:
- Not as configurable as 1 operator per release
- We are more coupled between operator charts of different versions. 

Approach 2 Pros:
- Each operator can be configured, e.g. resources more easily, RBAC, etc. 
- Each operator installation is "independent" (its own helm release). Less coupling

Approach 2 Cons:
- More complex logic to sync/reconcile the operators to update, add, delete

thoughs @stackrox/draco  ?

fleetshard/pkg/central/operator/upgrade.go
SimonBaeumer: This is a hard kubernetes label. Ideally the Fleet-Manager API forbids image names >29characters. If this limit is not enforced on fleetshard, fleetshard cannot apply the Helm chart.

SimonBaeumer:
@ludydoo The first implementation in @kurlov PR was installing it with custom Helm installations for each version. After a discussion it was changed to the current approach for simplicity for the cons you've listed in "Approach 2 Cons":

> More complex logic to sync/reconcile the operators to update, add, delete

I prefer Approach 1 more because only one Helm chart release is applied which feels more natural and causes less side-effects, i.e. accidental overwrites of resources because they are applied multiple times.
Changing the operator installation is relatively easy later in production because it does not affect the SLA.
I am suggesting moving on with the one Helm chart implementation.

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
ludydoo: or 
```suggestion
              value: {{ substr 0 64 (lower (printf "rhacs.redhat.com/version=" $.Values.operator.centralLabelSelector)) }}
```

ludydoo: @porridge I added this bit (CENTRAL_LABEL_SELECTOR) a few weeks ago, as I assumed that there would be 1 helm release per operator. But as @SimonBaeumer explained, it was actually decided to use 1 helm release for all operators.

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
+			strs := strings.Split(img, ":")
+			if len(strs) != 2 {
+				return nil, fmt.Errorf("failed to split image and tag from %s", img)
+			}
+			tag := strs[1]
+			if len(tag) > maxOperatorDeploymentSuffixLength {
+				return nil, fmt.Errorf("tag version %s contains more than %d characters and cannot be used as a deployment suffix", tag, maxOperatorDeploymentSuffixLength)
+			}
+			if _, used := uniqueTags[tag]; !used {
```
ludydoo, line 41: would that mean that the operator cannot install two images from different repositories with the same tag ? Why not just check the uniqueness based on the repository + tag instead ?

fleetshard/pkg/central/operator/upgrade.go
ludydoo: Perhaps we could consider
- having fleetshard sync compute the suffix, and pass it as an additional parameter in the `images`. eg. `images: [{repository: foo, tag: bar, suffix: foobar}]`. 
- using a hash of the repository + tag in the deployment suffix, so that we can install operators from different repositories that have the same tag. 
- creating a distinct namespace per operator `operator-<repository>-<version>`

fleetshard/pkg/central/operator/upgrade_test.go
```diff
@@ -88,18 +109,108 @@ func TestOperatorUpgradeFreshInstall(t *testing.T) {
...
+				{"repository": operatorRepository, "tag": "3.74.2"},
+			},
+		},
+		"should ignore duplicate operator images": {
+			images: []string{operatorImage1, operatorImage1},
+			expected: []map[string]string{
+				{"repository": operatorRepository, "tag": "3.74.1"},
+			},
+		},
+		"fail if image does contain colon": {
```
ludydoo, line 186: Should this be supported ? 

```suggestion
		"should accept images from multiple repositories with the same tag": {
			images: []string{"repo1:tag","repo1:tag"},
			expected: []map[string]string{
				{"repository":"repo1", "tag": "tag"},
				{"repository":"repo2", "tag": "tag"},
			},
		},
		"fail if image does contain colon": {
```

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
```diff
@@ -78,13 +82,11 @@ spec:
                   containerName: manager
                   resource: limits.memory
                   divisor: '0'
-            - name: OPERATOR_CONDITION_NAME
-              value: rhacs-operator.v3.74.0
-            {{ if .Values.operator.centralLabelSelector -}}
+            {{- if $.Values.operator.centralLabelSelector -}}
```
ludydoo, line 85:
```suggestion
            {{- if $.Values.operator.centralLabelSelector -}}
```

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
```diff
@@ -78,13 +82,11 @@ spec:
...
             - name: CENTRAL_LABEL_SELECTOR
-              value: {{ .Values.operator.centralLabelSelector | quote }}
+              value: {{ $.Values.operator.centralLabelSelector }}
             {{- end }}
```
ludydoo, line 88:
```suggestion
            {{- end }}
```

fleetshard/pkg/central/operator/upgrade.go
kurlov: I like the idea to pass full name from FSS.

fleetshard/pkg/central/operator/upgrade.go
kurlov: I like it

fleetshard/pkg/central/operator/upgrade.go
kurlov: You are right, I've changed the implementation

fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
```diff
@@ -3,6 +3,15 @@
 # Declare variables to be passed into your templates.
 
 operator:
```
ludydoo, line 5: Not sure about this, but consider renaming `images` to `operators`, or `deployments` such as 
```
deployments:
- operator:
    image:
      repository: foo
      tag: bar
    resources: {}
  kubeProxy:
    resources: {}
 ...
```

fleetshard/README.md
kurlov: I was wrong. I thought that OperatorGroup is shared between different operators, but it's not and it's safe to delete it. I've added instructions about it

fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
kurlov: I agree, but I would prefer to do it in the separate PR where more fields will be added

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
+		strs := strings.Split(img, ":")
+		if len(strs) != 2 {
+			return nil, fmt.Errorf("failed to split image and tag from %s", img)
+		}
+		repo, tag := strs[0], strs[1]
+		if len(operatorDeploymentPrefix+"-"+tag) > maxOperatorDeploymentNameLength {
+			return nil, fmt.Errorf("%s-%s contains more than %d characters and cannot be used as a deployment name", operatorDeploymentPrefix, tag, maxOperatorDeploymentNameLength)
+		}
+		if _, used := uniqueTags[repo+tag]; !used {
+			uniqueTags[repo+tag] = true
```
ludydoo, line 48: This is minor, but consider
- Renaming `uniqueTags` to `uniqueImages` or something similar, as this suggests that it only holds `tags`.
- very minor: `uniqueTags := map[string]struct{}` rather than `uniqueTags := map[string]bool`

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
ludydoo: I think this still needs to be removed

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
 
+func parseOperatorImages(images []string) ([]chartutil.Values, error) {
+	if len(images) == 0 {
+		return nil, fmt.Errorf("the list of images is empty")
+	}
+	var operatorImages []chartutil.Values
+	uniqueTags := make(map[string]bool)
+	for _, img := range images {
+		if !strings.Contains(img, ":") {
+			return nil, fmt.Errorf("failed to parse image %s", img)
```
porridge, line 37:
```suggestion
			return nil, fmt.Errorf("failed to parse image %q", img)
```
similarly below

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,42 +4,80 @@ package operator
...
+		}
+		strs := strings.Split(img, ":")
+		if len(strs) != 2 {
+			return nil, fmt.Errorf("failed to split image and tag from %s", img)
+		}
+		repo, tag := strs[0], strs[1]
+		if len(operatorDeploymentPrefix+"-"+tag) > maxOperatorDeploymentNameLength {
```
porridge, line 44: Nit: you're replicating the concatenation that also happens in the chart. It would be more robust to just pass the concatenated name to the chart, as it makes it impossible for these two operations ever get out of sync and cause trouble.

porridge (approved):
LGTM overall, some nitpicks inline and +1 to those from @ludydoo

fleetshard/pkg/central/operator/upgrade.go
kurlov: Renamed, but `bool` is used because `true` is  inserted for every unique image

fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
```diff
@@ -3,6 +3,15 @@
...
-  # TODO: Split image into registry and image name + tag
-  image: quay.io/rhacs-eng/stackrox-operator:3.74.0
-  # TODO: and values for resource limits and requests for both proxy and manager container
+  deploymentPrefix: rhacs-operator-controller-manager
```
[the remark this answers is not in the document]
kurlov: I dropped `-controller` part

PR #965, opened 2023-04-18
improve probe logging

stehessel:
## Description

Small improvement of the probe logging:
* Remove warnings emitted frequently while the probe is waiting for the instance to be ready/deleted. This warning creates a lot of spam in the logs.
* Log target configuration when a probe run starts.


dhaus67 (approved):
While we're at it, in the past when dealing with failing probe runs, the error messages for e.g. creation request failing where something along the lines of:
```
central creation failed: 403 forbidden
```
It would make initial triaging a lot faster when we could also print the error message within the logs contained within the `http.Response` we receive. OK for me to do this in a follow-up PR (I'm also open to change this if you don't feel like it right now since we are all relying on the probe service one way or another :) ).

PR #972, opened 2023-04-20
ROX-15981: force reconcile flag

johannes94:
## Description
There are situation where we want to reconcile a central request in fleetshard-sync even if it is in Ready state and has not changed. For example to restore set a new DB password for the tenant RDS DB after we restored a delete central tenant. (Distaster Recovery for accidental deleted tenant). Those cases are usually special action performed by system administrators.

To be able to force reconciliation in such cases this PR adds a ForceReconcile Flag to the private and admin API object and to the DB object aswell. The ForceReconcile Flag can be set through the admin PATCH API in order to force a reconciliation for a central object that would otherwise be skipped.

## Test manual

```
# To run tests locally run:
```


internal/dinosaur/pkg/api/private/model_managed_central.go
```diff
@@ -12,9 +12,10 @@ package private
...
-	Kind          string                      `json:"kind,omitempty"`
-	Metadata      ManagedCentralAllOfMetadata `json:"metadata,omitempty"`
-	Spec          ManagedCentralAllOfSpec     `json:"spec,omitempty"`
-	RequestStatus string                      `json:"requestStatus,omitempty"`
+	Id             string                      `json:"id,omitempty"`
+	Kind           string                      `json:"kind,omitempty"`
+	Metadata       ManagedCentralAllOfMetadata `json:"metadata,omitempty"`
+	Spec           ManagedCentralAllOfSpec     `json:"spec,omitempty"`
+	RequestStatus  string                      `json:"requestStatus,omitempty"`
+	ForceReconcile bool                        `json:"forceReconcile,omitempty"`
```
vladbologa, line 20: Setting this will field will force reconcile on this Central forever, until another request is made to set ForceReconcile to false?

My concern is that we might forget instances with this flag on. So I'd prefer a one-off "forceReconcile", but maybe that's not trivial to implement?

vladbologa (changes requested):
I like the idea, left a few comments.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -110,10 +110,10 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 	if err != nil {
 		return nil, errors.Wrapf(err, "checking if central changed")
 	}
-
+	useCache := !changed && !remoteCentral.ForceReconcile
```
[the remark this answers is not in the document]
johannes94: Done.

internal/dinosaur/pkg/api/private/model_managed_central.go
johannes94: Actually I had also this idea in mind. I could use a new int field e.g `ReconcileTrigger` and set it to a random number, each time we want to reconcile. But I went with this approach because it seems easier to understand and might be tedious in runbooks if multiple reconcile steps are needed to say "Call this API, wait for reconcile, see if it did everything you need, if not trigger another reconcile until it does".

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -110,8 +110,9 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 	if err != nil {
 		return nil, errors.Wrapf(err, "checking if central changed")
 	}
+	needsReconcile := changed || remoteCentral.ForceReconcile
```
SimonBaeumer, line 113: Like it!
Perfect if extracted in a private func `needsReconcile()`

vladbologa (approved):

internal/dinosaur/pkg/migrations/202304200000_add_force_reconcile_to_central_request.go
```diff
@@ -0,0 +1,41 @@
...
+		Migrate: func(tx *gorm.DB) error {
+			if !tx.Migrator().HasColumn(&CentralRequest{}, colName) {
+				if err := tx.Migrator().AddColumn(&CentralRequest{}, colName); err != nil {
+					return errors.Wrapf(err, "adding column %s in migration %s", colName, id)
+				}
+				glog.Infof("Successfully added the %s column", colName)
+			}
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
```
SimonBaeumer, line 31: Can you remove the rollback of the migration?
It creates headaches if a rollback is executed and might result in a data-loss, when in reality we fixed a nil-pointer exception.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -321,6 +321,7 @@ func updateCentralRequest(request *dbapi.CentralRequest, updateRequest *private.
 	}
 
 	new.DesiredCentralVersion = updateRequest.CentralVersion
+	new.ForceReconcile = updateRequest.ForceReconcile
```
SimonBaeumer, line 324: How is it guaranteed that the `forceReconcile` flag is temporary and set to `false` after the reconcile was executed?

SimonBaeumer (changes requested):
.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
johannes94: It's not see https://github.com/stackrox/acs-fleet-manager/pull/972#discussion_r1173557094 . What do you think, should we go with forcing single reconciles instead?

fleetshard/pkg/central/reconciler/reconciler.go
johannes94: Done.

internal/dinosaur/pkg/migrations/202304200000_add_force_reconcile_to_central_request.go
johannes94: Discussed on slack, that we're going to stay with the current approach for Rollbacks at least until we have zero downtime deployments. The main reason for this was that this Rollback is only executed on explicit command of engineers and not automatically. This makes unexpected data loss very unlikely.

SimonBaeumer (approved):

internal/dinosaur/pkg/api/private/model_managed_central.go
johannes94: @SimonBaeumer @vladbologa please re-review, I've changed ForceReconcile to be a string, mainly because of the problem with gorm and updating default boolean values. But a nice side effect is that this allows us to always force reconcilation by setting it to `always` and trigger one time reconcilation by setting it to any string we like, so that the `changed` bool will be true

internal/dinosaur/pkg/api/dbapi/central_request_types.go
```diff
@@ -91,6 +91,12 @@ type CentralRequest struct {
 
 	// All we need to integrate Central with an IdP.
 	AuthConfig
+
+	// ForceReconcile will be set by the admin API to indicate to fleetshard-sync that this instance needs
+	// to be reconciled even if it has not changed and is in a state were reconciliation should be skipped.
+	// Set this to "always" to force reconcilation until unset or to any other string to force a
```
vladbologa, line 97: Does unsetting work, or it actually causes the same problem with gorm as setting a bool to false?

vladbologa (approved):

internal/dinosaur/pkg/handlers/admin_dinosaur.go
SimonBaeumer: > should we go with forcing single reconciles instead?

I've expected to trigger one reconcile in fleetshard-sync. I think the use-cases clash here - restoring DB credentials affects triggering a reconcile.
Why does the reconcile not be triggered when the Central is restored automatically? Some field must change to restore a Central in an accidental delete.

johannes94: Actually no field changes in this case, with the current implementation we can't really restore a "central_request". You can see the draft for the restore process here: https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/-/merge_requests/1.

It is basically creating a new central_request and after that swap the DB data underneath it with the data of the delete tenant. That's why no field for the central_request ever changes during this process.

internal/dinosaur/pkg/api/dbapi/central_request_types.go
johannes94: Unsetting to "" doesn't work but setting it to anything other than that works. This is actually something that is a flaw in the whole REST API of fleet-manager if any string in an update operation is empty string it will not be updated by the default gorm update operations.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
SimonBaeumer: That makes sense :approve:

SimonBaeumer (approved):

internal/dinosaur/pkg/api/dbapi/central_request_types.go
vladbologa: Then just a nit: I would remove the "until unset" part of the comment, because it doesn't actually work.

johannes94: Done.

PR #973, opened 2023-04-20
Add response logging to probe service

dhaus67:
## Description

Add logging of the HTTP response body in case of errors to the probe service.

Previously, if an error occurred, we only saw the HTTP status code within the log message:
```log
5536 probe.go:195] central instance id-42 not reachable: : 500
```

While this may hint to a problem, we can also add the HTTP response body and the error code + reason to the log statement, thus making triaging issues a bit easier.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/probe/pkg/probe/probe.go
+++ b/probe/pkg/probe/probe.go
@@ -11,8 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/stackrox/rox/pkg/utils"
-
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/constants"
@@ -22,6 +20,7 @@ import (
 	"github.com/stackrox/acs-fleet-manager/probe/config"
 	"github.com/stackrox/acs-fleet-manager/probe/pkg/metrics"
 	"github.com/stackrox/rox/pkg/httputil"
+	"github.com/stackrox/rox/pkg/utils"
 )
 
 // Probe is a wrapper interface for the core probe logic.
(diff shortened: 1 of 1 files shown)


dhaus67:
* main
  * **PR #973** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/973" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/973?utm_source=stack-comment).

probe/pkg/probe/probe.go
```diff
@@ -5,11 +5,14 @@ import (
...
 	"crypto/rand"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
 	"time"
 
+	"github.com/stackrox/rox/pkg/utils"
+
```
stehessel, line 15: Is the newline intentional? Below there is already a `stackrox/rox` module.

probe/pkg/probe/probe.go
```diff
@@ -288,3 +294,16 @@ func retryUntilSucceededWithResponse(ctx context.Context, fn func(context.Contex
 		}
 	}
 }
+
+func extractCentralError(resp *http.Response) string {
+	var centralError public.Error
+	if resp == nil || resp.Body == nil {
+		return ""
+	}
+	defer utils.IgnoreError(resp.Body.Close)
```
stehessel, line 303: Why did you move the `Close` to this function? I would expect the close to be within the same function that declared the response. Otherwise it feels like a side effect of the extract function, even though in this case it doesn't really matter.

probe/pkg/probe/probe.go
dhaus67: Good old golangci-lint

probe/pkg/probe/probe.go
dhaus67: Initially did it since the response body wasn't really used by any other function, but yeah let's add it to the same function where it's declared.

probe/pkg/probe/probe.go
stehessel: What I mean is, if you remove the newline it will be sorted below with the other modules.

dhaus67: I thought I fixed it in the last commit, my bad.

stehessel (approved):

PR #984, opened 2023-04-25
ROX-16161: Change the AWS authentication method in Fleetshard Sync

kovayur:
## Description
This PR changes the authentication method in AWS for Fleetshard Sync from RHSSO to Service Account tokens issued by the OAuth Server built-in to the Data Plane cluster. This will allow to grant more granular permission to apps, be independent from RHSSO and use STS for another apps, where we don't want to use IAM User access keys (like cloudwatch-exporter).

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/docs/development/setup-aws-idp-on-osd.md
+++ b/docs/development/setup-aws-idp-on-osd.md
@@ -1,0 +1,70 @@
+# Authenticate ACSCS services on AWS using OSD cluster Identity Provider
+
+## Overview
+The goal is to authenticate ACSCS applications (pods) to AWS. The recommended approach is to use STS so that each application assumes an IAM role with policies describing what that application is allowed to do.
+Every Openshift cluster (OCP) has a built-in OAuth server that is used to authenticate service accounts within the cluster. OSD is no exception. This document describes how to add an Identity Provider to the AWS IAM configuration to use the OAuth server to authenticate ACSCS pods to AWS.
+
+## Implementation
+In scope are the environments that run on OSD clusters, namely _stage_ and _prod_. Engineers may want to set up their own OSD cluster to make experiments. The setup procedure is described  [below](#manual-setup-dev).
+The main idea is the following: The OAuth server has been added to the AWS IAM configuration as an Identity Provider. Each cluster has its own server, so each must be declared in the AWS IAM configuration. AWS has a limit of 100 identity providers per account [1].
+The complication is that server is not reachable from AWS STS, therefore, it couldn't load the public keys to verify a JWT token. The recommended approach is to expose OAuth server's public keys in a **public** S3 bucket [2].
+
+S3 bucket name has the following pattern: `<osdClusterName>-<randomAlphanumeric>-oidc`
+- `osdClusterName` is to identify the cluster associated with this bucket
+- `randomAlphanumeric` - 32-character random string to ensure uniqueness of the S3 bucket
+- `oidc` suffix is chosen to indicate that bucket corresponds to an OAuth server. This suffix also corresponds to the pattern used in the `ccoctl` tool [3].
+
+## Manual setup (dev)
+> ❗️This instruction is intended only for OSD clusters. For local clusters, it is recommended to use the OCM token for authentication.
+
+1. Prepare the environment
+    ```shell
+    ENVIRONMENT=dev
+    AWS_REGION=us-east-1
+    GITROOT=$(git rev-parse --show-toplevel)
+    CLUSTER_ID="<ID of the cluster>"
+    CLUSTER_NAME="<name of the cluster>"
+    IDP_NAME="${CLUSTER_NAME}-${CLUSTER_ID}"
+    OUTPUT_DIR="${IDP_NAME}"
+    cd $GITROOT/tmp
+    mkdir $OUTPUT_DIR
+    ```
+1. Login into the osd cluster
+    ```shell
+    ocm cluster login $CLUSTER_ID
+    # follow instructions to login
+    ```
+1. Retrieve a public key from the cluster
+    ```shell
+    oc get configmap --namespace openshift-kube-apiserver bound-sa-token-signing-certs --output json | jq --raw-output '.data["service-account-001.pub"]' > "${OUTPUT_DIR}/serviceaccount-signer.public"
+    ```
+1. Authenticate in AWS
+    ```saml
+    make -C $GITROOT $GITROOT/bin/tools_venv
+    source $GITROOT/bin/tools_venv/bin/activate
+    aws-saml.py
+    ```
+1. Create an Identity Provider
+    ```shell
+    ccoctl aws create-identity-provider --output-dir $OUTPUT_DIR --name $IDP_NAME --region $AWS_REGION
+    ```
+1. Change the cluster authentication
+    ```shell
+    oc apply -f ./dev-dp-idp/manifests/cluster-authentication-02-config.yaml
+    ```
+1. Make sure that the cluster authentication has an appropriate url pointing to the S3 bucket
+    ```shell
+    oc get authentication -o yaml
+    ```
+
+## Tear down
+```shell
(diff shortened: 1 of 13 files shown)


docs/development/setup-aws-idp-on-osd.md
```diff
@@ -0,0 +1,70 @@
+# Authenticate ACSCS services on AWS using OSD cluster Identity Provider
+
+## Overview
+The goal is to authenticate ACSCS applications (pods) to AWS. The recommended approach is to use STS so that each application assumes an IAM role with policies describing what that application is allowed to do.
+Every Openshift cluster (OCP) has a built-in OAuth server that is used to authenticate service accounts within the cluster. OSD is no exception. This document describes how to add an Identity Provider to the AWS IAM configuration to use the OAuth server to authenticate ACSCS pods to AWS.
+
+## Implementation
+In scope are the environments that run on OSD clusters, namely _stage_ and _prod_. Engineers may want to set up their own OSD cluster to make experiments. The setup procedure is described  [below](#manual-setup-dev).
```
dhaus67, line 8: nit:
```suggestion
In scope are the environments that run on OSD clusters, namely _stage_ and _prod_. Engineers may want to set up their own OSD cluster to make experiments. The setup procedure is described [below](#manual-setup-dev).
```

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
```diff
@@ -382,8 +381,8 @@ func newRdsClient(awsConfig config.AWS, auth fleetmanager.Auth) (*rds.RDS, error
 	}
 	stsClient := sts.New(sess)
 
-	roleProvider := stscreds.NewWebIdentityRoleProviderWithOptions(stsClient, awsConfig.RoleARN, "",
-		&tokenFetcher{auth: auth})
+	roleProvider := stscreds.NewWebIdentityRoleProviderWithOptions(stsClient, awsConfig.RoleARN, "rds",
+		stscreds.FetchTokenPath(awsConfig.TokenFile))
```
dhaus67, line 385: So, [service account volume projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection) are special in the case that, compared to the default service account token that's mounted, they will expire. The [default value for that is 1 hour](https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken).

From fleetshard-sync's POV we don't have to take care about refreshing things, however we have to ensure we load the refreshed token within the rds client. As I see it now, the code currently will fail to provision any central after one hour of running, as the token is expired and we won't re-read it again from the mount path.

There's a couple of options we can choose from OTOH:
- in case of an error within the client, retry calling `stscreds.FetchTokenPath` again.
- upon changes within the volume mount, propagate the token to the existing client and update it, using the existing construct of [pkg/k8scfgwatch](https://github.com/stackrox/stackrox/tree/master/pkg/k8scfgwatch).

The first one is a bit awkward, since `stscreds` only provides you with the generic error type `ErrCodeWebIdentity` and nothing concretely regarding credentials. We may retry more than we'd ideally like to.

The second one will require a bit of refactoring on how the client is created (i.e. guarded behind a `sync.Once`, starting the watch handler etc.), but at the very least this is somewhat stable and long-used within stackrox for reading the admin password secret, as well as within declarative config.

Curious to hear your thoughts about this on what'd be a fitting solution for this.

dhaus67 (changes requested):

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
kovayur: Hi @dhaus67, first of all I appreciate your thoughtful review 👊.
You are absolutely right, the service account volume projection was used intentionally not only to issue a time-bound token, but also to explicitly set the `aud` claim. From my POV, this should tighten the token validation and therefore increase security. 

However, I see that the token file is retrieved every time AWS is requested. `stscreds.FetchTokenPath(awsConfig.TokenFile)`returns an implementation of the `TokenFetcher` interface, not the actual token. I believe that the `TokenFetcher.FetchToken` function is called on every call to AWS.
The token is reissued automatically by the built-in OAuth provider, so it is always up to date. 
I just checked on my dev OSD cluster, where fleetshard-sync has been running for a couple of days now, and it can provision the RDS DB just fine.

The reconciliation designed in a way that it always runs in a loop. So, in case of unsuccessful call to AWS, it should retry on the next iteration, no explicit retry logic is required. 

Does it make sense?

dhaus67: Ah, I missed that part of the code 🤦 

Well, then I guess this is obsolete, and everything is handled.

dhaus67 (approved):

docs/development/setup-aws-idp-on-osd.md
```diff
@@ -0,0 +1,70 @@
...
+The complication is that server is not reachable from AWS STS, therefore, it couldn't load the public keys to verify a JWT token. The recommended approach is to expose OAuth server's public keys in a **public** S3 bucket [2].
+
+S3 bucket name has the following pattern: `<osdClusterName>-<randomAlphanumeric>-oidc`
+- `osdClusterName` is to identify the cluster associated with this bucket
+- `randomAlphanumeric` - 32-character random string to ensure uniqueness of the S3 bucket
+- `oidc` suffix is chosen to indicate that bucket corresponds to an OAuth server. This suffix also corresponds to the pattern used in the `ccoctl` tool [3].
+
+## Manual setup (dev)
+> ❗️This instruction is intended only for OSD clusters. For local clusters, it is recommended to use the OCM token for authentication.
+
```
johannes94, line 19: Add as a Prerequisite:
```
    go install github.com/openshift/cloud-credential-operator/cmd/ccoctl@latest
```

johannes94 (changes requested):
Is it right that this will break stage if we merge before the aws-config: https://github.com/stackrox/acs-fleet-manager-aws-config/pull/75 is applied.

docs/development/setup-aws-idp-on-osd.md
```diff
@@ -0,0 +1,70 @@
...
+The main idea is the following: The OAuth server has been added to the AWS IAM configuration as an Identity Provider. Each cluster has its own server, so each must be declared in the AWS IAM configuration. AWS has a limit of 100 identity providers per account [1].
+The complication is that server is not reachable from AWS STS, therefore, it couldn't load the public keys to verify a JWT token. The recommended approach is to expose OAuth server's public keys in a **public** S3 bucket [2].
+
+S3 bucket name has the following pattern: `<osdClusterName>-<randomAlphanumeric>-oidc`
+- `osdClusterName` is to identify the cluster associated with this bucket
+- `randomAlphanumeric` - 32-character random string to ensure uniqueness of the S3 bucket
+- `oidc` suffix is chosen to indicate that bucket corresponds to an OAuth server. This suffix also corresponds to the pattern used in the `ccoctl` tool [3].
+
+## Manual setup (dev)
+> ❗️This instruction is intended only for OSD clusters. For local clusters, it is recommended to use the OCM token for authentication.
```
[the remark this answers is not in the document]
kovayur: You are right, it should be the static token here instead of OCM token. I originally thought of using an OCM token, but then gave up on that idea and forgot to change it here. It's fixed now.
The static token is loaded from secrets manager and the procedure for running fleetshard locally is already described in`./fleetshard-sync/README.md`. No extra steps are required.

docs/development/setup-aws-idp-on-osd.md
kovayur: done

docs/development/setup-aws-idp-on-osd.md
kovayur: fixed

kovayur: Fixed.

johannes94 (approved):

PR #991, opened 2023-04-27
ROX-15980 set resource requests and limits to the egress-proxy

ludydoo:
Sets the resource requests and limits for the egress-proxy.

The values were derived from the observed metrics on prometheus/grafana. These resource defaults are a bit overkill for the actual observed usage. I was not comfortable putting less than this for a production deployment. 

It also seems like the `values.yaml` file was ignored for the `tenant-resources`. This PR also adds changes to use these values by default and apply overrides on top of them. 

The current cluster configuration deploys 1 replica of the `egress-proxy`, so I've changed the value to reflect that as well.

### DIFF
--- a/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
+++ b/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
@@ -2,6 +2,13 @@
 egressProxy:
   image: ubuntu/squid:5.2-22.04_beta
   replicas: 2
+  resources:
+    limits:
+      cpu: 100m
+      memory: 128Mi
+    requests:
+      cpu: 200m
+      memory: 256Mi
 
 labels: {}
 annotations: {}
--- a/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
+++ b/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
@@ -1,6 +1,6 @@
 egressProxy:
   image: ubuntu/squid:5.2-22.04_beta
-  replicas: 1
+  replicas: 3
   resources:
     requests:
       cpu: 100m
--- a/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
+++ b/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
@@ -1,7 +1,6 @@
-
 egressProxy:
   image: ubuntu/squid:5.2-22.04_beta
-  replicas: 2
+  replicas: 1
   resources:
     limits:
       cpu: 100m
@@ -9,6 +8,5 @@ egressProxy:
     requests:
       cpu: 200m
       memory: 256Mi
-
 labels: {}
 annotations: {}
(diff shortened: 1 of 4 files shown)


fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
```diff
@@ -1,7 +1,12 @@
...
 egressProxy:
   image: ubuntu/squid:5.2-22.04_beta
-  replicas: 2
-
+  replicas: 1
+  resources:
+    limits:
+      cpu: 100m
+      memory: 128Mi
+    requests:
```
kylape, line 8: I think you got the request and limit values backwards.

ludydoo: 🤦

fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
```diff
@@ -1,7 +1,12 @@
-
 egressProxy:
   image: ubuntu/squid:5.2-22.04_beta
-  replicas: 2
-
+  replicas: 1
```
kylape, line 3: this doesn't need to be fixed in this PR, but this does make me wonder if we should actually run two replicas in the cloud service for all the usual reasons.

kylape (approved):
lgtm!

fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
porridge: Or even 3.

ludydoo: @kylape @porridge perhaps with a nodeAntiAffinity on other egress proxies (preferredDuringScheduling) ?

fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
```diff
@@ -1,7 +1,12 @@
...
+  resources:
+    requests:
+      cpu: 100m
+      memory: 128Mi
+    limits:
+      cpu: 200m
+      memory: 256Mi
```
porridge, line 10: How about we use requests==limits at least until we have more replicas? We'll likely be setting resources for a bunch of pods in the next rollouts (see sibling PRs) so the risk of facing evictions will be higher than normal and I'm a bit worried with one replica this may cause service degradation for non-trivial central configurations.

porridge (approved):

fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
ludydoo: I can add the replicas in this PR perhaps?

porridge: I wouldn't mind.

kurlov (approved):

kylape (approved):

PR #998, opened 2023-04-28
Add admin API client

SimonBaeumer:
## Description

Add Admin API command. Currently only listing Centrals is supported, eventually more commands are added.
This is a pre-requisiste to add a `set-version` command.

## Test manual

 ```
# Create an ACS instance and query via admin API on a local cluster
$ export OCM_TOKEN=<refresh token>
$ export STATIC_TOKEN=<token>
$. /fleet-manager central create --name=test1 --region=standalone --provider=standalone
$ ./fleet-manager admin centrals list
```

### CODE REVIEW

File: internal/dinosaur/pkg/cmd/admin/centrals/cmd.go [hunk 1]
```diff
@@ -0,0 +1,24 @@
+// Package centrals contains the admin central CLI interface.
+package centrals
+
+import "github.com/spf13/cobra"
+
+const (
+	apiErrorMsg = "%s Admin Central failed: To fix this ensure you are authenticated, fleet-manager endpoint is configured and reachable. Status Code: %s."
```
kurlov, line 7: Nitpicking:
- let;'s use `%q` for command name to highlight that it's a name
- should Admin be lowercase?

internal/dinosaur/pkg/cmd/fleetmanagerclient/client.go
```diff
@@ -13,16 +13,59 @@ import (
...
+
+	ocmClientStaticToken         *fleetmanager.Client
+	singletonStaticTokenInstance sync.Once
 )
 
 const (
 	defaultFleetManagerEndpoint = "http://localhost:8000"
 	fleetManagerEndpointEnvVar  = "FMCLI_FLEET_MANAGER_ENDPOINT"
 	ocmRefreshTokenEnvVar       = "OCM_TOKEN"
+	staticTokenEnvVar           = "STATIC_TOKEN"
```
kurlov, line 26: I think it's better to reuse const from `auth_static_token.go` to avoid potential mismatch in the future
https://github.com/stackrox/acs-fleet-manager/blob/fc0f2be99a1d2929606cf984754fa2e857acb7cf/pkg/client/fleetmanager/auth_static_token.go#L10

kurlov (approved):
Looks great! Just few nitpicks which totally fine to ignore

internal/dinosaur/pkg/cmd/fleetmanagerclient/client.go
SimonBaeumer: This constant does not fit here because it describes the name of the authentication method and not the env variable's name.
Nevertheless the constant can be used for the initialization of the clients to reduce duplicate code. 
The environment variable names are hardcoded in struct tags and only accessible via reflection.

internal/dinosaur/pkg/cmd/fleetmanagerclient/client.go
```diff
@@ -13,16 +13,59 @@ import (
...
 )
 
-// AuthenticatedClientWithOCM returns a rest client to the fleet-manager and receives the OCM refresh token.
+// AuthenticatedClientWithStaticToken returns a rest ocmClient to the fleet-manager using a static token.
```
johannes94, line 29:
```suggestion
// AuthenticatedClientWithStaticToken returns a rest client for fleet-manager API using a static OCM token for authentication.
```

I'm still struggling with calling this an ocmClient. In my opinion OCM is an API/CLI tool we depend on. If something is called ocmClient, I expect it to have the sole purpose of helping me communicate with the OCM API. But this is returning a client for fleet-manager API with a preconfigured `STATIC_TOKEN` that should be set to a valid static token for ocm.

internal/dinosaur/pkg/cmd/fleetmanagerclient/client.go
```diff
@@ -13,16 +13,59 @@ import (
...
+		ocmClientStaticToken, err = fleetmanager.NewClient(fleetManagerEndpoint, auth)
+		if err != nil {
+			glog.Fatalf("Failed to create connection: %s", err)
+			return
+		}
+	})
+
+	// sleep timer necessary to avoid "token issued in future" errors for time lags between fleet-manager running on a
+	// local VM and the OCM server.
+	if fleetManagerEndpoint == defaultFleetManagerEndpoint {
```
johannes94, line 62: I don't think we should have this in the code base, because it seems to be an error that only occured on you local system setup.

internal/dinosaur/pkg/cmd/fleetmanagerclient/client.go
```diff
@@ -13,16 +13,59 @@ import (
...
+		time.Sleep(5 * time.Second)
+	}
+	return ocmClientStaticToken
+}
+
+// AuthenticatedClientWithOCM returns a rest ocmClient to the fleet-manager and receives the OCM refresh token.
```
johannes94, line 68: Same as stated above, I wouldn't call this an ocmClient. It's a fleet-manager client using OCM to obtain a valid authentication token from OCM using an OCM refresh token

johannes94 (changes requested):

internal/dinosaur/pkg/cmd/admin/centrals/cmd.go
SimonBaeumer: Changed admin to lower case.

> let's use %q for command name to highlight that it's a name

I don't think so, %q is used whenever a value can be empty to indicate that something is variable. In this case it is a static word templated.

internal/dinosaur/pkg/cmd/admin/centrals/list.go
```diff
@@ -0,0 +1,42 @@
...
+	admin "github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/admin/private"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/cmd/fleetmanagerclient"
+	"github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager"
+)
+
+// NewListCommand creates a new command for listing centrals.
+func NewListCommand() *cobra.Command {
```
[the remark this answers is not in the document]
SimonBaeumer: No hard opinion. Your suggestion is more readable, used it.

internal/dinosaur/pkg/cmd/admin/centrals/update.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/cmd/fleetmanagerclient"
+	"github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager"
+)
+
+const (
+	centralOperatorVersion = "central-operator-version" // TODO(sbaumer): move operator version to a separate command.
+)
+
+// NewUpdateCommand creates a new command for updating centrals.
+func NewUpdateCommand() *cobra.Command {
```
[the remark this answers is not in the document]
SimonBaeumer: removed the upgrade command from this PR.

internal/dinosaur/pkg/cmd/admin/centrals/update.go
```diff
@@ -0,0 +1,62 @@
+package centrals
```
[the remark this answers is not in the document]
SimonBaeumer: removed the upgrade command from this PR.

internal/dinosaur/pkg/cmd/fleetmanagerclient/client.go
SimonBaeumer: renamed

internal/dinosaur/pkg/cmd/fleetmanagerclient/client.go
johannes94: For me it works with following setup:
```
# run local fleet-manager in a local shell
./fleet-manager serve

# run the CLI command as fast as possible in another shell
OCM_TOKEN=$(ocm token --refresh) ./fleet-manager central create --name=test1 --region=standalone --provider=standalone

# same for admin commands
RHOAS_TOKEN=$(rhoas authtoken) ./fleet-manager admin centrals list
```

johannes94 (changes requested):
Good work 👍 just two more things to consider from my side:

- removing the 5 sec sleep before merging or make it not default
- `admin centrals` for admin CLI command vs `central` for non-admin, I'd prefer plural for both

PR #1007, opened 2023-05-03
fix e2e admin tests by reducing required resources

johannes94:
## Description
The e2e admin tests fail because they set resource requests and limits for central to values that did not allow central pods to be created properly. This PR adjusts those resources so that they are close to the default values used in the "non-admin" tests. With this the test should work again.

- ~~[ ] Added test description under `Test manual`~~

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -319,12 +319,12 @@ var _ = Describe("Central", func() {
 
 		centralResources := private.ResourceRequirements{
 			Requests: map[string]string{
-				corev1.ResourceCPU.String():    "501m",
-				corev1.ResourceMemory.String(): "201M",
+				corev1.ResourceCPU.String():    "210m",
+				corev1.ResourceMemory.String(): "210M",
 			},
 			Limits: map[string]string{
-				corev1.ResourceCPU.String():    "502m",
-				corev1.ResourceMemory.String(): "202M",
+				corev1.ResourceCPU.String():    "310m",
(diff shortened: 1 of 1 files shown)


johannes94:
> Was it just CPU that was causing issues? Because it looks like memory requests were actually increased.

True, I mixed something up here. I put values that are basically the `defaultValue + 10`. I was unsure what is causing it because we're missing meaningful logs. I'll undo the changes for memory.

kylape:
Do you know if the cpu limits need to be reduced?  I worry a bit that CI tests would take longer since the CPU limits will be lowered, but I'm not familiar with the details around the issues.

johannes94:
> Do you know if the cpu limits need to be reduced? I worry a bit that CI tests would take longer since the CPU limits will be lowered, but I'm not familiar with the details around the CI issues.

Perhaps we could also increase the resources we require in the openshift-ci config. I don't think that reducing the limit in this test will have a big impact on time required for an e2e tests. It's just the resources for the central/scanner pods. In this test they are basically created and deleted nothing more happens. So having lower CPU limits shouldn't be a problem because there's no processing happening in this pods.

kylape (approved):

johannes94:
@kylape failing CI showed that actually the memory requests were to low and not the cpu requests to high (or maybe both). I think the initial approach with getting all values very similar to what the "default" values are worked fine so I reverted the "revert" of the memory changes.

PR #1028, opened 2023-05-11
ROX-16506: Add Fleet-Manager OperatorImage configuration

SimonBaeumer:
## Description

This PR adds the fleet-manager logic to use fleet-manager to manage Central and Operator versions.

**Workflow:**

The workflow implementation is based declaratively in a sense that Fleet-Manager defines the desired state, but does not take care of status reporting. Status reporting is implemented via alerting later.

 1. CentralRequest is created and fetched default image from the default version table (table to be renamed by @johannes94 ), see [here](https://github.com/stackrox/acs-fleet-manager/pull/1028/files#diff-080490e019286b765c01a138eab3a47c8dd81e8c5c942fc8545be9d8fa8ae4e9R252-R257)
 2. `operator_image` field is set on creation
 3. Fleetshard-Sync sees new operator image in its runtime
 4. Fleetshard-Sync creates for each operator-image a new deployment

**Changes different from design document:**

 - `previous_*` fields were removed
 - `actual_*` fields were removed
 - `operator_image` field was added

_Status reporting_
Status reporting needs separate implementation and design to have a sustainable solution to query states in fleet-manager. My biggest concern is that structured status reporting creates an overhead in adding new statuses to fleet-manager,i .e. by adding new database fields and reconcilation logic.

_Previous Image_
To rollback an instance we need to store the previous revision of a Central. Instead of as initially saving only the `previous_image` the whole CentralRequest object can be revisioned.  (Ticket: https://issues.redhat.com/browse/ROX-8987)
This may need a small implementation discussion.

_End to end tests_

End-to-end, integration and unit-tests are added in a later PR. Currently all features are hidden behind a feature flag.

**Removals:**

 - Some operator version related code was removed which was not used anymore because of the changes to the fields.
 - Follow up removal: acceptance manager
 - Follow up removal: operator & central version legacy logic

**DoD and follow-ups:**

 - [x] Set default version when instance is created
 - [x] Add image fields to Central table
 - [x] Fleetshard should see new fields and install operator based on these
 - [x] Migrate existing Centrals (ticket: https://issues.redhat.com/browse/ROX-17759)
 - [x] Fetch operator images from Fleet-Manager
 - [x] Keep history of previously applied operator image to an instance for possible rollbacks (https://issues.redhat.com/browse/ROX-8987)
 - [x] use correct label selector to support multiple operator instances (ticket: https://issues.redhat.com/browse/ROX-17718)

## Test manual

 - Install fleet-manager locally via `./scripts/install_operator_canary.sh`
 - Create stackrox-operator namespace
 - Add `quay-ips` secret manually to stackrox-operator namespace
 - Create Central via `./scripts/create-central.sh`
 - Run `kubectl get pods -n stackrox-operator` to see new installed operators


internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
```diff
@@ -0,0 +1,93 @@
...
+				return err
+			}
+			if err := dropIfColumnExists(tx, &CentralRequest{}, "central_operator_upgrading"); err != nil {
+				return err
+			}
+			if err := dropIfColumnExists(tx, &CentralRequest{}, "central_upgrading"); err != nil {
+				return err
+			}
+
+			if err := tx.Create(&dbapi.CentralDefaultVersion{Version: updatedDefaultVersion}).Error; err != nil {
```
SimonBaeumer, line 83: Update existing Centrals with operator_image

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -158,6 +158,7 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 			Annotations: map[string]string{
 				managedServicesAnnotation: "true",
 				orgNameAnnotationKey:      remoteCentral.Spec.Auth.OwnerOrgName,
+				// TODO(ROX-17718): Set reconciler selection label for Central
```
ludydoo, line 161: Does this need to be addressed in this PR or is this for later?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -957,8 +958,7 @@ func (r *CentralReconciler) chartValues(remoteCentral private.ManagedCentral) (c
 
 func (r *CentralReconciler) shouldSkipReadyCentral(remoteCentral private.ManagedCentral) bool {
 	return r.wantsAuthProvider == r.hasAuthProvider &&
-		isRemoteCentralReady(remoteCentral) &&
-		remoteCentral.Spec.Versions.ActualVersion == remoteCentral.Spec.Versions.DesiredVersion
+		isRemoteCentralReady(remoteCentral)
```
ludydoo, line 961: Can you explain why it's necessary to remove the version check ?

internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
```diff
@@ -0,0 +1,93 @@
+package migrations
+
+// Migrations should NEVER use types from other packages. Types can change
```
ludydoo, line 3: According to this comment, there should be no imports in this file right?

internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
```diff
@@ -0,0 +1,93 @@
...
+		Routes                        api.JSON   `json:"routes"`
+		RoutesCreated                 bool       `json:"routes_created"`
+		Namespace                     string     `json:"namespace"`
+		RoutesCreationID              string     `json:"routes_creation_id"`
+		DeletionTimestamp             *time.Time `json:"deletionTimestamp"`
+		Central                       api.JSON   `json:"central"`
+		Scanner                       api.JSON   `json:"scanner"`
+		OperatorImage                 string     `json:"operator_image"`
+	}
+	migrationID := "20230510000000"
```
ludydoo, line 55: Maybe the migration ID should be updated to e.g. `20230616...`

internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
ludydoo, line 83: Is this to update the `default central version` ? Should this be a `tx.Update` ?

internal/dinosaur/pkg/services/clusters.go
```diff
@@ -730,35 +728,10 @@ func buildClusterSpec(cluster *api.Cluster) *types.ClusterSpec {
...
-	if err != nil {
-		return false, fmt.Errorf("retrieving ready central operator versions: %w", err)
-	}
-	for _, version := range readyDinosaurOperatorVersions {
-		if version.Version == dinosaurOperatorVersion {
-			return true, nil
-		}
-	}
-	return false, nil
+	return true, nil
```
ludydoo, line 731: Why remove this ?

internal/dinosaur/pkg/services/clusters.go
```diff
@@ -730,35 +728,10 @@ func buildClusterSpec(cluster *api.Cluster) *types.ClusterSpec {
...
-				if kversion.Version == dinosaurVersion {
-					kVvalid = true
-					break
-				}
-			}
-			return kVvalid, nil
-		}
-	}
-	return false, nil
+	return true, nil
```
ludydoo, line 736: And this ?

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -244,6 +250,14 @@ func (k *dinosaurService) RegisterDinosaurJob(dinosaurRequest *dbapi.CentralRequ
 		return err
 	}
 
+	if features.TargetedOperatorUpgrades.Enabled() {
```
ludydoo, line 253: What happens when this feature flag is disabled?

internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
SimonBaeumer: This must be done later to not have Centrals in an invalid state. By updating Central operator images direclty without using them the desired state can never match cluster state.

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: For later, this PR only introduces the fields. 
The main reason is too separate the logic of handling the image ref and introducing the data model.

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: Because it is not necessary anymore when using the `operator_image` field, it is cached like every other resource.
When the `operator_image` changes the reconciler will run again as it did without the check.

internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
SimonBaeumer: No, imports are okay, only data models and types should not be imported from another package. That's why the struct of the database table is implemented in every migration function again.
Do you see more concerns?

internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
SimonBaeumer: done

internal/dinosaur/pkg/migrations/20230510000000_add_acs_operator_version.go
SimonBaeumer: No, the default version is managed as a new entry. The default version service will always return the latest default version. 
@johannes94 refactor's this table to a defaults table which might reflect your suggestion.

internal/dinosaur/pkg/services/clusters.go
SimonBaeumer: The functions were removed because they referenced struct fields which were removed like `AcutalOperatorVersion`. The alternative to deleting the logic would be commenting out, I decided to go with removing because the logic is not used by us.

All clusters we add to fleet-manager are added in `ready` state and do not transition in the state machine.

internal/dinosaur/pkg/services/clusters.go
SimonBaeumer: See above

internal/dinosaur/pkg/services/dinosaur.go
SimonBaeumer: Nothing, the logic is the same as today. A `CentralRequest` does not set the `operator_image` if it is not used. This would be an invalid state because the `operator_image` is never a state reconciled on the cluster.

Worst case, `CentralRequests` have version `image:4.1` in its database, we introduce the change but `4.1` was released and would accidentally rolling back existing Central instances.

kurlov: I think it is fine to drop feature flag check here

scripts/install_operator_canary.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+export INSTALL_OLM="false"
+export INSTALL_OPERATOR="false"
+make deploy/bootstrap
+make deploy/dev
+kubectl set env -n acsms deploy/fleetshard-sync FEATURE_FLAG_UPGRADE_OPERATOR_ENABLED=true
```
kurlov, line 8: 🔥

kurlov (commented):
Looks great! Just a few comments though

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -69,6 +69,10 @@ type ACSOperatorImage struct {
 
 // InstallOrUpgrade provisions or upgrades an existing ACS Operator from helm chart template
 func (u *ACSOperatorManager) InstallOrUpgrade(ctx context.Context, images []ACSOperatorImage) error {
+	if len(images) == 0 {
```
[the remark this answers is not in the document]
SimonBaeumer: Removed it

fleetshard/pkg/runtime/runtime.go
```diff
@@ -229,23 +230,17 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
 	}
 }
 
-func (r *Runtime) upgradeOperator() error {
+func (r *Runtime) upgradeOperator(list private.ManagedCentralList) error {
+	var operatorImages []operator.ACSOperatorImage
+	for _, central := range list.Items {
+		operatorImages = append(operatorImages, operator.ACSOperatorImage{
+			Image:      central.Spec.OperatorVersion.OperatorImage,
+			InstallCRD: false,
```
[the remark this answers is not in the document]
SimonBaeumer: Done, `TODO(ROX-15080)` added

scripts/install_operator_canary.sh
SimonBaeumer: 🔥🔥🔥

internal/dinosaur/pkg/services/dinosaur.go
SimonBaeumer: No, this flag cannot be removed. As described above without the feature flag a state which is never reached is persistent in the database.

ebensh:
Two higher level questions that might've been answered elsewhere:

> Status reporting is implemented via alerting later.
1. Why is status tracking done via alerting instead of a feedback loop?

> Previous Image
> To rollback an instance we need to store the previous revision of a Central. Instead of as initially saving only the previous_image the whole CentralRequest object can be revisioned. (Ticket: https://issues.redhat.com/browse/ROX-8987)
This may need a small implementation discussion.
2. If we're considering storing the previous image requested as well then is this similar to the DesiredCentralVersion/ActualCentralVersion but with s/Version/OperatorImage/ ? If so ignore my comment about collapsing the OperatorVersion object into a single string

internal/dinosaur/pkg/services/clusters.go
SimonBaeumer: The usages are maybe hidden because of the DI container. Anyway, the usage is removed in a follow-up PR to keep this focused on the operator image changes: https://github.com/stackrox/acs-fleet-manager/pull/1097

internal/dinosaur/pkg/services/clusters.go
SimonBaeumer: See pull-request: https://github.com/stackrox/acs-fleet-manager/pull/1097

SimonBaeumer:
> Why is status tracking done via alerting instead of a feedback loop?

Because currently we do not report statuses to fleet-manager, except the state of the state machine. Keeping the state in grafana a canonical location for the systems health status is used.
Eventually we can add more status reporting to fleet-manager, but may want to design something which is sustainable and scalable. Introducing a new field for every status is too much maintenance overhead.

> If we're considering storing the previous image requested as well then is this similar to the DesiredCentralVersion/ActualCentralVersion but with s/Version/OperatorImage/ ? If so ignore my comment about collapsing the OperatorVersion object into a single string

Focusing for restoring the previous state only on one field does not scale. Instead a ticket was created to rollback the entire central_request to a previous state. See https://issues.redhat.com/browse/ROX-17758

openapi/fleet-manager-private.yaml
```diff
@@ -285,8 +283,8 @@ components:
                   properties:
                     host:
                       type: string
-                versions:
-                  $ref: "#/components/schemas/ManagedCentralVersions"
+                operatorVersion:
+                  $ref: "#/components/schemas/OperatorVersion"
```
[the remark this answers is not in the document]
SimonBaeumer: Good point, changed it to `OperatorImage` as a string field

internal/dinosaur/pkg/presenters/admin_dinosaur.go
```diff
@@ -47,23 +47,22 @@ func PresentDinosaurRequestAdminEndpoint(request *dbapi.CentralRequest, _ accoun
...
+		CloudProvider:  request.CloudProvider,
+		MultiAz:        request.MultiAZ,
+		Region:         request.Region,
+		Owner:          request.Owner,
+		Name:           request.Name,
+		Host:           request.GetUIHost(), // TODO(ROX-11990): Split the Host in Fleet Manager Public API to UI and Data hosts
+		CreatedAt:      request.CreatedAt,
+		UpdatedAt:      request.UpdatedAt,
+		FailedReason:   request.FailedReason,
+		InstanceType:   request.InstanceType,
```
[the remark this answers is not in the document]
SimonBaeumer: The API is not yet implemented and introduced at a later point. Currently it is only managed via the default images.

kurlov (approved):

SimonBaeumer:
@ebensh 
> General q about latency - how long does it take to provision an operator the first time its image is requested?

~1min

> Briefly about security, this opens the door to arbitrary images being installed in the cluster; can you add a comment to the Central PUT request (which I'm pretty sure is disabled?) to ensure that this field cannot be changed by the client? It would be bad if they could insert arbitrary images to be installed instead of a version number

Good catch! This is not possible, the public API does not support the field. See the public API's struct:
https://github.com/stackrox/acs-fleet-manager/blob/caae304f6a042e2c6c45fd1407e4c17f2452c6e3/internal/dinosaur/pkg/api/public/model_central_request_payload.go#L14-L14

ebensh (approved):

PR #1030, opened 2023-05-11
Change secret key to alertmanager.yaml

kylape:
Prometheus operator specifically looks for `alertmanager.yaml`, NOT `alertmanager.yml`.  If it cannot find that specific key, it will use a default/dummy congiruation instead, effectively disabling alertmanager.


stehessel (approved):
At we know that the dead man switch works.

PR #1034, opened 2023-05-15
ROX-17083: Send stage alerts directly to Slack

kylape:
## Description
Only critical production alerts get sent to PagerDuty.

This commit sets up the Slack receiver, which includes formatting the Slack messages.  A few notes on the format:

* Due to the use of a shared webhook config, the icon, bot name, and target channel are not configurable.
* Only the alert name is in the message title (as well as whether the alert is firing or resolved)
* The message body has four sections:
  * Cluster name
  * Namespace
  * Summary
  * Description
* Two action buttons: SOP and Prometheus URL

~~- [ ] Unit and integration tests added~~ alertmanager config linter to come

## Test manual

Created a very simple prometheus exporter that increments a counter when an enpoint is invoked:

```python
from flask import Flask, make_response, request, redirect, url_for, abort, session, jsonify
from prometheus_client import Counter, make_wsgi_app
from werkzeug.middleware.dispatcher import DispatcherMiddleware

c = Counter('calls', 'Count of calls to endpoint')
app = Flask(__name__)

@app.route("/inc")
def inc():
    c.inc()
    return ""

# Add prometheus wsgi middleware to route /metrics requests
app.wsgi_app = DispatcherMiddleware(app.wsgi_app, {
    '/metrics': make_wsgi_app()
})
```

Ran a local prometheus with this config:

```yaml
# my global config
global:
  scrape_interval: 1s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  evaluation_interval: 5s # Evaluate rules every 15 seconds. The default is every 1 minute.
  # scrape_timeout is set to the global default (10s).

# Alertmanager configuration
alerting:
  alertmanagers:
  - static_configs:
    - targets:
      - localhost:9093

# Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
rule_files:
- "rules.yml"

# A scrape configuration containing exactly one endpoint to scrape:
# Here it's Prometheus itself.
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name: "prometheus"

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    static_configs:
    - targets: ["localhost:5000"]
```

and rules file 

```yaml
groups:
  - name: rules
    rules:
      - alert: FAKE-RateTooHigh
        expr: rate(calls_total[15s]) > 0.5
        for: 15s
        labels:
          severity: critical
          observability: managed-rhacs
          rhacs_environment: stage
          rhacs_cluster_name: fake-acs-stage-dp-03
          namespace: test
        annotations:
          summary: "Counter rate too high"
          description: "THIS IS A FAKE ALERT to test slack alerting.  Count rate is at {{ $value }}. This is a longer message because the description is supposed to be longer!"
          sop_url: "dummy"
```

Ran a local alertmanager with a config pulled from the changes in the PR and the Slack API URL.

Once the receiver config was finalized, I needed to validate the helm chart would render the secret correctly.  To do this, I added to the `slack_api_url` secret to the `observability` secret in the dev AWS account.  I then ran `terraform_cluster.sh` to see that the chart was rendered correctly with the value pulled from AWS Secret Manager.

```
# To run tests locally run:
```

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
@@ -8,24 +8,30 @@ stringData:
     global:
       resolve_timeout: 5m
     route:
-      receiver: default-receiver
+      receiver: slack
       repeat_interval: 12h
       routes:
       - receiver: managed-rhacs-pagerduty
         match:
           observability: managed-rhacs
           severity: critical
+          rhacs_environment: prod
       - receiver: managed-rhacs-deadmanssnitch
         repeat_interval: 5m
         match:
           alertname: DeadMansSwitch
           observability: managed-rhacs
     receivers:
-    - name: default-receiver
     - name: managed-rhacs-pagerduty
       pagerduty_configs:
       - service_key: {{ .Values.pagerduty.key | quote }}
     - name: managed-rhacs-deadmanssnitch
       webhook_configs:
       - url: {{ .Values.deadMansSwitch.url | quote }}
+    - name: slack
+      slack_configs:
+      - send_resolved: true
+        api_url: {{ .Values.slack.apiURL | quote }}
+        title_link: '{{ "{{" }} template "slack.default.titlelink" . {{ "}}" }}'
+{{ .Files.Get "slack-receiver-config.yaml" | indent 8 }}
 type: Opaque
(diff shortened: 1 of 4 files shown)


dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
```diff
@@ -8,24 +8,30 @@ stringData:
...
     route:
-      receiver: default-receiver
+      receiver: slack
       repeat_interval: 12h
       routes:
       - receiver: managed-rhacs-pagerduty
         match:
           observability: managed-rhacs
           severity: critical
+          rhacs_environment: prod
```
stehessel, line 18: Not all metrics have this label - only ones scraped from pods that we deploy. We might be able to add this label to the external labels, although we need to check if one can match on external labels here.
Otherwise I'm afraid we need some more templating here, it's surprisingly difficult to add a generic label to all Prometheus metric series.

kylape:
Pivoting approach to https://github.com/stackrox/acs-fleet-manager/pull/1039

PR #1038, opened 2023-05-16
ROX-16735: Add `vector.dev` Helm chart

0x656b694d:

## Test manual

```
# To run tests locally run:
```


0x656b694d:
* main
  * **PR #1038** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1038" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1038?utm_source=stack-comment).

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -91,3 +91,43 @@ logging:
...
+        filename_extension: "json"
+        batch:
+          timeout_secs: 10
+        encoding:
+          codec: "json"
+  env:
+    - name: AWS_ACCESS_KEY_ID
+      valueFrom:
+        secretKeyRef:
+          name: cloudwatch
```
mtodor, line 127: We will not use `cloudwatch` credentials to write access logs. Every cluster will have its own access keys for a bucket dedicated to it.

scripts/lib/helm.sh
```diff
@@ -6,6 +6,11 @@ function invoke_helm() {
     local -r release="${1}"
     shift
 
+    helm repo add vector "https://helm.vector.dev"
```
mtodor, line 9: For me, it's not clear to me in which namespace aggregator will be deployed.

I think it will be `rhacs`. Is that fine for us, or should it be dedicated namespaces as we have for observability?

mtodor (changes requested):
We should adjust a few configurations for the vector aggregator.

dp-terraform/helm/rhacs-terraform/values.yaml
0x656b694d: Oh!

scripts/lib/helm.sh
0x656b694d: I mean it will be `rhacs` if we don't change it somehow.

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -91,3 +91,78 @@ logging:
...
+    annotations:
+      rhacs.redhat.com/cluster-name: ""
+      rhacs.redhat.com/environment: ""
+      service.beta.openshift.io/serving-cert-secret-name: rhacs-vector-tls-secret
+  affinity:
+    podAntiAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        - labelSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/name
```
mtodor, line 107: This is great! I would suggest one change.

We should use `podLabels` to add our own labels. And then use that for anti-affinity. The reason for that is - to avoid a potential collision with matching labels with pods created by the logging operator. Because logging operator will also use vector: https://docs.openshift.com/container-platform/4.13/logging/v5_6/logging-5-6-release-notes.html#logging-5-6-dep-notice_logging-5.6-release-notes

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -91,3 +91,78 @@ logging:
...
+        region: ""
+        bucket: ""
+        key_prefix: '{{ "{{" }} .tenant_id {{ "}}" }}/%F/'
+        inputs: ["http_server"]
+        compression: none
+        filename_extension: "json"
+        healthcheck:
+          enabled: false
+        batch:
+          timeout_secs: 10
```
mtodor, line 150: I have looked a bit at the configuration of helm:
https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/README.md

And S3 sink:
https://vector.dev/docs/reference/configuration/sinks/aws_s3/#buffer.max_size

From what I see, enabling `persistence` will create PV/PVC, but it will not configure the application to use it. So we have to configure Vector accordingly to use that persistence.

Also, size of the buffer (including disk) has to be at least 256MB - https://vector.dev/docs/reference/configuration/sinks/aws_s3/#buffer.max_size

My proposal would be:
- 270 MB buffer - 300 MB disk
- sync every 1 min
- max batch size 2.5MB
- if buffer is full, block requests -> this should pop-up errors in central

```suggestion
        batch:
          timeout_secs: 60
          max_bytes: 2621440
        buffer:
          type: "disk"
          max_size: 283115520
          when_full: "block"
```

mtodor (commented):
Nice progress! I've added a few additional suggestions related to the configuration of Vector. Please look at it by yourself if something can be even more improved.

Maybe there are some good suggestions here: https://vector.dev/docs/setup/going-to-prod/arch/aggregator/

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -91,3 +91,85 @@ logging:
...
+    app: rhacs-vector
+  affinity:
+    podAntiAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        - labelSelector:
+            matchExpressions:
+              - key: app
+                operator: In
+                values:
+                  - vector
```
mtodor, line 112: Should we use `rhacs-vector` instead of `vector`? Since we have config:
```
  podLabels:
    app: rhacs-vector
```

mtodor (commented):
I think we still need a minor tweak for `matchExpressions`.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -157,7 +159,14 @@ invoke_helm "${SCRIPT_DIR}" rhacs-terraform \
   --set observability.observatorium.metricsClientId="${OBSERVABILITY_OBSERVATORIUM_METRICS_CLIENT_ID}" \
   --set observability.observatorium.metricsSecret="${OBSERVABILITY_OBSERVATORIUM_METRICS_SECRET}" \
   --set observability.pagerduty.key="${OBSERVABILITY_PAGERDUTY_ROUTING_KEY}" \
-  --set observability.deadMansSwitch.url="${OBSERVABILITY_DEAD_MANS_SWITCH_URL}"
+  --set observability.deadMansSwitch.url="${OBSERVABILITY_DEAD_MANS_SWITCH_URL}" \
+  --set vector.enabled=true \
```
mtodor, line 163: @0x656b694d let's set this to `false` - until AWS Terraform config is ready. So that we can merge this PR without breaking staging. And when AWS Terraform is ready and applied, we can just switch it on.
```suggestion
  --set vector.enabled=false \
```

mtodor (approved):
We can go forward and merge this. The next step will be to adjust some settings and enable vector.

PR #1039, opened 2023-05-17
ROX-17083: Make PagerDuty the default receiver

kylape:
## Description
This is part of a larger effort to stop treating stage alerts as critical alerts in PagerDuty.  The specific changes being made in this PR:

* Pass the `severity` parameter in the PagerDuty receiver configuration. This will be used in production to capture non-critical production events in PagerDuty without triggering a critical notification to on-call engineers.  (These events are currently dropped in Alertmanager)
* Switch to the PagerDuty v2 Events API.  This is required for PD to recognize the `severity` parameter.  This is done by switching from `service_key` to `routing_key`.
* Updated the key in `terraform_cluster.sh` (and thus AWS Secrets Manager) to reflect the routing key change.
* Made PagerDuty the default receiver in the Alertmanager config.  This will send non-critical alerts to PagerDuty, treating it as a hub for all events across all data plane clusters.

It's implied that the PagerDuty routing key for the stage environment will be different from the one in production.  This is so the stage service can be configured to force all incidents to be considered low priority to avoid paging on-call engineers.

## Test manual

* get the `routing_key`
* set the template string in the PD severity `{{ or .GroupLabels.severity 'info' }}`
* render helm chart and get `alertmanager.yaml`
* fire up test flask app (prometheus exporter)
* configure alert to be a non-critical severity
* fire up alertmanager and prometheus locally
* trigger alert with curl command
* this should create an incident in PD and send to slack, but NOT notify me personally
* let it resolve
* configure the alert to be a critical severity
* trigger alert with curl command
* this should page me as well as create a slack message (won't normally happen
  after updating notifcation config in stage PD service)
* let it resolve

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
@@ -20,6 +20,21 @@ stringData:
     - name: managed-rhacs-pagerduty
       pagerduty_configs:
       - routing_key: {{ .Values.pagerduty.key | quote }}
+        {{- /*
+          We want the severity to be based on the severity label coming from the
+          alert itself. If there is no severity label common to the group of
+          alerts, then default to info. That looks like:
+              `or .GroupLabels.severity "info"`
+          in Go templating. To properly escape for Helm templating, the Helm
+          templating engine needs to output the literal string `{{`, since
+          Alertmanager templating syntax is the same as Helm. To do that,
+          the expression `{{` is used inside the double bracket syntax for
+          evaluating Go template expressions. Thus: `{{ "{{" }}`.
+
+          The inner double quotes work because Helm evaluates the expression
+          that includes the inner double quotes before the document is parsed
+          as yaml.
+        */}}
         severity: "{{ "{{" }} or .GroupLabels.severity \"info\" }}"
     - name: managed-rhacs-deadmanssnitch
       webhook_configs:
--- a/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
@@ -8,23 +8,19 @@ stringData:
     global:
       resolve_timeout: 5m
     route:
-      receiver: default-receiver
+      receiver: managed-rhacs-pagerduty
       repeat_interval: 12h
       routes:
-      - receiver: managed-rhacs-pagerduty
-        match:
-          observability: managed-rhacs
-          severity: critical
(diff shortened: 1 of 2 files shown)


kylape:
I went ahead and added the routing key to the Secret Manager for stage (but not for prod).

dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
```diff
@@ -8,23 +8,19 @@ stringData:
...
         match:
           alertname: DeadMansSwitch
           observability: managed-rhacs
     receivers:
-    - name: default-receiver
     - name: managed-rhacs-pagerduty
       pagerduty_configs:
-      - service_key: {{ .Values.pagerduty.key | quote }}
+      - routing_key: {{ .Values.pagerduty.key | quote }}
+        severity: "{{ "{{" }} or .GroupLabels.severity \"info\" }}"
```
stehessel, line 23: Is this even correct yaml? Looks like some `"` are not correct escaped here? Can you maybe explain what this is supposed to do? How are `info` alerts handled?

stehessel:
I think we have to notify on-call engineers + SRE to set up their high/low urgency notification settings correctly. Otherwise it might have the opposite effect of *more* pages over the weekend, when someone has low urgency linked to their phone and we now sent alerts for `warning`/`info`.

kylape:
That's true for the production environment.  For stage, the schedule is just me, so no one will get paged for those.  For this weekend, I can contact the on-call engineer to have them set up the notifications correctly before the weekend starts.

dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
kylape: > I guess this is what you get when you have merge yaml, Helm and PagerDuty templating
Exactly :)

Honestly I thought the quotes inside the quotes wouldn't work, but it worked when I tested it.  Let me try to explain this line in plain english.

I want the severity to be based on the severity label coming from the alert itself.  If there is no severity label common to the group of alerts, then default to `info`.  That looks like `or .GroupLabels.severity "info"` in Go templating.  To properly escape for Helm templating, I needed to have the Helm templating engine output the literal string `{{`, since Alertmanager templating syntax is the same as Helm.  To do that, I use the expression `"{{"` inside the double bracket syntax for evaluating Go template expressions.  Thus `{{ "{{" }}`.

stehessel: Alright thanks for the explanation. Can you please put this explanation as a comment above this line?

stehessel:
I guess it'll only be rolled out for stage anyway, as we would need a new release to rollout the prod changes.

PR #1040, opened 2023-05-19
Prefactor: Make steps more explicit in fleetshard reconciler loop

rhybrillou:
## Description

Minor refactor in fleetshard: extract functions to reduce the size of the central reconciler function.

## Checklist (Definition of Done)

~~- [ ] Unit and integration tests added~~
- [ ] Added test description under `Test manual`
~~- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~~
- [ ] CI and all relevant tests are passing
~~- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`~~
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.
- [ ] Add secret to app-interface Vault or Secrets Manager if necessary

## Test manual

**TODO:** Add manual testing efforts

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup OCM_OFFLINE_TOKEN=<ocm-offline-token> OCM_ENV=development
make verify lint binary test test/integration
```

PR #1041, opened 2023-05-19
ROX-16734: Reconcile Central Audit log notifier configuration

rhybrillou:
## Description

The point is to auto-generate a notifier for Central to be able to push audit logs to an external target.

## Test manual

```
# To run tests locally run:
```

Manual steps to test:
```
CLUSTER_TYPE=colima make deploy/bootstrap
CLUSTER_TYPE=colima make deploy/dev

# extract STATIC_TOKEN as per procedure in https://github.com/stackrox/acs-fleet-manager/blob/main/docs/development/setup-developer-osd-cluster.md step 16

# instantiate central
curl --location --request POST "http://localhost:8000/api/rhacs/v1/centrals?async=true" --header "Content-Type: application/json" --header "Accept: application/json" --header "Authorization: Bearer ${STATIC_TOKEN}" --data-raw "{\"name\":\"test-on-cluster\",\"cloud_provider\":\"standalone\",\"multi_az\":true}" | jq .

# check that central deployment has ROX_DECLARATIVE_CONFIG activated
# look into the container env for the variable to be "true"
oc describe -n rhacs-... deploy/central

# check that the sensible-declarative-configs secret was created
# and that the content for key com.redhat.rhacs.auditLogNotifier is as expected

# check in the DB that there is an entry in the notifiers table for Audit Logs
```


rhybrillou:
* main
  * **PR #1040** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1040" target="_blank">Graphite</a> 
    * **PR #1041** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1041" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1041?utm_source=stack-comment).

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
rhybrillou: IDE autocompletion. Other files in the repo import v2.
I'll upgrade to v3.

SimonBaeumer (changes requested):
Can you add automated tests please?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -236,6 +250,9 @@ func (r *CentralReconciler) getInstanceConfig(remoteCentral *private.ManagedCent
 	envVars := getProxyEnvVars(remoteCentralNamespace)
 
 	scannerComponentEnabled := v1alpha1.ScannerComponentEnabled
+	// Activate Declarative configuration feature
+	envVars = ensureDeclarativeConfigurationEnabled(envVars)
```
dhaus67, line 254: Since we are now using 4.1, you can skip explicitly setting the environment variable since it will be true by default.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -383,6 +407,69 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+		var isAuditLogNotifierContentCorrect = true
+
+		existingConfigData, foundAuditLogNotifier := secret.Data[auditLogNotifierKey]
+		if foundAuditLogNotifier {
+			configurations, err := declarativeconfig.ConfigurationFromRawBytes(existingConfigData)
+			if err != nil {
+				return fmt.Errorf("getting declarative configuration data for audit log notifier: %w", err)
+			}
+			foundAuditLogNotifier = false
+			for _, configObj := range configurations {
```
dhaus67, line 444: nit: theoretically we could skip this part of iterating through the configurations given that we only store a single configuration here, but we can leave it as well.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -383,6 +407,69 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+			return nil
+		}
+		if !foundAuditLogNotifier || !isAuditLogNotifierContentCorrect {
+			encodedNotifierCfg, err := yaml.Marshal(auditLogNotifierConfig)
+			if err != nil {
+				return fmt.Errorf("encoding audit log notifier configuration: %w", err)
+			}
+			if secret.Data == nil {
+				secret.Data = make(map[string][]byte, 0)
+			}
```
dhaus67, line 466: Personally would handle this in the beginning of the function before accessing the `secret.Data` to ensure that we don't mistakenly forget it in some cases.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -383,6 +407,69 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
 	return nil
 }
 
+func (r *CentralReconciler) reconcileCentralAuditLogNotifier(ctx context.Context, remoteCentral *private.ManagedCentral) error {
```
dhaus67, line 410: Right now this function creates the secret and that's essentially it.

Since declarative config only uses config map and no CRs where a status could be populated, it's quite hard to know whether the config was successfully applied or not.
In the case of auth providers it's somewhat easier since the auth providers are exposed on an unauthenticated route on the Central's API and one can use that to verify the auth provider has been created successfully.

Do we have any ideas in mind on how to alert on this in case the creation fails, i.e. providing the wrong configuration? I'd be more comfortable if there was a part within the E2E tests that would verify that the notifier is created, if that's even possible with the current setup.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -566,6 +653,71 @@ func (r *CentralReconciler) collectReconciliationStatus(ctx context.Context, rem
...
+		delete(secret.Data, auditLogNotifierKey)
+		if len(secret.StringData) > 0 {
+			delete(secret.StringData, auditLogNotifierKey)
+		}
+
+		canDeleteSecret = len(secret.Data) <= 0 && len(secret.StringData) <= 0
+	}
+	if canDeleteSecret {
+		return r.client.Delete(ctx, secret)
+	}
```
dhaus67, line 683: Given that this will be invoked when the reconciliation is in "deletion" phase, why bother explicitly checking the secret's data and only conditionally deleting the secret?

Since this secret is solely used by the reconciler, we might as well just generically delete the complete secret here, irrespective of the specific configuration type of declarative configuration, i.e. that this both applies for auth provider as well as notifiers (and whatever else we might introduce in the future).

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -1164,6 +1290,60 @@ func (r *CentralReconciler) needsReconcile(changed bool, forceReconcile string)
 
 var resourcesChart = charts.MustGetChart("tenant-resources", nil)
 
+func (r *CentralReconciler) ensureSecretExists(
+	ctx context.Context,
+	remoteCentralNamespace string,
+	nameForErrors string,
+	actualName string,
```
dhaus67, line 1297: Those parameter namings were a tad confusing to me.

`nameForErrors` in my opinion is not required at all, the pattern we have within the reconciler to log errors for K8S objects is to specify the object's name and namespace as such:
```go
fmt.Errorf("updating secret %s/%s: %w", actualName, remoteCentralNamespace, err)
```

I'd just drop the `nameForErrors` in favor of logging the explicit name + namespace combination.

dhaus67 (commented):
Would definitely be good to enhance the existing unit tests for the reconciler and think about whether it's possible to enhance the E2E tests to include verifying the creation of the notifier within the Central.

fleetshard/pkg/central/reconciler/reconciler.go
rhybrillou: Tests are indeed required. I'll see how I can test that.

fleetshard/pkg/central/reconciler/reconciler.go
rhybrillou: I did not want to make assumptions on what was in the secret or not. I think this leaves the door open for later changes without having to go back to adjust that part because the initial assumption is not valid anymore.

fleetshard/pkg/central/reconciler/reconciler.go
rhybrillou: Here again, I avoided making assumptions. At the end of the flow, the secret should be removed. Nevertheless, if a later step fails, the secret is scrubbed (and possibly removed).

dhaus67: We can make the changes either within this PR or within the PR that introduces the declarative config for authn+z, but it's simpler in those cases to just remove the complete secret. I don't think there's any real use-case where you'd want to keep either one of those entries within the secret.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+			return
+		}
+		annotationVal, annotationFound := secret.ObjectMeta.Annotations[managedServicesAnnotation]
+		if !annotationFound {
+			return
+		}
+		assert.Equal(t, annotationVal, "true")
+	}
+	assert.Equal(t, len(expectedSecret.Data), len(secret.Data))
+	for dataKey, expectedVal := range expectedSecret.Data {
```
dhaus67, line 1354: I _think_ (but you probably have to double check) you should be able to just do `assert.Equal` here.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+}
+
+func noProxyEnvPayloadContainsException(envValue string, targetHost string, targetPort int) bool {
+	targetString := fmt.Sprintf("%s:%d", targetHost, targetPort)
+	for _, elem := range strings.Split(envValue, ",") {
+		if elem == targetString {
+			return true
+		}
+	}
+	return false
```
dhaus67, line 1458: Couldn't this be:
```suggestion
	targetString := fmt.Sprintf("%s:%d", targetHost, targetPort)
	assert.Contains(t, strings.Split(envValue, ","), targetString)
```

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+}
+
+func TestGetInstanceConfigSetsNoProxyEnvVarsForAuditLog(t *testing.T) {
+	for _, auditLogConfig := range []config.AuditLogging{defaultAuditLogConfig, vectorAuditLogConfig} {
```
dhaus67, line 1462: Probably a table test with separate `t.Run` and then using `require` where applicable to skip returns makes this easier to extend and more readable.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+		defaultAuditLogConfig,
+	)
+	declarativeConfigEnvVarSet := false
+	declarativeConfigEnable := false
+	centralConfig, err := r.getInstanceConfig(&simpleManagedCentral)
+	assert.NoError(t, err)
+	assert.NotNil(t, centralConfig)
+	if centralConfig == nil {
+		return
+	}
```
dhaus67, line 1512: Use `require` here.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+	assert.NotNil(t, centralConfig.Spec.Customize)
+	if centralConfig.Spec.Customize == nil {
+		return
+	}
```
dhaus67, line 1516: Use `require` here.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+		noDBClient,
+		defaultReconcilerOptions,
+		defaultAuditLogConfig,
+	)
+	centralConfig, err := r.getInstanceConfig(&simpleManagedCentral)
+	assert.NoError(t, err)
+	assert.NotNil(t, centralConfig)
+	if centralConfig == nil {
+		return
+	}
```
dhaus67, line 1540: Use `require` here.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+	assert.NotNil(t, centralConfig.Spec.Central)
+	if centralConfig.Spec.Central == nil {
+		return
+	}
```
dhaus67, line 1544: Use `require` here.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+	assert.NotNil(t, centralConfig.Spec.Central.DeclarativeConfiguration)
+	centralCRDeclarativeConfig := centralConfig.Spec.Central.DeclarativeConfiguration
+	if centralCRDeclarativeConfig == nil {
+		return
+	}
```
dhaus67, line 1549: Use `require` here.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+	assert.NotZero(t, len(centralCRDeclarativeConfig.Secrets))
+	sensibleDeclarativeConfigSecretFound := false                  // pragma: allowlist secret
+	for _, secretRef := range centralCRDeclarativeConfig.Secrets { // pragma: allowlist secret
+		if secretRef.Name == sensibleDeclarativeConfigSecretName { // pragma: allowlist secret
+			sensibleDeclarativeConfigSecretFound = true // pragma: allowlist secret
+		}
+	}
```
dhaus67, line 1556: Use `assert.Contains(t, centralCRDeclarativeConfig.Secret, sensibleDeclarativeConfigSecretName)`.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+		vectorAuditLogConfig.AuditLogTargetPort,
+		centralNamespace,
+	)
+
+	const otherItemKey = "other.item.key"
+
+	testCases := []struct {
+		name                    string
+		auditLogConfig          config.AuditLogging
+		preexistingSecret       bool
```
dhaus67, line 1763:
```suggestion
		preExistingSecret       bool
```

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+			},
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			ctx := context.TODO()
+			fakeClient, _, r := getClientTrackerAndReconciler(t, simpleManagedCentral, noDBClient, defaultReconcilerOptions, testCase.auditLogConfig)
+			if testCase.preexistingSecret {
+				secret := populateNotifierSecret(t, centralNamespace, testCase.notifierConfigs)
+				assert.NoError(t, fakeClient.Create(ctx, secret))
```
dhaus67, line 1868: Also probably a `require` is better in this case, since this is somewhat a test setup.

dhaus67 (changes requested):

fleetshard/pkg/central/reconciler/proxy.go
```diff
@@ -34,6 +35,10 @@ func getProxyEnvVars(namespace string) []corev1.EnvVar {
 			)
 		}
 	}
+	for _, noProxyURL := range additionalNoProxyTargets {
+		noProxyEndPoint := fmt.Sprintf("%s:%s", noProxyURL.Hostname(), noProxyURL.Port())
```
dhaus67, line 39: super nit: alternatively could just use `noProxyURL.Host` which will get you `host:port`.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -57,13 +60,21 @@ const (
 	operatorVersionKey        = "stackrox.io/operator-version"
 	defaultOperatorVersion    = "rhacs-operator.v3.74.0"
 
+	declarativeConfigurationFeatureFlagName = "ROX_DECLARATIVE_CONFIGURATION"
```
dhaus67, line 63: Once we remove the env variable check we can remove this as well.

fleetshard/pkg/central/reconciler/reconciler.go
dhaus67: I think we can now just remove it, with the latest release we are now at 4.1.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"gopkg.in/yaml.v3"
```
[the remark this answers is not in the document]
rhybrillou: https://issues.redhat.com/browse/ROX-18693

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -383,6 +413,90 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+	auditLogTargetHost string,
+	auditLogTargetPort int,
+	namespace string,
+) *declarativeconfig.Notifier {
+	auditLogEndpoint := fmt.Sprintf("https://%s:%d", auditLogTargetHost, auditLogTargetPort)
+	return &declarativeconfig.Notifier{
+		Name: auditLogNotifierName,
+		GenericConfig: &declarativeconfig.GenericConfig{
+			Endpoint:            auditLogEndpoint,
+			SkipTLSVerify:       true,
```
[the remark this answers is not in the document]
rhybrillou: During the last e2e tests, it was not ready.
Made this parameter configurable through environment variables.

fleetshard/pkg/central/reconciler/reconciler_test.go
```diff
@@ -1123,3 +1280,602 @@ func Test_stringMapNeedsUpdating(t *testing.T) {
...
+
+func TestEnsureSecretExists(t *testing.T) {
+	fakeClient, _, r := getClientTrackerAndReconciler(
+		t,
+		defaultCentralConfig,
+		noDBClient,
+		defaultReconcilerOptions,
+		defaultAuditLogConfig,
+	)
+	secretModifyFunc := func(secret *v1.Secret) (bool, error) {
```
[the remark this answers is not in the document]
rhybrillou: As far as I can tell, the "core" logic is tested in `TestGetAuditLogNotifierConfig`, `TestShouldUpdateAuditLogNotifierSecret` and `TestReconcileCentralAuditLogNotifier`

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -385,6 +416,90 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+				{
+					Key:   auditLogTenantIDKey,
+					Value: namespace,
+				},
+			},
+		},
+	}
+}
+
+func shouldUpdateAuditLogNotifierSecret(
```
SimonBaeumer, line 439: Is this logic worth the overhead of code?
I am questioning this because the reconciler is only executed when fleetshard-sync restarts or the Central changed, thus I would argue that the Secret object can just be optimistically updated when it already exists unconditionally.
This reduces testing efforts and makes the code easier to read.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -233,10 +248,19 @@ func (r *CentralReconciler) getInstanceConfig(remoteCentral *private.ManagedCent
 		return nil, errors.Wrap(err, "converting Scanner DB resources")
 	}
 
-	// Set proxy configuration
-	envVars := getProxyEnvVars(remoteCentralNamespace)
+	var envVars []corev1.EnvVar
+	if r.auditLogging.Enabled {
+		// Set proxy configuration
+		additionalNoProxyURL := url.URL{
+			Host: r.auditLogging.Endpoint(),
```
SimonBaeumer, line 255: The no proxy URL can be added unconditionally. There no risk in adding it to the no-proxy, when the host does not exist in the namespace Central just can't reach it.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -385,6 +416,90 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+				r.auditLogging,
+				remoteCentral.Metadata.Namespace,
+			)
+			if secret.Data == nil {
+				secret.Data = make(map[string][]byte)
+			}
+			shouldUpdate, err := shouldUpdateAuditLogNotifierSecret(secret, auditLogNotifierConfig)
+			if err != nil || !shouldUpdate {
+				return false, err
+			}
```
SimonBaeumer, line 492: This check be removed if done unconditionally.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -385,6 +416,90 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+	return r.ensureSecretExists(
+		ctx,
+		remoteCentral.Metadata.Namespace,
+		sensibleDeclarativeConfigSecretName,
+		func(secret *corev1.Secret) (bool, error) {
+			auditLogNotifierConfig := getAuditLogNotifierConfig(
+				r.auditLogging,
+				remoteCentral.Metadata.Namespace,
+			)
+			if secret.Data == nil {
```
SimonBaeumer, line 486: When called unconditionally the notifier config can always overwrite this field and initialize the field while updating:

```
secret.Data = map[string][]byte{
  auditLogNotifierKey: encodedNotifierConfig,
}
```

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -1166,6 +1277,55 @@ func (r *CentralReconciler) needsReconcile(changed bool, forceReconcile string)
...
+		return fmt.Errorf("getting %s/%s secret: %w", namespace, actualName, err)
+	}
+	if err == nil {
+		changed, modificationErr := secretModifyFunc(secret)
+		if modificationErr != nil {
+			return fmt.Errorf("updating %s/%s secret content: %w", namespace, actualName, modificationErr)
+		}
+		if !changed {
+			return nil
+		}
```
SimonBaeumer, line 1300: Is there a reason to update the secret conditionally?

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
rhybrillou: There are more parts than just the Audit Log Notifier that will be configured through declarative configuration and the associated secret (see PR #1156 for the auth provider part).
The point to consider here is whether we'd like to open the declarative configuration part for cloud service users. If so, I think focusing on the part of the `declarative configuration secret` and not overwriting the user-provided configuration part of the secret would be the way to go.

fleetshard/pkg/central/reconciler/reconciler.go
rhybrillou: Do we want to re-push the secret on every fleetshard-sync run even if it did not change ?

SimonBaeumer: I expect the update calls to be in a range of 5-10 in 4 weeks. So, in this case it is okay to update it on every reconcile. Additionally, in an operator it is considered best-pracitce to have an idempotent reconciler which does not cache states between reconcile runs.
We cache the Central changes to "mimikry" an event driven system in a polling-based mechanism.

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: > the target endpoint configuration can be expected to not be valid.

There is a default configured. How could this be not there? Could it be a simple unit test which verifies that the host is added unconditionally?
Reducing N-path complexity improves code readability, maintenance and testing.

fleetshard/pkg/central/reconciler/reconciler.go
rhybrillou: I think it's better not to write back the secret when its content has not changed.
Otherwise, the secrets would be potentially re-written on each fleetshard-sync reconciliation run.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -385,6 +416,65 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
 }
 
+func getAuditLogNotifierConfig(
+	auditLoggingConfig config.AuditLogging,
+	namespace string,
+) *declarativeconfig.Notifier {
+	return &declarativeconfig.Notifier{
+		Name: auditLogNotifierName,
+		GenericConfig: &declarativeconfig.GenericConfig{
+			Endpoint:            fmt.Sprintf("https://%s", auditLoggingConfig.Endpoint()),
```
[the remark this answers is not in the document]
rhybrillou: it's used in one location without, and here with. I can change that, though.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -385,6 +411,65 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+		return nil
+	}
+	namespace := remoteCentral.Metadata.Namespace
+	exists, err := r.checkSecretExists(ctx, namespace, sensibleDeclarativeConfigSecretName)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return nil
+	}
```
dhaus67, line 461: super nit: can be simplified to
```suggestion
	if err != nil || exists {
		return err
	}
```

dhaus67 (approved):

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -385,6 +411,65 @@ func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remote
...
+	return r.ensureSecretExists(
+		ctx,
+		namespace,
+		sensibleDeclarativeConfigSecretName,
+		func(secret *corev1.Secret) error {
+			r.configureAuditLogNotifier(secret, namespace)
```
SimonBaeumer, line 467: Should the error be returned here?

SimonBaeumer (approved):

PR #1074, opened 2023-06-06
Introduce acsfleetctl binary

johannes94:
## Description

Move the Fleet-Manager CLI to it's own binary instead of embedding it in the fleet-manager binary. Some motivating facts for this PR:

- Usage info for the CLI are massive, because of all the flags added to the root fleet-manager command:
- Simple errors in CLI usage lead to panic logs with massive stack trace, crashing is fine for a service like fleet-manager serve because we don't want it to start with wrong flags/configuration. But for the CLI this is unnecessary
- Usage of glog.Fatal() on multipel occasion leads to unmanaged and unrecoverable panic output of stacktraces and goroutines

Slack thread: https://redhat-internal.slack.com/archives/C02HTD51SMD/p1685620429880799

- ~~[ ] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)~~
- ~~[ ] Check AWS limits are reasonable for changes provisioning new resources~~

## Test manual

Run local fleet-manager and test CLI commands against it:

```
make binary

# local crc cluster started and kubectl configured
./fleet-manager serve

export OCM_TOKEN=$(ocm token --refresh)
rhoas login --auth-url=https://auth.redhat.com/auth/realms/EmployeeIDP

./acsfleetctl centrals create --name "test-central-1" --region "standalone" --provider "standalone"
# Instance got created

./acsfleetctl centrals list
# Prints CentralRequestList with the instance created

central_id=$(./acsfleetctl centrals list | jq '.items[0].id' -r )
./acsfleetctl centrals get --id $central_id
# Print the central request as JSON

export RHOAS_TOKEN=$(rhoas authtoken)
./acsfleetctl admin centrals list
# Print CentralList (Admin API DTO) with the created central

./acsfleetctl centrals delete --id $central_id
# Prints {status_code: 202}
```

```
# To run tests locally run:
```


internal/dinosaur/pkg/cmd/cliflags/cliflags.go
```diff
@@ -0,0 +1,51 @@
+// Package cliflags defines util methods for flags used in acsfleetctl
+package cliflags
```
SimonBaeumer, line 2: There is a package: `pkg/flags/flags.go`

SimonBaeumer (commented):
LGTM, looks simpler than expected. I am convinced 👍

internal/dinosaur/pkg/cmd/cliflags/cliflags.go
johannes94: I didn't use it on purpose, see the changes in comments I did for `pkg/flags/flags.go`. The problem is the package glog.Fatals a lot if flags are not set. With glog.Fatal there is a problem:

It is unrecoverable because it calls ox.Exit(2) directly after logging and not panic (that's why I changed the comment), this does not allow us to intercept the exit on top level with the recover function and modify output to the output we desired for the CLI. I didn't want to change the fleet-manager serve/migrate behaviour that's why I created another cliflags package just for CLI.

SimonBaeumer: In this case I think we should return an error instead of `os.Exit` or `panic`.
What is your concern about `serve` and `migrate` behaviour? 
I see these changes as uncritical because it is parsing CLI flags which does not crash in prod (except we messed something up pretty bad).

johannes94: > What is your concern about serve and migrate behaviour?
No concern that can not be addressed, just don't wanted to change anything in that regards in this specific PR.

SimonBaeumer: I still see disadvantages for the pruposed solution and not convinced by it. The panic in the library shows its architectural weakness in two disadvantages here:

- Source code is duplicated for a panic when it could return an error instead and be supported by both libraries. It shows a leaky abstraction because the same problem is solved twice to exchange its output format (here panic & glog.Fatal). However, there could be wrapper functions to panic or glog.Fatal and the caller decides what to use.
- Deferrable recovery is more complex to read compared to an `if err != nil` because an engineer 
  1. most look into the deferred function
  2. keep a map in mind that there is a deferred function executed

johannes94: We had short call to discuss this, Outcome:
- Duplication of flags package removed `glog.Fatal` replaced by `panic` calls
- CLI implementation itself stays with calling "MustGet*" function from the flags package for required flags.

SimonBaeumer (commented):
I see similar messages twice now locally:

```
❯ ./acsfleetctl centrals get --debug
Error: required flag(s) "id" not set
Usage:
  acsfleetctl centrals get [flags]

Flags:
  -h, --help        help for get
      --id string   Central ID (required)

Global Flags:
      --debug   use debug output

Error executing command: required flag(s) "id" not set%
```

And could not see any changes when I removed the panic recovery and no difference with the `--debug` flag enabled.

```
❯ ./acsfleetctl centrals --debug get
Error: required flag(s) "id" not set
Usage:
  acsfleetctl centrals get [flags]

Flags:
  -h, --help        help for get
      --id string   Central ID (required)

Global Flags:
      --debug   use debug output

Error executing command: required flag(s) "id" not set%
❯ ./acsfleetctl centrals --debug delete
```

@johannes94 Can you confirm this behaviour?

johannes94:
I can confirm, the behaviour above. I removed the second log from the code @SimonBaeumer

SimonBaeumer (approved):

PR #1078, opened 2023-06-06
ROX-13709: Introduce the external secrets operator on the Data Plane

kovayur:

## Test manual

```
# To run tests locally run:
```


stehessel (approved):
Very nice! Thanks for picking up ROX-13709 in its unrefined state 🚀 . We can do a follow up for other service secrets (like probe and cloudwatch exporter).

What do you think about enabling etcd encryption? That way the secrets are encrypted at rest in the cluster as well as in AWS - see https://docs.openshift.com/container-platform/4.13/security/encrypting-etcd.html on how to do that.

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync-external-secrets.yaml
```diff
@@ -0,0 +1,43 @@
...
+            namespace: external-secrets
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: fleetshard-sync-ext-secret
+  namespace: {{ .Release.Namespace }}
+spec:
+  secretStoreRef:
+    name: secrets-manager-secret-store
```
stehessel, line 25: Could you make the store ref parameterizable such that one can more easily swap it out for a fake store when testing?

deploy/helm/external-secrets/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [[ "${ENVIRONMENT}" != "dev" ]]; then
+    init_chamber
+    load_external_config "cluster-${CLUSTER_NAME}" CLUSTER_
+    oc login --token="${CLUSTER_ROBOT_OC_TOKEN}" --server="$CLUSTER_URL"
+fi
+
+helm repo add external-secrets https://charts.external-secrets.io
+
+helm upgrade -i external-secrets \
```
SimonBaeumer, line 28: Can the external secret operator helm chart be added as a subchart to our Helm chart, like i.e. observability?

I am worried that we eventually have multiple deployment scripts for different components which makes it difficult to replicate environments.

SimonBaeumer (changes requested):
I like that secrets are updated automatically via the external-secrets operator.
I have some follow up questions on this:

1. How are secrets in local development managed after this change? I am afraid that our local environment diverges a lot from staging/prod env.
2. Is a migration from Pod env vars to Secrets following in a separate PR?
3. How is a new Secret added via the external-secret operator? (Link to documentation is sufficient for me)

deploy/helm/external-secrets/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+    echo "Usage: $0 [environment] [cluster]" >&2
+    echo "Known environments: dev stage prod"
+    echo "Cluster typically looks like: acs-{environment}-dp-01"
+    exit 2
+fi
+
+ENVIRONMENT=$1
+CLUSTER_NAME=$2
+
+if [[ "${ENVIRONMENT}" != "dev" ]]; then
```
porridge, line 20: Since you already have this part gated in the script, can you maybe change the `upgrade` below to "`upgrade` or `lint` depending on environment" and remove the `if` in the action that calls this script, so we can at least partially validate this script and chart in PRs?

porridge (approved):
Sounds like this operator supports other cloud providers too, so this seems like a wise move!

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync-external-secrets.yaml
kovayur: Done

.github/workflows/deploy-data-plane.yaml
```diff
@@ -40,13 +40,54 @@ env:
...
   terraform:
     name: Re-terraform ${{ inputs.acs_environment }} clusters
     runs-on: ubuntu-latest
     permissions:
       id-token: write
       contents: read
     environment: ${{ inputs.github_environment }}
+    # We need to deploy the external-secrets operator BEFORE the actual terraforming to create the necessary CRDs on the clusters
+    needs: deploy-external-secrets
+    # Run this job if deploy-external-secrets is skipped
```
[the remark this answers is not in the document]
kovayur: Obsolete, because the external-secret chart is now a dependency of rhacs-terraform

deploy/helm/external-secrets/deploy.sh
kovayur: I already did this in `install_addon.sh`, although there's no `lint` yet 🥲

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -179,3 +167,35 @@ secured-cluster:
...
       key: ""
+
+external-secrets:
+  fullnameOverride: rhacs-external-secrets
+  installCRDs: false
+  image:
+    repository: quay.io/app-sre/external-secrets
+    tag: v0.9.5
+  securityContext:
+    runAsUser: 1001130000
```
ebensh, line 178: Sorry for my ignorance here, where do these user IDs come from? I know the default of 1000 but what's the source for these?

ebensh (approved):

dp-terraform/helm/rhacs-terraform/values.yaml
kovayur: They came from openshift, the default is 1000, but this is outside the allowed range of values. I took the first values from the range

SimonBaeumer (approved):

kovayur:
> What do you think about enabling etcd encryption? That way the secrets are encrypted at rest in the cluster as well as in AWS - see https://docs.openshift.com/container-platform/4.13/security/encrypting-etcd.html on how to do that.

@stehessel raised a follow-up ticket for this: ROX-20619

PR #1083, opened 2023-06-12
ROX-17712: wait for central to be deleted

ludydoo:
## Description
https://issues.redhat.com/browse/ROX-17712
Adding retries to wait for Central to be deleted, not just marked for deletion.

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"sync/atomic"
 
 	"github.com/golang/glog"
 	openshiftRouteV1 "github.com/openshift/api/route/v1"
@@ -27,6 +26,7 @@ import (
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/pointer"
 	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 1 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -741,26 +744,45 @@ func (r *CentralReconciler) getDBPasswordFromSecret(ctx context.Context, central
...
+			if apiErrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, errors.Wrapf(err, "failed to get central CR %v", centralKey)
 		}
 
-		return false, errors.Wrapf(err, "delete central CR %s/%s", central.GetNamespace(), central.GetName())
-	}
+		// avoid being stuck in a deprovisioning state due to the pause reconcile annotation
+		if err := r.disablePauseReconcileIfPresent(ctx, central); err != nil {
```
ebensh, line 763: This gives me an odd feeling - I would think of the annotation as a manual lock/unlock and this feels like a preemption that makes sense as an optimization (we get the central deleted faster / won't get stuck), OTOH it feels like reconciliation pause shouldn't have caveats that increase the state space ("Unless it was marked as deleted at that time in which case the annotation is removed...") Unless that's the path we want to go down.

Maybe I'm being overly cautious though - looking at the pause annotation today it shouldn't be paused *and* upgrading at the same time, because then it should be in the unpaused state. And users should be allowed to delete their instance whenever without it being paused/locked for days while we complete a version upgrade.

ebensh (approved):

fleetshard/pkg/central/reconciler/reconciler.go
vladbologa: This line isn't new, it was added with this PR: https://github.com/stackrox/acs-fleet-manager/pull/905

My main motivation for adding it was the following:
1. a customer's instance runs into some problems, so we make some manual changes, and then add the `pause-reconcile` annotation so that the operator doesn't overwrite these changes
2. the customer deletes their instance
3. instance remains forever stuck in "deprovisioning"

Taking this approach was discussed a few times on slack, if you're interested I could try to find where.

PR #1102, opened 2023-06-15
ROX-17485: Add delete chart function

kurlov:

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/central/charts/charts.go
+++ b/fleetshard/pkg/central/charts/charts.go
@@ -18,6 +18,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/chartutil"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


fleetshard/pkg/central/charts/charts.go
```diff
@@ -149,3 +150,35 @@ func InstallOrUpdateChart(ctx context.Context, obj *unstructured.Unstructured, c
 	}
 	return nil
 }
+
+// DeleteChartResources delete chart resources from cluster
+func DeleteChartResources(ctx context.Context, client ctrlClient.Client, releaseName string, namespace string, chart *chart.Chart, chartVals chartutil.Values) (bool, error) {
+	objs, err := RenderToObjects(releaseName, namespace, chart, chartVals)
```
SimonBaeumer, line 156: How does the Helm chart know which resources are deleted?

For example, in my mind following scenario:

1. Fleetshard-Sync installs 3 operator versions
2. Fleetshard-Sync receives new list, only needs to maintain 2 operator versions

Is the obsolete operator deployment deleted in such a case?
In the current scenario it looks like the whole chart is deleted.

fleetshard/pkg/central/charts/charts.go
```diff
@@ -149,3 +150,35 @@ func InstallOrUpdateChart(ctx context.Context, obj *unstructured.Unstructured, c
 	}
 	return nil
 }
+
+// DeleteChartResources delete chart resources from cluster
+func DeleteChartResources(ctx context.Context, client ctrlClient.Client, releaseName string, namespace string, chart *chart.Chart, chartVals chartutil.Values) (bool, error) {
```
johannes94, line 155: How do we ensure that a resource removed from a chart throughout different chart version are properly deleted. Suppose we have following scenarion:
1. We have a chart and apply it
2. We remove a single resource from the chart at some point
3. We delete the chart

I think with this approach the resource that was in the old version of the chart would not be deleted. Because RenderToObjects would only create the objects present in the new version of the chart.

PR #1103, opened 2023-06-16
ROX-17718: add operator labeling

SimonBaeumer:
## Description

- Add Operator Label selector by setting `CENTRAL_LABEL_SELECTOR` in the deployment. Use hardcoded value and later it will be replaced with GitRef from GitOps configmap.
- Add Central CR label `rhacs.redhat.com/version-selector` for specifying selecting. Temporary operator image tag is used. It will be replaced with GitRef from GitOps configmap.
- Add new struct `DeploymentConfig` which will contain data comes from GitOps configmap necessary for operator deployment
- Replace repository + tag with image in operator deployment
- Added `github.com/containers/image/docker` image validation
- Added `apimachinery` label validation
- Operator upgrade can use images with digest 🎉 

## Test manual

```
# Reset your k8s local cluster
# below commad is for colima
colima k reset

# Bootstrap and deploy using canary upgrade script
scripts/install_operator_canary.sh

# Make sure Operator pod is ready
k get pods -n rhacs
NAME                                    READY   STATUS    RESTARTS       AGE
rhacs-operator-4.0.1-669ff97584-nnnpg   2/2     Running   2 (3h5m ago)   4h40m

# Check CENTRAL_LABEL_SELECTOR is set correctly
k describe pod rhacs-operator-4.0.1-669ff97584-nnnpg manager -n rhacs
...
    Environment:
      ENABLE_WEBHOOKS:                  false
      RELATED_IMAGE_MAIN:
      RELATED_IMAGE_SCANNER:
      RELATED_IMAGE_SCANNER_SLIM:
      RELATED_IMAGE_SCANNER_DB:
      RELATED_IMAGE_SCANNER_DB_SLIM:
      RELATED_IMAGE_COLLECTOR_SLIM:
      RELATED_IMAGE_COLLECTOR_FULL:
      RELATED_IMAGE_ROXCTL:
      RELATED_IMAGE_CENTRAL_DB:
      ROX_OPERATOR_MAIN_REGISTRY:       quay.io/rhacs-eng
      ROX_OPERATOR_COLLECTOR_REGISTRY:  quay.io/rhacs-eng
      MEMORY_LIMIT_BYTES:               1073741824 (limits.memory)
      CENTRAL_LABEL_SELECTOR:           rhacs.redhat.com/version-selector=4.0.1
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-c5p4h (ro)
...


# Create an instance
curl -XPOST --header "Authorization: Bearer $(ocm token)" 'http://localhost:8000/api/rhacs/v1/centrals?async=true' --data '{"region": "standalone", "cloud_provider": "standalone", "multi_az": true, "name": "testsel"}' | jq
{
  "id": "civju48go10001p9nd10",
  "kind": "CentralRequest",
  "href": "/api/rhacs/v1/centrals/civju48go10001p9nd10",
  "status": "accepted",
  "cloud_provider": "standalone",
  "multi_az": true,
  "region": "standalone",
  "owner": "rh-ee-akurlov",
  "name": "testsel",
  "created_at": "2023-07-25T03:18:41.369854009Z",
  "updated_at": "2023-07-25T03:18:41.369854009Z",
  "instance_type": "eval"
}

# Wait until it's ready
curl --header "Authorization: Bearer $(ocm token)" http://localhost:8000/api/rhacs/v1/centrals | jq
{
  "kind": "CentralRequestList",
  "page": 1,
  "size": 1,
  "total": 1,
  "items": [
    {
      "id": "civju48go10001p9nd10",
      "kind": "CentralRequest",
      "href": "/api/rhacs/v1/centrals/civju48go10001p9nd10",
      "status": "ready",
      "cloud_provider": "standalone",
      "multi_az": true,
      "region": "standalone",
      "owner": "rh-ee-akurlov",
      "name": "testsel",
      "centralUIURL": "https://acs-civju48go10001p9nd10.cluster.local",
      "centralDataURL": "acs-data-civju48go10001p9nd10.cluster.local:443",
      "created_at": "2023-07-25T03:18:41.369854Z",
      "updated_at": "2023-07-25T03:22:10.891232Z",
      "instance_type": "eval"
    }
  ]
}

# Check Central CR labels
k describe central testsel
Name:         testsel
Namespace:    rhacs-civju48go10001p9nd10
Labels:       app.kubernetes.io/managed-by=rhacs-fleetshard
              rhacs.redhat.com/instance-type=eval
              rhacs.redhat.com/org-id=11009103
              rhacs.redhat.com/tenant=civju48go10001p9nd10
              rhacs.redhat.com/version-selector=4.0.0
Annotations:  platform.stackrox.io/managed-services: true
              platform.stackrox.io/obsolete-central-pvc: true
              rhacs.redhat.com/org-name: stubbed-name
              rhacs.redhat.com/revision: 1
API Version:  platform.stackrox.io/v1alpha1
Kind:         Central
....

# Check operator logs
k logs -f rhacs-operator-4.0.1-669ff97584-nnnpg manager -n rhacs
...
2023-07-25T05:09:02Z	DEBUG	controllers.Central	performing update for testsel
2023-07-25T05:09:02Z	DEBUG	controllers.Central	dry run for testsel
2023-07-25T05:09:03Z	INFO	controllers.Central	Release reconciled	{"central": "rhacs-civju48go10001p9nd10/testsel", "name": "testsel", "version": 1}
2023-07-25T05:09:03Z	INFO	KubeAPIWarningLogger	unknown field "status.deployedRelease.manifest"
2023-07-25T05:09:03Z	INFO	KubeAPIWarningLogger	unknown field "status.deployedRelease.name"
2023-07-25T08:23:38Z	DEBUG	controllers.Central	Reconciliation triggered	{"central": "rhacs-civju48go10001p9nd10/testsel"}
2023-07-25T08:23:38Z	DEBUG	controllers.Central	preparing upgrade for testsel
2023-07-25T08:23:39Z	DEBUG	controllers.Central	performing update for testsel
2023-07-25T08:23:39Z	DEBUG	controllers.Central	dry run for testsel
2023-07-25T08:23:39Z	INFO	controllers.Central	Release reconciled	{"central": "rhacs-civju48go10001p9nd10/testsel", "name": "testsel", "version": 1}
...

# Delete the instance
curl -XDELETE --header "Authorization: Bearer $(ocm token)" 'http://localhost:8000/api/rhacs/v1/centrals/civju48go10001p9nd10?async=true' | jq

# Make sure it's deleted
curl --header "Authorization: Bearer $(ocm token)" http://localhost:8000/api/rhacs/v1/centrals | jq
{
  "kind": "CentralRequestList",
  "page": 1,
  "size": 0,
  "total": 0,
  "items": []
}
```

### DIFF
--- a/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
+++ b/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
@@ -82,7 +82,7 @@ spec:
                   containerName: manager
                   resource: limits.memory
                   divisor: '0'
-            {{- if $.Values.operator.centralLabelSelector }}
+            {{- if .labelSelector }}
             - name: CENTRAL_LABEL_SELECTOR
               value: "rhacs.redhat.com/version-selector={{ .labelSelector }}"
             {{- end }}
--- a/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
+++ b/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
@@ -86,7 +86,7 @@ spec:
             - name: CENTRAL_LABEL_SELECTOR
               value: "rhacs.redhat.com/version-selector={{ .labelSelector }}"
             {{- end }}
-          image: "{{ .repository }}:{{ .tag }}"
+          image: "{{ .image }}"
           imagePullPolicy: IfNotPresent
           livenessProbe:
             failureThreshold: 3
--- a/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
+++ b/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
@@ -82,9 +82,9 @@ spec:
                   containerName: manager
                   resource: limits.memory
                   divisor: '0'
-            {{- if $.Values.operator.centralLabelSelector -}}
+            {{- if .labelSelector }}
             - name: CENTRAL_LABEL_SELECTOR
-              value: "rhacs.redhat.com/tag={{ substr 0 64 (lower .tag) }},rhacs.redhat.com/repository={{ substr 0 64 (lower .repository )}}"
+              value: "rhacs.redhat.com/version-selector={{ .labelSelector }}"
             {{- end }}
           image: "{{ .repository }}:{{ .tag }}"
           imagePullPolicy: IfNotPresent
--- a/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
+++ b/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
@@ -82,7 +82,7 @@ spec:
                   containerName: manager
                   resource: limits.memory
                   divisor: '0'
-            {{- if $.Values.operator.centralLabelSelector -}}
+            {{- if $.Values.operator.centralLabelSelector }}
             - name: CENTRAL_LABEL_SELECTOR
               value: "rhacs.redhat.com/version-selector={{ .labelSelector }}"
             {{- end }}
--- a/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
+++ b/fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
@@ -82,7 +82,7 @@ spec:
                   containerName: manager
                   resource: limits.memory
                   divisor: '0'
-            {{- if .labelSelector }}
+            {{- if $.Values.operator.centralLabelSelector -}}
             - name: CENTRAL_LABEL_SELECTOR
               value: "rhacs.redhat.com/version-selector={{ .labelSelector }}"
             {{- end }}
--- a/fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
+++ b/fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
@@ -3,15 +3,15 @@
 # Declare variables to be passed into your templates.
 
 operator:
-  # Each item in images should contain `repository` and `tag` key
+  # Each item in images should contain `deploymentName`, `image` and `labelSelector` keys
   # example:
   #   images:
   #   - deploymentName: rhacs-operator-manager-4.0.0
-  #     repository: quay.io/rhacs-eng/stackrox-operator
-  #     tag: 4.0.0
+  #     image: quay.io/rhacs-eng/stackrox-operator:rhacs-operator-manager-4.0.0
+  #     labelSelector: 4.0.0
   #   - deploymentName: rhacs-operator-manager-4.0.1
-  #     repository: quay.io/rhacs-eng/stackrox-operator
-  #     tag: 4.0.1
+  #     image: quay.io/rhacs-eng/stackrox-operator:rhacs-operator-manager-4.0.1
+  #     labelSelector: 4.0.1
   images: []
   # TODO: add values for resource limits and requests for both proxy and manager container
   # TODO: add value for kube-rbac-proxy image
--- a/fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
+++ b/fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
@@ -7,10 +7,10 @@ operator:
   # example:
   #   images:
   #   - deploymentName: rhacs-operator-manager-4.0.0
-  #     image: quay.io/rhacs-eng/stackrox-operator:rhacs-operator-manager-4.0.0
+  #     image: quay.io/rhacs-eng/stackrox-operator:4.0.0
   #     labelSelector: 4.0.0
   #   - deploymentName: rhacs-operator-manager-4.0.1
-  #     image: quay.io/rhacs-eng/stackrox-operator:rhacs-operator-manager-4.0.1
+  #     image: quay.io/rhacs-eng/stackrox-operator:4.0.1
   #     labelSelector: 4.0.1
   images: []
   # TODO: add values for resource limits and requests for both proxy and manager container
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -39,8 +39,7 @@ func parseOperatorConfigs(operators []DeploymentConfig) ([]chartutil.Values, err
 			return nil, err
 		}
 		image := imageReference.String()
-		errs := validation.IsValidLabelValue(operator.GitRef)
-		if errs != nil {
+		if errs := validation.IsValidLabelValue(operator.GitRef); errs != nil {
 			return nil, fmt.Errorf("label selector %s is not valid: %v", operator.GitRef, errs)
 		}
 		deploymentName := generateDeploymentName(operator.GitRef)
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -26,7 +26,8 @@ const (
 	maxOperatorDeploymentNameLength = 63
 )
 
-func GetRepoAndSHA256FromImage(img string) (string, string, error) {
+// GetRepoAndTagFromImage returns the repo and image tag
+func GetRepoAndTagFromImage(img string) (string, string, error) {
 	if !strings.Contains(img, ":") {
 		return "", "", fmt.Errorf("failed to parse image %q", img)
 	}
@@ -43,7 +44,7 @@ func parseOperatorImages(images []string) ([]chartutil.Values, error) {
 	var operatorImages []chartutil.Values
 	uniqueImages := make(map[string]bool)
 	for _, img := range images {
-		repo, tag, err := GetRepoAndSHA256FromImage(img)
+		repo, tag, err := GetRepoAndTagFromImage(img)
 		if err != nil {
 			return nil, err
 		}
@@ -66,12 +67,6 @@ type ACSOperatorManager struct {
 	client         ctrlClient.Client
 	crdURL         string
 	resourcesChart *chart.Chart
-}
-
-// ACSOperatorImage operator image representation which tells when to download CRD or skip it
-type ACSOperatorImage struct {
-	Image      string
-	InstallCRD bool
 }
 
 var urlRegexExp *regexp.Regexp
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -4,7 +4,6 @@ package operator
 import (
 	"context"
 	"fmt"
-	"regexp"
 	"strings"
 
 	"github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/charts"
@@ -91,7 +90,7 @@ func parseOperatorConfigs(operators []DeploymentConfig) ([]chartutil.Values, err
 			return nil, err
 		}
 
-		deploymentName := generateDeploymentName(tag)
+		deploymentName := generateDeploymentName(operator.Version)
 		// deployment name has the same requirements as label values
 		if !IsValidLabel(deploymentName) {
 			return nil, fmt.Errorf("deployment name %s is not valid", deploymentName)
@@ -118,21 +117,6 @@ type ACSOperatorManager struct {
 	client         ctrlClient.Client
 	crdURL         string
 	resourcesChart *chart.Chart
-}
-
-var urlRegexExp *regexp.Regexp
-
-func init() {
-	var err error
-	urlRegexExp, err = regexp.Compile("/[^a-z0-9\\-_]/g")
-	if err != nil {
-		panic(fmt.Errorf("invalid URL regex, could not be compiled %w", err))
-	}
-}
-
-// GetValidSelectorTag returns a valid selector string which can be used in a kubernetes metadata label.
-func GetValidSelectorTag(tag string) string {
-	return urlRegexExp.ReplaceAllString(tag, "")
 }
 
 // InstallOrUpgrade provisions or upgrades an existing ACS Operator from helm chart template
@@ -236,8 +220,8 @@ func (u *ACSOperatorManager) RemoveUnusedOperators(ctx context.Context, desiredI
 	return nil
 }
 
-func generateDeploymentName(tag string) string {
-	return operatorDeploymentPrefix + "-" + tag
+func generateDeploymentName(version string) string {
+	return operatorDeploymentPrefix + "-" + version
 }
 
 func (u *ACSOperatorManager) generateCRDTemplateUrls(tag string) []string {
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -55,7 +55,12 @@ func parseOperatorImages(images []string) ([]chartutil.Values, error) {
 		}
 		if _, used := uniqueImages[repo+tag]; !used {
 			uniqueImages[repo+tag] = true
-			img := chartutil.Values{"deploymentName": deploymentName, "repository": repo, "tag": tag, "labelSelector": GetValidSelectorTag(tag)}
+			img := chartutil.Values{
+				"deploymentName": deploymentName,
+				"repository":     repo,
+				"tag":            tag,
+				"labelSelector":  GetValidSelectorTag(tag),
+			}
 			operatorImages = append(operatorImages, img)
 		}
 	}
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -29,9 +29,8 @@ const (
 
 // DeploymentConfig represents operator configuration for deployment
 type DeploymentConfig struct {
-	Image         string
-	LabelSelector string
-	Version       string
+	Image  string
+	GitRef string
 }
 
 // GetRepoAndTagFromImage returns the repo and image tag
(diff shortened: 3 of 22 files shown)


fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
```diff
@@ -82,9 +82,9 @@ spec:
                   containerName: manager
                   resource: limits.memory
                   divisor: '0'
-            {{- if $.Values.operator.centralLabelSelector -}}
+            {{- if .labelSelector }}
             - name: CENTRAL_LABEL_SELECTOR
-              value: "rhacs.redhat.com/tag={{ substr 0 64 (lower .tag) }},rhacs.redhat.com/repository={{ substr 0 64 (lower .repository )}}"
+              value: "rhacs.redhat.com/version-selector={{ .labelSelector }}"
```
ludydoo, line 87: I think this should be
value: "{{ .labelSelector }}"

fleetshard/pkg/runtime/runtime.go
```diff
@@ -245,18 +246,27 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
...
+		}
+		desiredOperators = append(desiredOperators, operatorConfiguration)
 		desiredOperatorImages = append(desiredOperatorImages, central.Spec.OperatorImage)
 	}
+
 	ctx := context.Background()
 
-	for _, img := range desiredOperatorImages {
-		glog.Infof("Installing Operator: %s", img)
+	for _, oper := range desiredOperators {
```
SimonBaeumer, line 263:
```suggestion
	for _, operator := range desiredOperators {
```

fleetshard/pkg/runtime/runtime.go
```diff
@@ -245,18 +246,27 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
 }
 
 func (r *Runtime) upgradeOperator(list private.ManagedCentralList) error {
+	var desiredOperators []operator.DeploymentConfig
 	var desiredOperatorImages []string
 	for _, central := range list.Items {
+		// TODO: read GitRef ManagedCentral
```
SimonBaeumer, line 252: nit: sounds to me like the `GitRef` is part of a ManagedCentral.

```suggestion
		// TODO: read GitRef ManagedCentral list call
```

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
SimonBaeumer: Important to me:
I suggest we should take a step back from the generic labelling approach. The label selectors are too [powerful](https://upload.wikimedia.org/wikipedia/en/9/9f/Over_9000%21.png) imho. 
I think the advantage in having a fixed `labelSelector` is to disable computational part of the selector logic which is difficutl to be tested before deploying to production.

Centrals could still be paused via the pause-reconcile annotation.

fleetshard/pkg/central/operator/upgrade_test.go
```diff
@@ -162,8 +172,12 @@ func TestOperatorUpgradeDoNotInstallLongTagVersion(t *testing.T) {
 	fakeClient := testutils.NewFakeClientBuilder(t).Build()
 	u := NewACSOperatorManager(fakeClient, crdURL)
 
-	operatorImageWithLongTag := "quay.io/rhacs-eng/stackrox-operator:4.0.1-with-ridiculously-long-tag-version-name"
-	err := u.InstallOrUpgrade(context.Background(), []string{operatorImageWithLongTag}, crdTag1)
+	longVersionName := "4.0.1-with-ridiculously-long-version-name-like-really-long-one"
```
SimonBaeumer, line 175: +1 for the longVersionName 🚀

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -18,32 +19,91 @@ import (
...
+type DeploymentConfig struct {
+	Image  string
+	GitRef string
+}
+
+// GetRepoAndTagFromImage returns the repo and image tag
+func GetRepoAndTagFromImage(img string) (string, string, error) {
+	if !strings.Contains(img, ":") {
+		return "", "", fmt.Errorf("failed to parse image %q", img)
+	}
```
SimonBaeumer, line 40: I think `strings.Split(img, ":")` would in this case return `0`. So the second if-case catches this error too.

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -18,32 +19,91 @@ import (
...
 
-func parseOperatorImages(images []string) ([]chartutil.Values, error) {
+// DeploymentConfig represents operator configuration for deployment
+type DeploymentConfig struct {
+	Image  string
+	GitRef string
+}
+
+// GetRepoAndTagFromImage returns the repo and image tag
+func GetRepoAndTagFromImage(img string) (string, string, error) {
```
ludydoo, line 37: I think this logic wouldn't work when using `repo@sha256:abc123`

SimonBaeumer (commented):
some nits, otherwise looks good to me

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -18,32 +19,91 @@ import (
...
+		deploymentName := generateDeploymentName(operator.GitRef)
+		if len(deploymentName) > maxLabelLength {
+			return nil, fmt.Errorf("invalid name: %s. It contains more than %d characters", deploymentName, maxLabelLength)
 		}
 		if _, used := uniqueImages[repo+tag]; !used {
 			uniqueImages[repo+tag] = true
-			img := chartutil.Values{"deploymentName": deploymentName, "repository": repo, "tag": tag}
+			img := chartutil.Values{
+				"deploymentName": deploymentName,
+				"repository":     repo,
```
ludydoo, line 103: I think we should just use `image` rather than `repository` + `image`

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
ludydoo: I don't agree. If validation is a concern, we can have a github action to validate those label selectors as part of the git-ops config pipeline using vanilla kubernetes validation. 

By using a static label rather than allowing to inject the whole selector, we're diminishing the flexibility we have in deploying the operator. All we want is to add an environment variable to the operator deployment. I don't think we should wrap it in logic that is applicable to a single usecase that we foresee right now.

SimonBaeumer: > I don't agree. If validation is a concern, we can have a github action to validate those label selectors as part of the git-ops config pipeline using vanilla kubernetes validation.

Validation is not the main concern, it is to ensure that the correct Central deployments are reconciled by the correct Operator. We cannot easily test this behaviour in production, i.e. for a misconfigured Central label selector.

> I don't think we should wrap it in logic that is applicable to a single usecase that we foresee right now.

What use-cases do you see? 
What could we not do with a less flexible approach?

ludydoo: > Validation is not the main concern
To me the main concern is flexibility. And since a configuration change is 1 commit away using gitops, I don't see this as a big problem. I think that flexibility should supersede other concerns. 

> What use-cases do you see?
Off the top of my head: 
- Ability to deploy downstream/upstream on stage. 
- Use the label selector instead of the pause annotation which would make our helm-operator fork closer to upstream
- Deploy the operator multiple times, same version but different startup parameters to test some feature flag or something

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -18,32 +19,91 @@ import (
...
+	notAllowedStarts := []rune{'.', '-'}
+	if slices.Contains(notAllowedStarts, rune(tag[0])) {
+		return false
+	}
+	return utf8string.NewString(tag).IsASCII()
+}
+
+// IsValidLabel returns true if provided string could be a valid label
+// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+func IsValidLabel(label string) bool {
```
ludydoo, line 67: You could use `apimachinery` `ValidateLabelName` and `IsValidLabelValue`

fleetshard/pkg/central/operator/upgrade.go
ludydoo: Would it be possible to use `https://github.com/containers/image/blob/main/docker/reference/reference.go` to parse the image ?

fleetshard/pkg/central/charts/data/rhacs-operator/templates/rhacs-operator-deployment.yaml
SimonBaeumer: > And since a configuration change is 1 commit away using gitops, I don't see this as a big problem. 

Can you clarify "1 commit away using gitops"? 

> I think that flexibility should supersede other concerns.

This sounds a bit black and white. Flexibility has downsides, like usability and testability. 
A more obvious example: When we want to have ultimate flexibility, why not giving customers a programming language interface instead of the policies UI in StackRox?

> I don't think we can test the actual functionality of a kubernetes label selector, whether it is a "static" selector or a selector expression. We would need some e2e test for that.

fyi @kurlov starts implementing the e2e tests for the labelling.
Hm, I think I was not very clear in my expression. It is the evaluation logic of the label selector, i.e. you can use operands like `in|notin|and` which can not be tested and introduce hidden logic.
Secondly, reading the evaluations is hard for engineers. Preferably reading the code should be easier than writing it:

```
operators:
- image: quay.io/rhacs-operator:4.1.1
  labelSelector: 'rhacs.io/version=4.1.1 and rhacs.io/paused notin (true) and rhacs.io/downstream notin (true)'
- image: quay.io/rhacs-operator:4.2.0
  labelSelector: 'rhacs.io/version=4.2.0 and rhacs.io/paused notin (true) and rhacs.io/downstream notin (true)'
- image: registry.redhat.io/advanced-cluster-security/rhacs-operator:4.2.0
  labelSelector: 'rhacs.io/version:4.2.0 and rhacs.io/paused notin (true) and rhacs.io/downstream=true'

vs.

operators:
- image: quay.io/rhacs-operator:4.1.1
   gitRef: 4.1.1
- image: quay.io/rhacs-operator:4.2.0
   gitRef: 4.2.0-upstream
- image: registry.redhat.io/advanced-cluster-security/rhacs-operator:4.2.0
   gitRef: 4.2.0
```

> - Ability to deploy downstream/upstream on stage.
> - Use the label selector instead of the pause annotation which would make our helm-operator fork closer to upstream
> - Deploy the operator multiple times, same version but different startup parameters to test some feature flag or something

Downstream images would be deployed via the same mechanism as we deploy upstream images. 
Deploying the operator in the same version is also possible via using different `GitRef` parameters, i.e. using `4.0.0-testing`.
An additional disadvantage for pause-reconcile in label selector is that every RHACS Operator must add the exclusion of the `rhacs.io/paused notin (true)` condition explicitly in the operator's label selector.
Every change to the operator images would require more caution of an reviewer + writer to these changes, who often might not work regularly on the repo as they are on-call.

ludydoo: My take on this is that being concerned with readability at this stage is a premature optimization. 

In general, we should be wary of being too smart too early and introducing unnecessary abstractions. All we are looking for here is to inject configuration. The specific use-case in this PR is to inject 1 environment variable for the label selector. But the same concept would apply to other forms of configuration, such as helm values, central CRs and the like. I would strongly advocate to pass config as plainly and directly as possible without inserting abstractions in the middle. 

Maybe once the whole thing is set up, and has been running for a while, we can make more informed decisions *if* we need additional abstractions, if it does become a pain, etc. But the first step should be to pass it as simply and as directly as possible. 

> Can you clarify "1 commit away using gitops"? 

I meant that if there is a mistake in the gitops configuration, we can simply make a corrective commit. It's much simpler than modifying the fleet-* code and re-releasing. 

> When we want to have ultimate flexibility, why not giving customers a programming language interface instead of the policies UI in StackRox?

IMHO this is a low hanging fruit. Instead of having logic to parse the label + selector, and having some logic around that, we can just inject the selector directly from the gitops config repo. And it is inherently more flexible. 

> Secondly, reading the evaluations is hard for engineers

I think the increase "cognitive complexity" is pretty low in this case. 

> which can not be tested and introduce hidden logic.

The label selector, whether the key is hardcoded, or whether we pass an actual expression, needs to be tested in an e2e environment. We cannot test the functionality using unit tests because it involves the rendering of helm charts, the behavior of helm-operator, controller-runtime and etcd.

fleetshard/pkg/runtime/runtime.go
kurlov: `operator` is already in the scope and such name can collide with `operator` package so I changed it to the `operatorDeployment`

fleetshard/pkg/central/operator/upgrade.go
kurlov: Replace whole function with docker container image parse as Ludovic suggested

kurlov (approved):
For Simon

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -33,75 +30,28 @@ type DeploymentConfig struct {
...
+		if _, used := uniqueImages[image]; !used {
+			uniqueImages[image] = true
 			img := chartutil.Values{
 				"deploymentName": deploymentName,
-				"repository":     repo,
-				"tag":            tag,
+				"image":          image,
 				"labelSelector":  operator.GitRef,
 			}
 			operatorImages = append(operatorImages, img)
```
ludydoo: I don't think it is necessary to check for uniqueness of images

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -33,75 +30,28 @@ type DeploymentConfig struct {
...
-
-		if !IsValidLabel(operator.GitRef) {
-			return nil, fmt.Errorf("label selector %s is not valid", operator.GitRef)
+		image := imageReference.String()
+		errs := validation.IsValidLabelValue(operator.GitRef)
+		if errs != nil {
+			return nil, fmt.Errorf("label selector %s is not valid: %v", operator.GitRef, errs)
 		}
 		deploymentName := generateDeploymentName(operator.GitRef)
 		if len(deploymentName) > maxLabelLength {
```
ludydoo, line 47: The name validation should use `apimachineryvalidation.NameIsDNSSubdomain`. See https://github.com/kubernetes/kubernetes/blob/d4fde1e92a83cb533ae63b3abe9d49f08efb7a2f/pkg/apis/apps/validation/validation.go#L456

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,46 +4,56 @@ package operator
...
+		imageReference, err := containerImage.Parse(operator.Image)
+		if err != nil {
+			return nil, err
 		}
-		strs := strings.Split(img, ":")
-		if len(strs) != 2 {
-			return nil, fmt.Errorf("failed to split image and tag from %q", img)
+		image := imageReference.String()
+		errs := validation.IsValidLabelValue(operator.GitRef)
+		if errs != nil {
```
ludydoo, line 43: minor: this could be inlined
```suggestion
		if errs := validation.IsValidLabelValue(operator.GitRef); errs != nil {
```

fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
```diff
@@ -3,15 +3,15 @@
...
 
 operator:
-  # Each item in images should contain `repository` and `tag` key
+  # Each item in images should contain `deploymentName`, `image` and `labelSelector` keys
   # example:
   #   images:
   #   - deploymentName: rhacs-operator-manager-4.0.0
-  #     repository: quay.io/rhacs-eng/stackrox-operator
-  #     tag: 4.0.0
+  #     image: quay.io/rhacs-eng/stackrox-operator:rhacs-operator-manager-4.0.0
```
ludydoo, line 10: should that be 
```suggestion
  #     image: quay.io/rhacs-eng/stackrox-operator:4.0.0
```

fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
```diff
@@ -3,15 +3,15 @@
...
+  #     labelSelector: 4.0.0
   #   - deploymentName: rhacs-operator-manager-4.0.1
-  #     repository: quay.io/rhacs-eng/stackrox-operator
-  #     tag: 4.0.1
+  #     image: quay.io/rhacs-eng/stackrox-operator:rhacs-operator-manager-4.0.1
```
ludydoo, line 13:
```suggestion
  #     image: quay.io/rhacs-eng/stackrox-operator:4.0.1
```

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -4,46 +4,55 @@ package operator
 import (
 	"context"
 	"fmt"
-	"strings"
-
+	containerImage "github.com/containers/image/docker/reference"
```
ludydoo, line 7: will this trigger the `style` validation check? (go-imported)

fleetshard/pkg/central/charts/data/rhacs-operator/values.yaml
```diff
@@ -3,15 +3,15 @@
...
+  #     labelSelector: 4.0.1
```
ludydoo, line 14: No longer needed if using the built-in kubernetes name validation
```suggestion
  #     labelSelector: 4.0.1
```

ludydoo (approved):

PR #1091, opened 2023-06-13
ROX-17485: Reconcile ACS Operator versions

kurlov:
## Description

- Add ACS Operator reconciler. It takes desired operator versions and makes cluster to run exactly those versions
- Add DeleteChart function to charts package

## Checklist (Definition of Done)

- [x] Unit and integration tests added
- [ ] Added test description under `Test manual`
- [x] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)
- [ ] CI and all relevant tests are passing
- [x] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [x] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.
- [x] Add secret to app-interface Vault or Secrets Manager if necessary
- [x] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)
- [x] Check AWS limits are reasonable for changes provisioning new resources

## Test manual

```
# To run tests locally run:
make deploy/dev-fast
```

PR #1118, opened 2023-06-21
ROX-17987: Remove unused Operator deployments

kurlov:
## Description

Add method for removing unused operator deployments based on desired slice of operator images. So any deployed image which **is not** presented in the desired slice should be removed.

## Test manual

```
# To run tests locally run:
make deploy/dev-fast
```

### DIFF
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -117,8 +117,8 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
 	return versionWithReplicas, nil
 }
 
-// Delete removes specified operator deployment based on images from the cluster
-func (u *ACSOperatorManager) Delete(ctx context.Context, images []string) error {
+// RemoveUnusedOperators removes unused operator deployments from the cluster
+func (u *ACSOperatorManager) RemoveUnusedOperators(ctx context.Context, desiredImages []string) error {
 	deployments := &appsv1.DeploymentList{}
 	labels := map[string]string{"app": "rhacs-operator"}
 	err := u.client.List(ctx, deployments,
@@ -129,16 +129,16 @@ func (u *ACSOperatorManager) Delete(ctx context.Context, images []string) error 
 		return fmt.Errorf("failed list operator deployments: %w", err)
 	}
 
-	var deleteDeployments []string
+	var unusedDeployments []string
 	for _, deployment := range deployments.Items {
 		for _, container := range deployment.Spec.Template.Spec.Containers {
-			if container.Name == "manager" && slices.Contains(images, container.Image) {
-				deleteDeployments = append(deleteDeployments, deployment.Name)
+			if container.Name == "manager" && !slices.Contains(desiredImages, container.Image) {
+				unusedDeployments = append(unusedDeployments, deployment.Name)
 			}
 		}
 	}
 
-	for _, deploymentName := range deleteDeployments {
+	for _, deploymentName := range unusedDeployments {
 		deployment := &appsv1.Deployment{}
 		err := u.client.Get(ctx, ctrlClient.ObjectKey{Namespace: operatorNamespace, Name: deploymentName}, deployment)
 		if err != nil && !errors.IsNotFound(err) {
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -129,7 +129,7 @@ func (u *ACSOperatorManager) Delete(ctx context.Context, images []string) error 
 		return fmt.Errorf("failed list operator deployments: %w", err)
 	}
 
-	var deleteDeps []string
+	var deleteDeployments []string
 	for _, deployment := range deployments.Items {
 		for _, container := range deployment.Spec.Template.Spec.Containers {
 			if container.Name == "manager" && slices.Contains(images, container.Image) {
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -117,7 +117,7 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
 	return versionWithReplicas, nil
 }
 
-// RemoveUnusedOperators removes unused operator deployments from the cluster
+// RemoveUnusedOperators removes unused operator deployments from the cluster. It receives a list of operator images which should be present in the cluster and removes all deployments which do not deploy any of the desired images.
 func (u *ACSOperatorManager) RemoveUnusedOperators(ctx context.Context, desiredImages []string) error {
 	deployments := &appsv1.DeploymentList{}
 	labels := map[string]string{"app": "rhacs-operator"}
--- a/fleetshard/pkg/central/operator/upgrade.go
+++ b/fleetshard/pkg/central/operator/upgrade.go
@@ -37,13 +37,13 @@ func parseOperatorImages(images []string) ([]chartutil.Values, error) {
 			return nil, fmt.Errorf("failed to split image and tag from %q", img)
 		}
 		repo, tag := strs[0], strs[1]
-		depName := generateDeploymentName(tag)
-		if len(depName) > maxOperatorDeploymentNameLength {
-			return nil, fmt.Errorf("%s-%s contains more than %d characters and cannot be used as a deployment name", operatorDeploymentPrefix, tag, maxOperatorDeploymentNameLength)
+		deploymentName := generateDeploymentName(tag)
+		if len(deploymentName) > maxOperatorDeploymentNameLength {
+			return nil, fmt.Errorf("%s contains more than %d characters and cannot be used as a deployment name", deploymentName, maxOperatorDeploymentNameLength)
 		}
 		if _, used := uniqueImages[repo+tag]; !used {
 			uniqueImages[repo+tag] = true
-			img := chartutil.Values{"deploymentName": depName, "repository": repo, "tag": tag}
+			img := chartutil.Values{"deploymentName": deploymentName, "repository": repo, "tag": tag}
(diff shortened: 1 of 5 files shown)


fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -129,6 +130,24 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
 	return versionWithReplicas, nil
 }
 
+// DeleteOperator removes specified operator deployment from the cluster
+func (u *ACSOperatorManager) DeleteOperator(ctx context.Context, version string) error {
```
SimonBaeumer, line 134: Where is this function executed and controller versions passed?

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -129,6 +130,24 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
...
+	depName := operatorDeploymentPrefix + "-" + version
```
SimonBaeumer, line 135: I've expected a function which will receive all expected operator versions, list all installed deployments and tries to delete all deployments which are not used anymore. 
Similar to a garbage collector.

Additionally the computation of the name does not match the computation of the helm chart (i.e. the chart truncates the deployment name). 
To keep the logic in one place please move the name computation to Go instead of the Helm template and reuse the function.

SimonBaeumer (changes requested):

fleetshard/pkg/central/operator/upgrade.go
kurlov: It will be executed here https://github.com/stackrox/acs-fleet-manager/pull/1091/files#diff-4a5f912092e037464fef8f19fd9a883a0587f4c5dedc2d15c40b5a4109551addR40-R41 

#1091 depends on this method

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -35,12 +37,13 @@ func parseOperatorImages(images []string) ([]chartutil.Values, error) {
 			return nil, fmt.Errorf("failed to split image and tag from %q", img)
 		}
 		repo, tag := strs[0], strs[1]
-		if len(operatorDeploymentPrefix+"-"+tag) > maxOperatorDeploymentNameLength {
+		depName := generateDeploymentName(tag)
```
SimonBaeumer, line 40:
```suggestion
		deploymentName := generateDeploymentName(tag)
```

dep == dependency

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -115,6 +117,46 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
...
+	err := u.client.List(ctx, deployments,
+		ctrlClient.InNamespace(operatorNamespace),
+		ctrlClient.MatchingLabels(labels),
+	)
+	if err != nil {
+		return fmt.Errorf("failed list operator deployments: %w", err)
+	}
+
+	var deleteDeps []string
+	for _, dep := range deployments.Items {
```
SimonBaeumer, line 133:
```suggestion
	for _, deployment := range deployments.Items {
```

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -115,6 +117,46 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
...
+		for _, c := range dep.Spec.Template.Spec.Containers {
+			if c.Name == "manager" && slices.Contains(images, c.Image) {
+				deleteDeps = append(deleteDeps, dep.Name)
+			}
+		}
+	}
+
+	for _, depName := range deleteDeps {
+		dep := &appsv1.Deployment{}
```
SimonBaeumer, line 142:
```suggestion
		deployment := &appsv1.Deployment{}
```

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -115,6 +117,46 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
...
+	var deleteDeps []string
+	for _, dep := range deployments.Items {
+		for _, c := range dep.Spec.Template.Spec.Containers {
+			if c.Name == "manager" && slices.Contains(images, c.Image) {
+				deleteDeps = append(deleteDeps, dep.Name)
+			}
+		}
+	}
+
+	for _, depName := range deleteDeps {
```
SimonBaeumer, line 141:
```suggestion
	for _, deploymentName := range deleteDeployments {
```

fleetshard/pkg/central/operator/upgrade_test.go
```diff
@@ -161,6 +164,45 @@ func TestOperatorUpgradeDoNotInstallLongTagVersion(t *testing.T) {
...
+
+	// delete another one
+	err = u.Delete(ctx, []string{operatorImage2})
+	require.NoError(t, err)
+	err = fakeClient.Get(context.Background(), client.ObjectKey{Namespace: operatorNamespace, Name: operatorDeployment2.Name}, operatorDeployment2)
+	require.True(t, errors.IsNotFound(err))
+
+	// delete multiple versions
+	err = u.InstallOrUpgrade(ctx, operatorImages, crdTag1)
+	require.NoError(t, err)
```
SimonBaeumer, line 196: Can you assert the operators were installed? 
Imho it is not always necessary to execute `InstallOrUpgrade`, you can also directly pass Deployment object to the fake client.

SimonBaeumer (changes requested):
Changes look good overall, only some nit-picks.

fleetshard/pkg/central/operator/upgrade.go
kurlov: I've changed it from IDE to skip compilation error

fleetshard/pkg/central/operator/upgrade_test.go
kurlov: I really like your suggestion about passing Deployment object to the fake client. 
I also added small check that only deployment is delete (e.g. serviceAccount still persists)

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -115,6 +117,46 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
...
+	labels := map[string]string{"app": "rhacs-operator"}
+	err := u.client.List(ctx, deployments,
+		ctrlClient.InNamespace(operatorNamespace),
+		ctrlClient.MatchingLabels(labels),
+	)
+	if err != nil {
+		return fmt.Errorf("failed list operator deployments: %w", err)
+	}
+
+	var deleteDeps []string
```
SimonBaeumer, line 132:
```suggestion
	var deleteDeployments []string
```

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -115,6 +117,46 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
...
+	for _, deployment := range deployments.Items {
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if container.Name == "manager" && slices.Contains(images, container.Image) {
+				deleteDeps = append(deleteDeps, deployment.Name)
```
SimonBaeumer, line 136: I think you can assume that all deployments are operator deployments because you've selected them by using the label `rhacs-operator`.
To check whether an operator should be garbage collected you can re-compute the name instead of using the images to track an unused deployment.
Alternateviley you could also maintain a data structure internally which maps deployment names to images before the Helm chart is applied, like a cache. 

WDYTH about these approaches?

SimonBaeumer (changes requested):

fleetshard/pkg/central/operator/upgrade.go
kurlov: I've change the method so now it cleans up unused operators. Although, I still stick to use images for checking what to delete because parsing deployment names introduces new potentials error to check.
Do you think it's better to call `parseOperatorImages`?

SimonBaeumer: > Do you think it's better to call parseOperatorImages?

What do you mean? 🤔 

>  Although, I still stick to use images for checking what to delete because parsing deployment names introduces new potentials error to check.

Sounds good!

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -115,6 +117,46 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
 	return versionWithReplicas, nil
 }
 
+// RemoveUnusedOperators removes unused operator deployments from the cluster
```
SimonBaeumer, line 120:
```suggestion
// RemoveUnusedOperators removes unused operator deployments from the cluster. It receives a list of operator images which should be present in the cluster and removes all deployments which do not deploy any of the desired images.
```

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -115,6 +117,46 @@ func (u *ACSOperatorManager) ListVersionsWithReplicas(ctx context.Context) (map[
...
+func (u *ACSOperatorManager) RemoveUnusedOperators(ctx context.Context, desiredImages []string) error {
```
SimonBaeumer, line 121: curiosity: What was the reason for you to not go with `GarbageCollectOperators` or similar?

SimonBaeumer (approved):

fleetshard/pkg/central/operator/upgrade.go
kurlov: I didn't want to put Garbage next to the Operator.
Seriously though, `garbage collect` sounds like it's about interacting with memory, having cycles/references/generations. But we just deleting deployment(s) and letting k8s do the rest for us

PR #1122, opened 2023-06-23
ROX-17712 prevent unnecessary central updates

ludydoo:
This PR adds some logic to prevent fleetshard from updating Centrals if there are no changes, which prevent unnecessary opreator reconciliation loops.

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -4,7 +4,9 @@ package reconciler
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"sync/atomic"
 	"time"
 
@@ -384,10 +386,11 @@ func (r *CentralReconciler) reconcileCentral(ctx context.Context, remoteCentral 
 
 	centralExists := true
 	existingCentral := v1alpha1.Central{}
-	err := r.client.Get(ctx, ctrlClient.ObjectKey{Namespace: remoteCentralNamespace, Name: remoteCentralName}, &existingCentral)
+	centralKey := ctrlClient.ObjectKey{Namespace: remoteCentralNamespace, Name: remoteCentralName}
+	err := r.client.Get(ctx, centralKey, &existingCentral)
 	if err != nil {
 		if !apiErrors.IsNotFound(err) {
-			return errors.Wrapf(err, "unable to check the existence of central %s/%s", central.GetNamespace(), central.GetName())
+			return errors.Wrapf(err, "unable to check the existence of central %v", centralKey)
 		}
 		centralExists = false
 	}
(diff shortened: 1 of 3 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -397,31 +401,67 @@ func (r *CentralReconciler) reconcileCentral(ctx context.Context, remoteCentral
...
 	} else {
-		glog.Infof("Update central %s/%s", central.GetNamespace(), central.GetName())
-		existingCentral.Spec = central.Spec
+		// perform a dry run to see if the update would change anything.
+		// This would apply the defaults and the mutating webhooks without actually updating the object.
+		// We can then compare the existing object with the object that would be resulting from the update.
+		// This will prevent unnecessary operator reconciliation loops.
+
+		desiredCentral := existingCentral.DeepCopy()
+		desiredCentral.Spec = *central.Spec.DeepCopy()
```
SimonBaeumer, line 419: What happens if an annotation was updated? It looks to me that changes on the metadata are not updated anymore.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -397,31 +401,67 @@ func (r *CentralReconciler) reconcileCentral(ctx context.Context, remoteCentral
...
-			return errors.Wrapf(err, "updating central %s/%s", central.GetNamespace(), central.GetName())
+		if err := r.client.Update(ctx, updatedCentral); err != nil {
+			return errors.Wrapf(err, "updating Central %v", centralKey)
 		}
 	}
 
 	return nil
 }
 
+func printCentralDiff(desired, actual *v1alpha1.Central) {
```
SimonBaeumer, line 446: Do you think the log verbosity could be problem?
When ~50 Centrals are reconciled, during an upgrade this log could get very verbose when it logs all Centrals which do not have any errors.
Could the diff be more conditional to not always print it?

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
ludydoo: Hey, yes I've added a feature flag for enabling the diffs. 
As for the central annotations and labels, I think those are also currently completely ignored. 
The current logic ignores them completely, and just takes the labels and annotations from the existing central:
```
existingCentral.Spec = central.Spec
if err := r.client.Update(ctx, &existingCentral); err != nil {
			return errors.Wrapf(err, "updating central %s/%s", central.GetNamespace(), central.GetName())
```

fleetshard/pkg/central/reconciler/reconciler.go
ludydoo: The current logic ignores them completely, and just takes the labels and annotations from the existing central:
```
existingCentral.Spec = central.Spec
if err := r.client.Update(ctx, &existingCentral); err != nil {
			return errors.Wrapf(err, "updating central %s/%s", central.GetNamespace(), central.GetName())
```

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -397,31 +401,103 @@ func (r *CentralReconciler) reconcileCentral(ctx context.Context, remoteCentral
...
 		}
-		existingCentral.Spec = *central.Spec.DeepCopy()
 
-		if err := r.client.Update(ctx, &existingCentral); err != nil {
-			return errors.Wrapf(err, "updating central %s/%s", central.GetNamespace(), central.GetName())
+		glog.Infof("Detected that Central %v is out of date and needs to be updated", centralKey)
+		printCentralDiff(wouldBeCentral, &existingCentral)
+
+		updatedCentral := existingCentral.DeepCopy()
+		updatedCentral.Spec = *central.Spec.DeepCopy()
```
SimonBaeumer, line 450: Would it be possible to build a hash over the labels/annotations + spec and use the computed hash to detect if an update is necessary?
Not useful for printing a diff though.

SimonBaeumer (changes requested):

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -397,31 +401,103 @@ func (r *CentralReconciler) reconcileCentral(ctx context.Context, remoteCentral
...
-			return errors.Wrapf(err, "creating new central %s/%s", remoteCentralNamespace, remoteCentralName)
+			return errors.Wrapf(err, "creating new Central %v", centralKey)
 		}
-		glog.Infof("Central %s/%s created", central.GetNamespace(), central.GetName())
+		glog.Infof("Central %v created", centralKey)
 	} else {
-		glog.Infof("Update central %s/%s", central.GetNamespace(), central.GetName())
-		existingCentral.Spec = central.Spec
+		// perform a dry run to see if the update would change anything.
+		// This would apply the defaults and the mutating webhooks without actually updating the object.
```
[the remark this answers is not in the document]
ludydoo: The dry run will run the update through the api server, setting all default fields, applying (if any) mutating/validating webhooks, etc, without storing the update. The object returned will be as if it was actually updated. Then only is it safe to compare the desired state (after a "would-be" update) with the current state. 

The problem with the central hash is that it contains fields that don't affect the actual Central CR, such as the status. So the fleetshard will still update the Central CR even if it didn't actually needs to change. 

Also, everytime fleetshard does so, it increments a "revision" annotation. But it turns out that in 95% of cases, the only thing that fleetshard updates is the "revision", without modifying any other field on the CR. This creates a lot of traffic for the operator. This is especially visible in the first minutes of deploying a Central, where there could be as many as 20-40 unnecessary reconciliation loops (without actual changes to the CR) triggered due to this. 

So basically this logic is to check whether or not the Central will change as a result of the update or not. And the dry-run is to mitigate any side effects that could be somehow applied either on the apiserver, or through webhooks.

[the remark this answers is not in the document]
ludydoo: https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff/#kubectl-diff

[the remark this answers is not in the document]
ludydoo: This is also how the helm operator detects if resources are changed or not, and if an upgrade needs to be performed or not. https://github.com/operator-framework/helm-operator-plugins/blob/c16a400954f7e43bb987b196e8ecf2f4d2d4ab0f/pkg/reconciler/reconciler.go#L720

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: Why is it less robust? 🤔

SimonBaeumer (approved):

fleetshard/pkg/central/reconciler/reconciler.go
ludydoo: My take on it is that the apiserver must be thought as some sort of black box. It might have various webhooks, and other defaulting strategies that are impossible to account for. So the `existingCentral` might differ from the `central` because of apiserver side effects. By using a dry-run, those side effects are accounted for

ludydoo: Are you suggesting that instead of `DeepCompare`, we would compare the hash of the `would be` central and the `existing` central ? What would be the advantage ?

SimonBaeumer (approved):

PR #1164, opened 2023-07-18
ROX-18428: add cipher package to fleetshard-sync

johannes94:
## Description
In this PR I added a cipher package to fleetshard-sync. It enables encryption and decryption using AES256 GCM encryption algorithm for local development. This will later be used to implement other Cipher implementation e.g AWS KMS and then encrypt secret data on a dataplane cluster so that fleetshard-sync can report the data back to fleet-manager for persistent storage.

See ticket, it's epic and the epic design document for more context.

- ~~[ ] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)~~
- ~~[ ] Check AWS limits are reasonable for changes provisioning new resources~~

## Test manual

- Unit Testing, no integration into fleetshard-sync code yet, so this should be sufficient.

```
# To run tests locally run:
```


fleetshard/pkg/cipher/cipher.go
```diff
@@ -0,0 +1,74 @@
...
+const keySize = 32
+
+// Cipher is the interface used to encrypt and decrypt content
+type Cipher interface {
+	Encrypt(plaintext []byte) ([]byte, error)
+	Decrypt(ciphertext []byte) ([]byte, error)
+}
+
+// AES256Cipher implements encryption and decryption using AES256 GCM
+type AES256Cipher struct {
```
ludydoo, line 21: Would it be possible/was it considered to use AWS KMS to encrypt and decrypt stuff, as well as managing encryption keys ?

johannes94: Yes I did consider that at first. But decided to do it with secrets manager instead. 

There were two main decision drivers for me:
- KMS does not expose the Key for encryption if the key is not an external key. Thus we would have to implement the encryption logic using KMS API. That would make implementation more complex, because you'd have to consider mocking the API for local testing or implementing different encryption logic for local/aws and for the planned implementation for Azure dataplanes sometime next year.
- The other option is to use external keys and store them in KMS, but that would essentially reduce KMS to the capabilities of secrets manager

Because of this I thought the most flexible solution would be to just specify an encryption key, probably with secret mount/ os env variable. Then we could get this key from anywhere we want and the implementation does not depend on any external service.

kurlov (approved):

fleetshard/pkg/cipher/cipher.go
johannes94: Had a slack conversation with @ludydoo. We agreed to use this only for local dev and implement another cipher using AWS KMS.

ludydoo (approved):

PR #1165, opened 2023-07-18
Rox 18494 introduce admin handle interface

ludydoo:
Simply introducing an interface for the admin api handler, as well as a stub for the gitops admin api handler

### DIFF
--- a/internal/dinosaur/pkg/handlers/admin_dinosaur.go
+++ b/internal/dinosaur/pkg/handlers/admin_dinosaur.go
@@ -36,6 +36,8 @@ type AdminCentralHandler interface {
 	Update(w http.ResponseWriter, r *http.Request)
 	// Delete a central
 	Delete(w http.ResponseWriter, r *http.Request)
+	// DbDelete deletes a central from the database
+	DbDelete(w http.ResponseWriter, r *http.Request)
 	// SetCentralDefaultVersion sets the default version for a central
 	SetCentralDefaultVersion(w http.ResponseWriter, r *http.Request)
 	// GetCentralDefaultVersion gets the default version for a central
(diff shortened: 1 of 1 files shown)


internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -392,3 +414,39 @@ func (h adminCentralHandler) GetCentralDefaultVersion(w http.ResponseWriter, r *
...
+
+func (g gitOpsAdminHandler) List(w http.ResponseWriter, r *http.Request) {
+	http.Error(w, "not implemented", http.StatusNotImplemented)
+}
+
+func (g gitOpsAdminHandler) Update(w http.ResponseWriter, r *http.Request) {
+	http.Error(w, "not implemented", http.StatusNotImplemented)
+}
+
+func (g gitOpsAdminHandler) Delete(w http.ResponseWriter, r *http.Request) {
```
SimonBaeumer, line 438: What is your idea on migrating verbs like delete to a configuration based approach? 🤔 

A graveyard list would funny:
```
graveyard:
  - id1
  - id2
```

SimonBaeumer: Someone who restores an instance from the graveyard would be a necromant, doing some dark magic.

SimonBaeumer (approved):

kylape:
It's not clear to me why we'd need a gitops implementation of the admin interface.  Instead, I was expecting to see an interface to access the data in the ConfigMap that would be used by the admin interface and the handler that provides the list of ACS instances to be applied to a given data plane cluster.

kurlov (approved):

PR #750, opened 2023-01-23
incident-20230120-cli

dhaus67:
## Description

```
./dataplane-migrators incident-20230120 --help                                                                                                                                                                                                                                                                        
This migration is based on the incident 20230120, which was reported due to a multitude of issues with the current setup
of the auth providers and groups.

For more information, visit: https://srox.slack.com/archives/C04L0BUNRKN/p1674263826620659

Usage:
  dataplane-migrator incident-20230120 [flags]

Flags:
      --client-id string          client ID for the dynamic client associated with central.
      --client-secret string      client secret for the dynamic client associated with central.
  -h, --help                      help for incident-20230120
      --id string                 id of the central instance
      --name string               name of the central instance
      --org-id string             organisation ID associated with the central.
      --output-directory string   output directory where the access logs should be stored.
  -u, --url string                URL of the central instance which should be migrated. The format should be https://<hostname>:<port> and should be the UI URL, not the data URL.
```

**How to execute migration**

1. Get all Centrals from GABI:
```
./stage-action.sh gabi "SELECT id,name,organisation_id,CONCAT('https://acs-',id,'.',host),client_id,client_secret FROM central_requests WHERE deleted_at IS NULL ORDER BY created_at DESC;" | jq '.result[1:][] | @csv ' --raw-output > issue-20230120-stage.csv
```

2. And run the script that will use that file:
```
./scripts/hack/fix-incident-20230120.sh --central-list-path ./issue-20230120-stage.csv
```

## Checklist (Definition of Done)
<!-- Please strikethrough options not relevant using two tildes ~~Text~~. Do not delete non relevant options -->
- [ ] Unit and integration tests added
- [ ] Added test description under `Test manual`
- [ ] Evaluated and added CHANGELOG.md entry if required
- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)
- [ ] CI and all relevant tests are passing
- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.

## Test manual

**TODO:** Add manual testing efforts

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup OCM_OFFLINE_TOKEN=<ocm-offline-token> OCM_ENV=development
make verify lint binary test test/integration
```

PR #1186, opened 2023-08-01
ROX-17336: Migrate authn/z declarative config

dhaus67:
## Description

```
./dataplane-migrator migrate authn-z --help                                                                                                                                                                                                                                                                          
Run the migration of authn/z resources to declarative configuration.
For existing instances, the default authn/z resources (auth provider and groups) have been created imperatively via
API calls. This makes migrating things hard (that's why we have to write migrations like these).

While the secret has been created already containing the declarative configuration, they are currently not applied
since the auth provider cannot be created due to name clashes with the existing one.

The goal of this migration is to do the following:
* Retrieve potential custom groups added by users within the default auth provider.
* Remove the default auth provider that's created imperatively.
* Re-apply any custom groups that have been created previously.

The time until the auth provider is added after removal should be no more than 20 seconds, as this is the interval
Central will read and reconcile declarative configurations. The expectation is that the downtime should be less than 1
minute. It's worthwhile to note that users will have to re-login, as the auth provider ID changes, thus tokens issued
by the specific auth provider ID will be seen as invalid.

Usage:
  dataplane-migrator migrate authn-z [flags]

Flags:
  -h, --help          help for authn-z
      --id string     id of the central instance
      --name string   name of the central instance
      --stage         specify whether running against stage
  -u, --url string    URL of the central instance which should be migrated. The format should be https://<hostname>:<port> and
                      should be the UI URL, not the data URL.
```

As a reference, [a authn/z migration was done previously in a similar fashion](https://github.com/stackrox/acs-fleet-manager/pull/750).

**Note: this is not expected to be merged.**

### How to run a migration

1. Get all Centrals from GABI:
```bash
./stage-action.sh gabi "SELECT id,name,CONCAT('https://acs-',id,'.',host),issuer FROM central_requests WHERE deleted_at IS NULL AND region = 'us-east-1' AND name NOT LIKE '%probe%' ORDER BY created_at DESC;" | jq '.result[1:][] | @csv ' --raw-output > input.csv
```

2. Run the script that will use that file as input:
```bash
./scripts/hack/migrate-authn-z.sh --central-list-path ./input.csv
```

<!-- Please strikethrough options not relevant using two tildes ~~Text~~. Do not delete non relevant options -->

## Test manual

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,8 @@
 cmd/fleet-manager/fleet-manager
 /fleetshard-sync
 /acsfleetctl
+/dataplane-migrator
+cmd/dataplane-migrator/dataplane-migrator
 # Ignore generated templates
 /templates/*-template.json
 
--- a/Makefile
+++ b/Makefile
@@ -303,7 +303,11 @@ acsfleetctl:
 	GOOS="$(GOOS)" GOARCH="$(GOARCH)" $(GO) build $(GOARGS) -o acsfleetctl ./cmd/acsfleetctl
 .PHONY: acsfleetctl
 
-binary: fleet-manager fleetshard-sync probe acsfleetctl
+dataplane-migrator:
+	GOOS="$(GOOS)" GOARCH="$(GOARCH)" $(GO) build $(GOARGS) -o dataplane-migrator ./cmd/dataplane-migrator
+.PHONY: dataplane-migrator
+
+binary: fleet-manager fleetshard-sync probe acsfleetctl dataplane-migrator
 .PHONY: binary
 
 # Install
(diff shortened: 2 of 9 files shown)


dhaus67:
* main
  * **PR #1186** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1186" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1186?utm_source=stack-comment).

stehessel (approved):

dhaus67:
Migration has been done.

As always, we will close this PR and _hopefully_ such extensive migrations won't be required in the future for authn/z now that we use declarative configuration.

PR #1190, opened 2023-08-04
ROX-16090: Start data plane CD on integration environment.

porridge:
## Description

This change also splits the IDP setup script to resolve some dependency loops between it and the terraform config.

- [ ] ~RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)~
- [ ] ~Check AWS limits are reasonable for changes provisioning new resources~

## Test manual

CI should suffice.


.github/workflows/deploy-integration.yaml
```diff
@@ -0,0 +1,16 @@
...
+      - main
+
+jobs:
+  call-deploy-workflow:
+    uses: ./.github/workflows/deploy-data-plane.yaml
+    secrets: inherit # pragma: allowlist secret
+    with:
+      acs_environment: integration
+      github_environment: integration
+      deploy_clusters: "acs-int-us-01"
```
ebensh, line 15: What happened to the plan to update the naming scheme?

docs/development/setup-osd-cluster-idp.md
```diff
@@ -14,11 +16,11 @@ The following IdPs will be created:
 - OIDC IdP using auth.redhat.com as backend.
 
 Before executing the script that manages the IdP setup, you have to ensure you are logged in with OCM.
-Based on the environment, you have to choose between `rhacs-managed-service-stage` or `rhacs-managed-service-prod` account.
+Based on the environment, you have to choose between `rhacs-managed-service-stage`, `rhacs-managed-service-integration` or `rhacs-managed-service-prod` account.
```
ebensh, line 19: This may be a nit but I thought the logical ordering was:
int -> stage -> prod
It would make sense to put integration first then for consistency. If I'm wrong and it is:
stage -> int -> prod
And "int" is "pre-prod" then it makes sense.

If this change is made, please reflect that below as well (stage|integration|prod -> integration|stage|prod)

dp-terraform/cd-robot-account-setup.sh
```diff
@@ -0,0 +1,141 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+# shellcheck source=scripts/lib/external_config.sh
+source "$SCRIPT_DIR/../scripts/lib/external_config.sh"
+
+if [[ $# -ne 2 ]]; then
+    echo "Usage: $0 [environment] [cluster]" >&2
+    echo "Known environments: stage integration prod"
+    echo "Cluster typically looks like: acs-{environment}-dp-01"
```
ebensh, line 12: This is no longer true, but we can leave it there for consistency until the naming scheme is updated

dp-terraform/cd-robot-account-setup.sh
```diff
@@ -0,0 +1,141 @@
...
+    chamber write -b secretsmanager "cluster-${CLUSTER_NAME}" "${key}" "${value}" --skip-unchanged
+}
+
+case $ENVIRONMENT in
+  stage)
+    EXPECT_OCM_ID="2ECw6PIE06TzjScQXe6QxMMt3Sa"
+    ;;
+
+  integration)
+    EXPECT_OCM_ID="2QVFzUvsbMGheHhoUDjtG0tpJ08"
```
ebensh, line 47: This does not match the value in https://docs.engineering.redhat.com/pages/viewpage.action?pageId=288990393#Environments(Dev,Int,Stg,Prd)-ACSManagedServiceEnvironments , which is `2NgB6hzFyCWU2EvO2Ro3MX4jEfF`. Which value is correct? Are both from OCM's production environment and for the correct organization?

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -45,6 +45,21 @@ case $ENVIRONMENT in
     SECURED_CLUSTER_ENABLED="true"
     ;;
 
+  integration)
+    FM_ENDPOINT="https://qj3layty4dynlnz.api.integration.openshift.com"
```
ebensh, line 49: Please add this to https://docs.engineering.redhat.com/pages/viewpage.action?pageId=288990393#Environments(Dev,Int,Stg,Prd)-ACSManagedServiceEnvironments

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -45,6 +45,21 @@ case $ENVIRONMENT in
...
+    OBSERVABILITY_GITHUB_TAG="master"
+    OBSERVABILITY_OBSERVATORIUM_GATEWAY="https://observatorium-mst.api.stage.openshift.com"
```
ebensh, line 51: Will this be a problem using the Observatorium stage environment? Will Stage and Int environments clobber each other?

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
```diff
@@ -45,6 +45,21 @@ case $ENVIRONMENT in
...
+    OBSERVABILITY_OPERATOR_VERSION="v4.2.1"
+    OPERATOR_USE_UPSTREAM="false"
+    OPERATOR_CHANNEL="stable"
+    OPERATOR_VERSION="v4.1.0"
+    FLEETSHARD_SYNC_CPU_REQUEST="${FLEETSHARD_SYNC_CPU_REQUEST:-"200m"}"
+    FLEETSHARD_SYNC_MEMORY_REQUEST="${FLEETSHARD_SYNC_MEMORY_REQUEST:-"1024Mi"}"
+    FLEETSHARD_SYNC_CPU_LIMIT="${FLEETSHARD_SYNC_CPU_LIMIT:-"1000m"}"
+    FLEETSHARD_SYNC_MEMORY_LIMIT="${FLEETSHARD_SYNC_MEMORY_LIMIT:-"1024Mi"}"
+    SECURED_CLUSTER_ENABLED="true"
```
ebensh, line 60: Is this going to work correctly when reporting values to the Dogfood instance?

dp-terraform/osd-cluster-idp-setup.sh
```diff
@@ -70,6 +58,10 @@ case $ENVIRONMENT in
     EXPECT_OCM_ID="2ECw6PIE06TzjScQXe6QxMMt3Sa"
     ;;
 
+  integration)
+    EXPECT_OCM_ID="2QVFzUvsbMGheHhoUDjtG0tpJ08"
```
ebensh, line 62: Same q here

ebensh (changes requested):
I like the cleanup for sure :) Thank you for splitting it out.

dp-terraform/cd-robot-account-setup.sh
porridge: Both are correct. `2Ng...` is the org id. `2QVF...` is the account id :-)

.github/workflows/deploy-integration.yaml
porridge: As it ran deeper and deeper I decided to not block the cluster creation on the decision. It was dragging for far too long already.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: I will once it's actually up.

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: Excellent point!
Some time ago @stehessel mentioned we could use the stage observatorium for this environment, but I guess we should first disambiguate the metrics. Taking a quick look at the config, the `tenant` helm value looks tempting, but I doubt we can just set it to something like `rhacs-int` in this case without some external setup first?

I guess the same concern applies to the `logging`, `cloudwatch` and perhaps `audit-logs` subcharts?

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
porridge: Just another secured cluster, no? But OK, let's make baby steps and set this to false for now...

dp-terraform/osd-cluster-idp-setup.sh
porridge: Same answer :-)

dp-terraform/cd-robot-account-setup.sh
ebensh: Ah dangit - good catch, thank you

dp-terraform/osd-cluster-idp-setup.sh
ebensh: :D

dp-terraform/helm/rhacs-terraform/terraform_cluster.sh
stehessel: Prometheus metrics have `clusterName` as an external label, so I don't see an issue with using the stage observatorium tenant. `cloudwatch` subchart should also be fine to re-use, just need to adapt the environment and cluster name.

porridge: @stehessel can you please clarify what you mean by

> just need to adapt the environment and cluster name

Is this something I need to do beforehand?
:sweat_smile:

Also, what about the `logging` and `audit-logs` subcharts? Are you familiar with them?

stehessel: Cluster name and environment are setup via https://github.com/stackrox/acs-fleet-manager/blob/main/dp-terraform/helm/rhacs-terraform/terraform_cluster.sh#L18-L19. For `logging` and `audit-logs` I'd suggest to take a look at the current Helm values for stage. You need to adapt a few parameters from `stage-dp-02` to `integration-...`.

porridge: I'm still at a loss at where I should be looking for "the current Helm values for stage". AFAICT the helm values are being set dynamically by the shell script you pointed at, so as long as `deploy_clusters` parameter in the GHA workflow file is set correctly (and it is) I don't need to change anything in the charts :thinking:

I had a closer look and determined the following w.r.t. to the child charts:

## audit-logs

Seems to configure feeding k8s (?) audit logs into AWS cloudwatch service in the configured account. Since `integration` uses a separate account, there should be no clash with `staging`.

BTW helm is called with`  --set audit-logs.annotations.rhacs\\.redhat\\.com/cluster-name="${CLUSTER_NAME}" \` but it's not clear whether this affects the streamed logs in any way.

## cloudwatch

Watches AWS resources and exports metrics for prometheus to scrape, so does not directly touch anything related to other environments.

## logging

This one is a bit mysterious to me since the README assumes knowledge of OCP technologies that I'm unfamiliar with. Either way, the log forwarder output configuration has `groupPrefix: {{ .Values.groupPrefix | quote }}` which we set as `  --set logging.groupPrefix="${CLUSTER_NAME}" \` when calling helm. So while it's not clear to me where these logs end up, I think this should ensure the results are not mixed with other clusters.

## observability

As discussed above, `clusterId` label is being set based on `  --set observability.clusterName="${CLUSTER_NAME}" \`

porridge: @ebensh can you please approve?

stehessel: I meant that there is probably nobody who knows all the charts and their values. So you could look at the generated Helm values for stage, note anything that is tied to the cluster name / env / AWS account, and either confirm that they are dynamically generated or set them manually. Which it looks like is what you did, so all good from my POV.

> So while it's not clear to me where these logs end up, I think this should ensure the results are not mixed with other clusters.

The logs should end up in CloudWatch. In the secrets manager there should be AWS credentials for this.

ebensh (approved):

PR #1196, opened 2023-08-10
ROX-18914: Add E2E tests for Canary Upgrade

kurlov:
## Description

Adds E2E tests to cover:
- Run ACS Operator from Fleetshard-sync service
- Run multiple ACS Operators
- Create Central with Canary Upgrade Feature turned on
- Upgrade Central

*OpenshiftCI integration will be implemented in upcoming PR*

## Test manual

```
# To run tests locally run:
```


e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,279 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	namespace = "acsms"
+)
+
+var _ = Describe("Fleetshard-sync Canary Upgrade", func() {
```
SimonBaeumer, line 26: nit: confusion before, this would be targeted, canary would be the implementation on the gradual rollout logic.
WDYTH?

```suggestion
var _ = Describe("Fleetshard-sync Targeted Upgrade", func() {
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,279 @@
...
+				WithPolling(defaultPolling).
+				Should(Succeed())
+			Eventually(getOperatorDeployments(ctx)).
+				WithTimeout(waitTimeout).
+				WithPolling(defaultPolling).
+				Should(HaveLen(2))
+
+		})
+
+		It("should run only the latest operator", func() {
```
SimonBaeumer, line 86: AFAIK, this test removes one operator
```suggestion
		It("should delete the removed operator", func() {
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,279 @@
...
+			ctx := context.Background()
+			twoOperatorVersionConfig := `
```
SimonBaeumer, line 88: imho it is sufficient because the context is clear and not multiple variables are used
```suggestion
			operatorConfig := `
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,279 @@
...
+					return fmt.Errorf("wrong image")
+				}
+				return nil
+			}).WithTimeout(waitTimeout).WithPolling(defaultPolling).Should(Succeed())
+		})
+	})
+
+})
+
+func updateOperatorConfig(ctx context.Context, configYAML string) error {
```
SimonBaeumer, line 184: Here passing the operator config structs and marshalling the YAML in this function.
- expected input is clear which config should be passed because of strong typing
- usage search is not working when using the unstructured approach. If someone modifies the data structure they will be notified likely only by CI that these tests failed

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return central, nil
+}
+
+func getCentralDeployment(ctx context.Context, name string, namespace string) (*appsv1.Deployment, error) {
+	deployment := &appsv1.Deployment{}
+	centralKey := ctrlClient.ObjectKey{Namespace: namespace, Name: name}
+	err := k8sClient.Get(ctx, centralKey, deployment)
```
SimonBaeumer, line 232: Is it possible to remove the `if err != nil` cases with an assertion? 
Or return `return deployment, err`? I think it is okay in a test to not 100% align with best pracitces but prefer an idomatic approach which reduces complexity in the code flow.

pkg/features/list.go
```diff
@@ -6,4 +6,7 @@ var (
 
 	// PrintCentralUpdateDiff enables printing the diff of the central update
 	PrintCentralUpdateDiff = registerFeature("Print the diff of the central update", "RHACS_PRINT_CENTRAL_UPDATE_DIFF", false)
+
+	// UseOperatorsConfigMap makes Fleetshard-sync service use ACS Operators configuration from ConfigMap. It is useful for E2E testing
+	UseOperatorsConfigMap = registerFeature("Use ACS Operators configuration from ConfigMap", "RHACS_USE_OPERATORS_CONFIGMAP", false)
```
SimonBaeumer, line 11: nit: Keeping the name a bit more general to debug mode gives more flexibility later, i.e. to use to mock fleet-manager API responses.

```suggestion
	UseOperatorsConfigMap = registerFeature("Use ACS Operators configuration from ConfigMap", "RHACS_DEBUG_MODE", false)
```

dev/env/manifests/rhacs-operator/03-operators-config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operator-config
+  namespace: "$ACSMS_NAMESPACE"
+data:
+  operator-config.yaml: |-
+    - gitref: 4.0.0
+      image: quay.io/rhacs-eng/stackrox-operator:4.1.1
```
SimonBaeumer, line 9: `gitRef` and image tag are different, this would act as a bad example. Can you adjust the values to match each other?

```suggestion
    - gitRef: 4.1.1
      image: quay.io/rhacs-eng/stackrox-operator:4.1.1
```

pkg/features/list.go
SimonBaeumer: I think you not had the time since our sync, commenting here for visibility to change the name.

fleetshard/pkg/runtime/runtime.go
```diff
@@ -250,28 +250,45 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
...
+
+	if features.UseOperatorsConfigMap.Enabled() {
+		configMapOperators, err := r.operatorManager.ReadOperatorConfigurationFromConfigMap(ctx)
+		if err != nil {
+			glog.Warningf("Failed reading operators configMap: %v", err)
+		}
+
+		if len(configMapOperators) == 0 {
+			glog.Infof("Extracted zero operators from configmap")
+		}
```
SimonBaeumer, line 264: Imho can be logged unconditionally.

```suggestion

        glog.Infof("Reading operator config map, extracted %d operators from configmap", len(configMapOperators))

```

fleetshard/pkg/runtime/runtime.go
```diff
@@ -250,28 +250,45 @@ func (r *Runtime) deleteStaleReconcilers(list *private.ManagedCentralList) {
...
+
+		desiredOperatorConfigs = configMapOperators
+	} else {
+		desiredOperatorConfigs = []operator.OperatorConfig{{
+			GitRef: "4.1.0",
+			Image:  "quay.io/rhacs-eng/stackrox-operator",
+		}}
```
SimonBaeumer, line 271: Adding this as a default operator might cause confusion. If I start fleetshard without it, the operator is created by default, but as soon as I create the configmap the operator is deleted and the other config is active.
This would remove the default operator unintentionally.

fleetshard/pkg/central/operator/upgrade.go
```diff
@@ -182,6 +186,26 @@ func (u *ACSOperatorManager) RemoveUnusedOperators(ctx context.Context, desiredI
 	return nil
 }
 
+// ReadOperatorConfigurationFromConfigMap reads Operator deployment configuration from ConfigMap
+func (u *ACSOperatorManager) ReadOperatorConfigurationFromConfigMap(ctx context.Context) ([]OperatorConfig, error) {
```
SimonBaeumer, line 190: nit:
```suggestion
func (u *ACSOperatorManager) ReadOperatorConfigFromConfigMap(ctx context.Context) ([]OperatorConfig, error) {
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+	return deployment, err
+}
+
+func validateOperatorDeployment(ctx context.Context, GitRef string, image string) error {
```
SimonBaeumer, line 250:
```suggestion
func validateOperatorDeployment(ctx context.Context, gitRef string, image string) error {
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+)
+
+const (
+	namespace = "acsms"
+)
+
+var _ = Describe("Fleetshard-sync Targeted Upgrade", func() {
+
+	BeforeEach(func() {
+		if !runCanaryUpgradeTests {
```
SimonBaeumer, line 30: Can you use the feature flag for this if-case? If we enable the feature-flag the tests automatically are re-activated and stop us from merging broken states accidentially.

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+			Expect(err).To(BeNil())
+
+			Eventually(validateOperatorDeployment(ctx, "4.1.1", "quay.io/rhacs-eng/stackrox-operator:4.1.1")).
+				WithTimeout(waitTimeout).
+				WithPolling(defaultPolling).
+				Should(Succeed())
+
+		})
+
+		It("should deploy two operators", func() {
```
SimonBaeumer, line 63:
```suggestion
		It("should deploy two operators with different versions", func() {
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+				{
+					GitRef: "4.1.2",
+					Image:  "quay.io/rhacs-eng/stackrox-operator:4.1.2",
+				},
+			}
+
+			err := updateOperatorConfig(ctx, twoOperatorVersionConfig)
+			Expect(err).To(BeNil())
+
+			Eventually(validateOperatorDeployment(ctx, "4.1.1", "quay.io/rhacs-eng/stackrox-operator:4.1.1")).
```
SimonBaeumer, line 78: Can you wrap this with `It` func blocks to provide more info in test failures?

Like:

```
It(`should reconcile operator via label selector 4.1.1`) {}
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+				WithTimeout(waitTimeout).
+				WithPolling(defaultPolling).
+				Should(Succeed())
+			Eventually(validateOperatorDeployment(ctx, "4.1.2", "quay.io/rhacs-eng/stackrox-operator:4.1.2")).
```
SimonBaeumer, line 82: Add `It` func block

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+				WithPolling(defaultPolling).
+				Should(Succeed())
+			Eventually(getOperatorDeployments(ctx)).
+				WithTimeout(waitTimeout).
+				WithPolling(defaultPolling).
+				Should(HaveLen(2))
+
+		})
+
+		It("should delete the removed operator", func() {
```
SimonBaeumer, line 93: 👍

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+				Should(HaveLen(1))
+
+		})
+
+	})
+
+	Describe("should upgrade the central", func() {
+
+		BeforeEach(func() {
+			Skip("Skipping canary upgrade test")
```
SimonBaeumer, line 118: Add feature flag check

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+		})
+
+		var createdCentral *public.CentralRequest
+		var centralNamespace string
+
+		It("create central", func() {
```
SimonBaeumer, line 124: Does this have more expectations? for example, creating Central in a specific version?

```suggestion
		It("creates central", func() {
```

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+			createdCentral = &resp
+			Expect(err).To(BeNil())
+			Expect(constants.CentralRequestStatusAccepted.String()).To(Equal(createdCentral.Status))
+
+			centralNamespace, err = services.FormatNamespace(createdCentral.Id)
+
+			ctx := context.Background()
+			oneOperatorVersionConfig := []operator.OperatorConfig{{
+				GitRef: "4.0.0",
+				Image:  "quay.io/rhacs-eng/stackrox-operator:4.1.1",
```
SimonBaeumer, line 143: This gives users a bad example on how `gitRef` and image tag should be used, if you want to have different values use the commit hash the tag is referring to.

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+			}}
+			err = updateOperatorConfig(ctx, oneOperatorVersionConfig)
+			Expect(err).To(BeNil())
+
+			Eventually(func() error {
+				centralCR, err := getCentralCR(ctx, createdCentral.Name, centralNamespace)
+				if err != nil {
+					return fmt.Errorf("failed finding central CR: %v", err)
+				}
+				if centralCR.Labels["rhacs.redhat.com/version-selector"] != "4.0.0" {
```
SimonBaeumer, line 153: Getter instead of direct access?

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,290 @@
...
+					return fmt.Errorf("wrong version-selector")
```
SimonBaeumer, line 154: What was expected? Does a user get enough information to debug the test failures easily?

e2e/e2e_suite_test.go
```diff
@@ -18,18 +18,20 @@ import (
...
+	routesEnabled         bool
+	route53Client         *route53.Route53
+	waitTimeout           = getWaitTimeout()
+	dpCloudProvider       = getEnvDefault("DP_CLOUD_PROVIDER", "standalone")
+	dpRegion              = getEnvDefault("DP_REGION", "standalone")
+	authType              = "OCM"
+	fleetManagerEndpoint  = "http://localhost:7000"
+	runAuthTests          bool
+	runCentralTests       bool
+	runCanaryUpgradeTests bool
```
SimonBaeumer, line 34: Use feature flag instead

e2e/e2e_suite_test.go
```diff
@@ -18,18 +18,20 @@ import (
...
+	dnsEnabled            bool
+	routesEnabled         bool
+	route53Client         *route53.Route53
+	waitTimeout           = getWaitTimeout()
+	dpCloudProvider       = getEnvDefault("DP_CLOUD_PROVIDER", "standalone")
+	dpRegion              = getEnvDefault("DP_REGION", "standalone")
+	authType              = "OCM"
+	fleetManagerEndpoint  = "http://localhost:7000"
+	runAuthTests          bool
+	runCentralTests       bool
```
SimonBaeumer, line 33: This should be defaulted to `true`

SimonBaeumer (changes requested):

e2e/e2e_suite_test.go
kurlov: Mapped value reading to feature flag

fleetshard/pkg/runtime/runtime.go
kurlov: I am not sure what would be the best approach here. It looks we need some sort of default version for local development until canary is roll out.I think it is explicitly expected that default operator is deleted if some custom config is used

e2e/e2e_canary_upgrade_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+const (
+	namespace = "acsms"
+)
+
+var _ = Describe("Fleetshard-sync Targeted Upgrade", func() {
+
+	BeforeEach(func() {
+
+		if !runCanaryUpgradeTests || !features.StandaloneMode.Enabled() {
```
SimonBaeumer, line 32: The test can run always when the targeted upgrades feature is enabled, at least in the e2e test suite. 

```suggestion
		if feature.TargetedUpgrades.Enabled() || !features.StandaloneMode.Enabled() {
```

dev/env/scripts/lib.sh
```diff
@@ -139,6 +139,7 @@ init() {
     export IGNORE_REPOSITORY_DIRTINESS=${IGNORE_REPOSITORY_DIRTINESS:-$IGNORE_REPOSITORY_DIRTINESS_DEFAULT}
     export DEBUG_PODS=${DEBUG_PODS:-$DEBUG_PODS_DEFAULT}
     export RHACS_TARGETED_OPERATOR_UPGRADES=${RHACS_TARGETED_OPERATOR_UPGRADES:-$RHACS_TARGETED_OPERATOR_UPGRADES_DEFAULT}
+    export RHACS_STANDALONE_MODE=${RHACS_STANDALONE_MODE:-$RHACS_STANDALONE_MODE_DEFAULT}
```
SimonBaeumer, line 142: What is the difference between `RHACS_USE_OPERATORS_CONFIGMAP` and `RHACS_STANDALONE_MODE`? I would prefer to have one environmental variable (`RHACS_STANDALONE_MODE`) for simplicity.

SimonBaeumer (changes requested):

dev/env/scripts/lib.sh
kurlov: There is no `RHACS_USE_OPERATORS_CONFIGMAP` anymore

SimonBaeumer: Than I found some dead code :skull: 

https://github.com/stackrox/acs-fleet-manager/pull/1196/files#diff-03ca885f8fbb44cbdf5d57ecccdefe690f69d32d4cf79db86a78f34723e07986R59-R62

PR #1198, opened 2023-08-11
ROX-18427: read central DB ID override from optional ConfigMap

vladbologa:
## Description

Currently the name of a Central DB is fixed and based on the Central's ID. E.g. for the ID `cjmugfm9rusboqitae5g` the DB cluster must be called `rhacs-cjmugfm9rusboqitae5g-db-cluster`.

This PR gives the possibility to override the default naming scheme. For example, if the above DB is restored from backup into `rhacs-examplebackup-db-cluster`, then fleetshard-sync can be instructed to use the alternate name in the following way:
`kubectl create configmap central-db-override --namespace=rhacs-cjmugfm9rusboqitae5g --from-literal=databaseID=examplebackup`

## Test manual

Tested E2E locally by creating a ConfigMap as described above and confirming that Central can connect to an overridden DB.

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -76,8 +76,9 @@ const (
 	dbUserTypeCentral    = "central"
 	dbCentralUserName    = "rhacs_central"
 
-	centralDbSecretName       = "central-db-password" // pragma: allowlist secret
-	centralDeletePollInterval = 5 * time.Second
+	centralDbSecretName        = "central-db-password" // pragma: allowlist secret
+	centralDbOverrideConfigMap = "central-db-override"
+	centralDeletePollInterval  = 5 * time.Second
 
 	sensibleDeclarativeConfigSecretName = "cloud-service-sensible-declarative-configs" // pragma: allowlist secret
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -1022,6 +1027,29 @@ func (r *CentralReconciler) ensureCentralDeleted(ctx context.Context, remoteCent
 	return globalDeleted, nil
 }
```
johannes94, line 1029: I'd move the comment to above the function as godoc comment, so that you can easily see it with editor tools. Also I'd change it a little to be more clear what actually happens:

```
// getDatabaseID returns the cloud database ID for a central tenant. 
// By default the database ID is equal to the centralID. It can be overridden by a ConfigMap.
```

johannes94 (approved):
I think the override approach is more safe than relying on tags, so i'm fine with that. Just a minor thing about the comment.

At some point we probably want to integrate the DB ID override to the GitOps configuration. Could be worth creating a ticket for that.

fleetshard/pkg/central/reconciler/reconciler.go
vladbologa: Done

PR #1199, opened 2023-08-14
Update alertmanager grouping confiugration

kylape:
Alerts in PagerDuty are currently all grouped together, even when it doesn't make any sense.

So instead, group PD alerts by rhacs_instance_id if rhacs_instance_id is on the alert; otherwise group by cluster_id.

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
@@ -10,12 +10,20 @@ stringData:
     route:
       receiver: managed-rhacs-pagerduty
       repeat_interval: 12h
+      group_by:
+      - cluster_id
       routes:
       - receiver: managed-rhacs-deadmanssnitch
         repeat_interval: 5m
+        continue: false
         match:
           alertname: DeadMansSwitch
           observability: managed-rhacs
+      - receiver: managed-rhacs-pagerduty
+        group_by:
+        - rhacs_instance_id
+        matchers:
+        - "rhacs_instance_id =~ \".*\""
     receivers:
     - name: managed-rhacs-pagerduty
       pagerduty_configs:


dp-terraform/helm/rhacs-terraform/charts/observability/templates/01-operator-03-secret-alertmanager.yaml
```diff
@@ -10,12 +10,20 @@ stringData:
     route:
       receiver: managed-rhacs-pagerduty
       repeat_interval: 12h
+      group_by:
+      - cluster_id
       routes:
       - receiver: managed-rhacs-deadmanssnitch
```
JoukoVirtanen, line 16: This was not changed in this PR, but is this a typo?
```suggestion
      - receiver: managed-rhacs-deadmansswitch
```

JoukoVirtanen: I don't think this is affecting anything.

JoukoVirtanen (approved):

stehessel:
I asked SRE some time ago about this (https://redhat-internal.slack.com/archives/C0313JYKH8W/p1689256242034729), and their recommendation is to prefer grouping when in doubt and rely on the PagerDuty "intelligent" grouping. Although I don't think that works very well, so I would prefer the dedup keys you proposed. Did you test how the alertmanager `group_by` works when in PagerDuty the intelligent grouping is enabled?

kylape:
> their recommendation is to prefer grouping when in doubt and rely on the PagerDuty "intelligent" grouping

So their counterproposal would be to send every alert as a separate event to PD and then rely on PD intelligent grouping, right?  Just to be clear, right now all alerts are sent as one big event to PD, i.e. everything is grouped together in Alertmanager, preventing PD from doing any further grouping at all.

> Did you test how the alertmanager group_by works when in PagerDuty the intelligent grouping is enabled?

I have not. I was kind of planning to test that in stage.  Do you think I should test that outside of stage?

stehessel:
> So their counterproposal would be to send every alert as a separate event to PD and then rely on PD intelligent grouping, right? Just to be clear, right now all alerts are sent as one big event to PD, i.e. everything is grouped together in Alertmanager, preventing PD from doing any further grouping at all.

I thought that the Alertmanager alerts first appear as alerts (https://redhat.pagerduty.com/alerts), which are them grouped as incidents based on different algorithms ("intelligent grouping" or explicit grouping keys). Tbh I'm not sure where the alertmanager grouping comes in or if it matters. 

Their proposal is indeed to let PagerDuty do the grouping, and favor less grouping in general. 
I think from their perspective, they want to avoid multiple pages for the same root cause at all costs (e.g. if we dedup by instance_id and the cluster goes down, we would get alerted for every instance vs just once for all). From my point of view the problem with that is that the PagerDuty grouping does not seem to work very well (maybe that is because the Alertmanager grouping is already problematic as you suggest), and it's hard to read alerts when they get bunched up into the same incident all the time.

kylape:
Take a look at this event received by PD yesterday: https://redhat.pagerduty.com/alerts/Q3DQB10U1T10BP/log_entries/RR7JTYGOQ8ISSQO8XJOGYXJKHG.  It shows how all the alerts in Alertmanager (except DeadMansSnith) are sent as a single event to PD.

kylape:
> hard to read alerts when they get bunched up into the same incident all the time.

I would describe the problem a bit more urgently.  When I tried to trigger an alert by forcing a production Central pod (my personal instance) into a crashloop, no alert was received by the on-all engineer because there was already a long-running incident in PD open that was unrelated to the instance I was working on.  Thus the on-call engineer was not aware of the new issue unless they knew to continually examine the already-open incident for more alerts in the details section.

PR #1197, opened 2023-08-10
ROX-18430: restore secrets from fleet-manager in fleetshard-sync

johannes94:
## Description

This PR depends on #1185, which stores secrets from data plane clusters in FM DB.

This PR includes the changes in fleet-manager private API to lazy load secrets stored in the FM DB. It also adds capabilities to fleetshard-sync reconciliation to load and restore stored secrets from fleet-manager if they are missing in the data plane cluster.

## Checklist (Definition of Done)

- [x] Unit and integration tests added
- [x] Added test description under `Test manual`
- [x] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)
- [x] CI and all relevant tests are passing
- [x] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [x] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.
- [x] Add secret to app-interface Vault or Secrets Manager if necessary
- ~~[ ] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)~~
- ~~[ ] Check AWS limits are reasonable for changes provisioning new resources~~

## Test manual

```
# Start an OSD cluster with infractl or using the ACSCS dev RH org
# Provision enough resources for an internal PostgreSQL DB at this point at least 1 node with 8 CPUs
# Install RHACS Operator on that cluster
# Configure AWS CLI to point to dev environment
# Create a fleet-manager cluster-config for the OSD cluster by editing /dev/config/dataplane-cluster-configuration-infractl-osd.yaml

make binary
make db/teardown db/setup db/migrate

# Start fleet-manager and fleetshard-sync in different shells
# Configure fleetshard-sync
export AWS_REGION=us-east-1
export SECRET_ENCRYPTION_TYPE=kms
export OCM_TOKEN=$(ocm token --refresh)
export AUTH_TYPE=OCM
export SECRET_ENCRYPTION_KEY_ID=<dev secret encryption key>
export CLUSTER_ID=<your-cluster-id>
./fleetshard-sync

# Run fleet-manager
./fleet-manager serve --dataplane-cluster-config-file "./dev/config/dataplane-cluster-configuration-infractl-osd.yaml"

# Create a central and make sure it get's to ready state without errors
./scripts/create-central.sh

# Store the hash of secret data to a tmp file
k get secret central-tls -n rhacs-cjaeg6v70jnii7c8gb4g -o yaml | yq .data | sha256sum > /tmp/central-tls-hash

[...]

PR #1207, opened 2023-08-17
ROX-18431: restore deleted tenant through API

johannes94:
## Description
This PR adds an API endpoint to the admin API of fleetmanager that allows to restore a deleted Central.

The restore operation resets the values of a deleted central in the database that are filled during the provisioning and deletion state machine flow. Then it resets it to provisioning and lets fleetshard-sync recreate the tenant. Additional logic in fleetshard-sync will search for RDS snapshots associated with cluster-id and restore the DB from a final snapshot if it exists.

The restore operation is currently limited, as it only restores the database to a state where the db password and master password needs to be restored. In order to get to a working tenant one has to follow: https://gitlab.cee.redhat.com/stackrox/acs-cloud-service/runbooks/-/blob/3c22117db0652ff52d54179038f625164d352f15/sops/dp-024-lost-tenant-db-password.md#L1

This PR will fix that: https://github.com/stackrox/acs-fleet-manager/pull/1197

- ~~[ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)~~ Docs will be updated in SOP repo.

## Test manual

Tested on an OSD cluster with all integrations (Route53, AWS RDS Database, RH SSO)
```
# Populate RH SSO dynamic client secrets, you can get them from app-interface vault of stage
# redhatsso-service.clientId and redhatsso-service.clientSecret files
# Populate Route53 secrets for dev, you can get them from OSCI vault
# aws.route53accesskey $aws.route53secretaccesskey
# Start fleetmanager with external DNS and your cluster config

make binary

./fleet-manager serve --enable-central-external-certificate --dataplane-cluster-config-file ./dev/config/dataplane-cluster-configuration-infractl-osd.yaml --force-leader --central-idp-client-id ""

# Start fleetshard-sync in another terminal
# Prepare environment
export CLUSTER_NAME=local_cluster                 
export MANAGED_DB_ENABLED
export MANAGED_DB_ENABLED=true
export AWS_AUTH_HELPER=aws-saml
export CREATE_AUTH_PROVIDER=true

# Start fleetshard-sync
./dev/env/scripts/exec_fleetshard_sync.sh

# In another terminal create an central instance
./scripts/create-central.sh

# Wait for central to be provisioned
# With above setup I included all integrations. Route53, RH SSO, RDS so it will take a while
# Once central is up login via Web Browser
# Create an API token in the integration tab of central
# We're going to use this token to ensure DB restore was successful later on so store it in a file
echo $token > token.txt

# Now delete the central tenant and wait for it's deprovisioning flow to complete
export central_id=<central-id>
export OCM_TOKEN=$(ocm token)
./scripts/fmcurl "rhacs/v1/centrals/$central_id?async=true" -XDELETE -v

# While waiting get an admin token
rhoas login --auth-url=https://auth.redhat.com/auth/realms/EmployeeIDP 
export OCM_TOKEN=$(rhoas authtoken)

# Once the deletion went through restore it through admin API
./scripts/fmcurl "rhacs/v1/admin/centrals/$central_id/restore" -XPOST -v

# Monitor restore operation, central will start in preparing and should go to ready. Through FM<>FS interaction.
# Once it's ready call the UI again and Login to test it's working
# Then as a final test use the API token
export ROX_API_TOKEN=(cat token.txt)
export ROX_ENDPOINT=acs-<central-id>.rhacs-dev.com:443
roxctl central whoami


```

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/central/cloudprovider/awsclient/rds.go
+++ b/fleetshard/pkg/central/cloudprovider/awsclient/rds.go
@@ -113,7 +113,7 @@ func (r *RDS) GetDBConnection(databaseID string) (postgres.DBConnection, error) 
 		var awsErr awserr.Error
 		if errors.As(err, &awsErr) {
 			if awsErr.Code() == rds.ErrCodeDBClusterNotFoundFault {
-				err = errors.Join(cloudprovider.ErrDBBackupInProgress, err)
+				err = errors.Join(cloudprovider.ErrDBNotFound, err)
 			}
 		}
 


fleetshard/pkg/central/cloudprovider/dbclient.go
```diff
@@ -16,7 +17,7 @@ type DBClient interface {
 	EnsureDBProvisioned(ctx context.Context, databaseID, acsInstanceID, passwordSecretName string, isTestInstance bool) error
 	// EnsureDBDeprovisioned is a non-blocking function that makes sure that a managed DB is deprovisioned (more
 	// specifically, that its deletion was initiated)
-	EnsureDBDeprovisioned(databaseID string, skipFinalSnapshot bool) error
+	EnsureDBDeprovisioned(databaseID string, deletedAt time.Time, skipFinalSnapshot bool) error
```
vladbologa, line 20: Not sure I understand, what is the reason for this change? The AWS API is already providing the snapshot creation time, is this just to document in the snapshot name when the deletion request was made?

vladbologa (changes requested):

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
```diff
@@ -154,7 +154,13 @@ func (r *RDS) ensureDBClusterCreated(clusterID, acsInstanceID, masterPassword st
 		return nil
 	}
 
+	finalSnapshotID, err := r.getFinalSnapshotIDIfExists(clusterID)
```
vladbologa, line 157: As a side note, this will not work if the default database ID is overridden, at least the way I started to implement this here: https://github.com/stackrox/acs-fleet-manager/pull/1198

In that PR the override is currently implemented as a ConfigMap in the data plane. Do you think the override name should actually be stored by FM?

johannes94: I was thinking about that. For now just assumed clusterID would be correct after #1198 in any case because I didn't know how you would handle that.

IIUC all overrides at some point should be done declaratively through GitOps. But we don't want to make the Backup/Restore epic depend on that.

One idea instead of configmap could be to use the tags you introduced to identified the clusterID. This would work as long as there is a DB running with appropriate tags. With this approach we wouldn't need a dedicated configmap to store the override, just some tags in AWS. But we would still have the problem that we loose overwrite information once a tenant was deleted. We wouldn't find a fitting final snapshot in that case.

Not sure if we're overthinking here. This is a edge case that is very unlikely to ever happen.

vladbologa: The idea with using the tags directly instead of the ConfigMap is worth considering either way, thanks!

I suppose we can discuss elsewhere whether this edge case should be handled automatically, or have some manual steps in an SOP. But it can definitely happen: once a DB override is in place, the restore deleted tenant operation will not work anymore.

fleetshard/pkg/central/cloudprovider/dbclient.go
vladbologa: Makes sense, thanks for the clarificaiton. I think I prefer the option where we don't change the interface, and use a uid like you suggested, to keep the interface cleaner - if you don't mind making this change.

fleetshard/pkg/central/cloudprovider/awsclient/rds.go
```diff
@@ -164,14 +170,52 @@ func (r *RDS) ensureDBClusterCreated(clusterID, acsInstanceID, masterPassword st
 		dataplaneClusterName: r.dataplaneClusterName,
 		isTestInstance:       isTestInstance,
 	}
-	_, err = r.rdsClient.CreateDBCluster(newCreateCentralDBClusterInput(input))
+
+	if finalSnapshotID != "" {
+		return r.restoreDBClusterFromSnapshot(finalSnapshotID, input)
```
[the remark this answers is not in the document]
vladbologa: Thanks!

fleetshard/pkg/central/cloudprovider/dbclient.go
johannes94: Did the change, let's see what CI says :)

fleetshard/pkg/central/cloudprovider/dbclient.go
```diff
@@ -14,10 +14,10 @@ import (
 type DBClient interface {
 	// EnsureDBProvisioned is a blocking function that makes sure that a database with the given databaseID was provisioned,
 	// using the master password given as parameter
-	EnsureDBProvisioned(ctx context.Context, databaseID, acsInstanceID, passwordSecretName string, isTestInstance bool) error
+	EnsureDBProvisioned(ctx context.Context, databaseID, acsInstanceID, password string, isTestInstance bool) error
```
johannes94, line 17: The variable name did not match to what was implemented. Implementation expect the DB password here.

PR #1214, opened 2023-08-23
Fix local dev env when using ACS 4.2 without OLM

vladbologa:
## Description

This PR fixes the ACS CS dev env (specifically the ability to reconcile a Central) for the following use-case:
* ACS version is 4.2.0
* the dev environment is set up with `make deploy/bootstrap` && `make deploy/dev`, and
* `INSTALL_OPERATOR=false`

In the scenario above, a locally run operator is failing to reconcile Centrals, with errors like
```
 namespace: \"openshift-monitoring\" from \"\": no matches for kind \"ServiceMonitor\" in version \"monitoring.coreos.com/v1\"\nensure CRDs are installed first]"
 ```
 
 This PR fixes this by installing the missing CRDs.

## Test manual

```
# Prepare local dev env (in the acs-fleet-manager dir)
INSTALL_OPERATOR=false make deploy/bootstrap
INSTALL_OPERATOR=false make deploy/dev

# Start a local operator (in the stackrox src dir)
make -C operator install run

# Create test central
./scripts/create-central.sh
```

The operator should be able to reconcile the Central, and the Central pod should start.

```
# To run tests locally run:
```

### DIFF
--- a/dev/env/scripts/bootstrap.sh
+++ b/dev/env/scripts/bootstrap.sh
@@ -54,6 +54,7 @@ if [[ "$INSTALL_OPERATOR" == "true" ]]; then
 else
     # We will be running without RHACS operator, but at least install our CRDs.
     apply "${MANIFESTS_DIR}/crds"
+    apply "${MANIFESTS_DIR}/monitoring"
 fi
 
 if [[ "$RHACS_STANDALONE_MODE" == "true" ]]; then
--- a/dev/env/manifests/monitoring/01-namespace.yaml
+++ b/dev/env/manifests/monitoring/01-namespace.yaml
@@ -1,0 +1,4 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: openshift-monitoring
--- a/dev/env/manifests/monitoring/02-monitoring.coreos.com_prometheusrules.yaml
+++ b/dev/env/manifests/monitoring/02-monitoring.coreos.com_prometheusrules.yaml
@@ -1,0 +1,129 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.11.1
+  creationTimestamp: null
+  name: prometheusrules.monitoring.coreos.com
+spec:
+  group: monitoring.coreos.com
+  names:
+    categories:
+      - prometheus-operator
+    kind: PrometheusRule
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: dev/env/scripts/bootstrap.sh [hunk 1]
```diff
@@ -54,6 +54,7 @@ if [[ "$INSTALL_OPERATOR" == "true" ]]; then
 else
     # We will be running without RHACS operator, but at least install our CRDs.
     apply "${MANIFESTS_DIR}/crds"
+    apply "${MANIFESTS_DIR}/monitoring"
```
vladbologa, line 57: I'm actually not sure if this should be done only when `INSTALL_OPERATOR != true`

I'm on arm64 and I have to build & run the operator locally, and can't test with `INSTALL_OPERATOR == true`

stehessel:
When you say local env, are you talking about an OpenShift cluster or plain Kubernetes? Because OpenShift should already have the CRD, and on plain Kubernetes the Helm chart should disable the monitoring resources based if `env.openshift=false`. What are the Helm values that the operator tries to reconcile?

vladbologa:
> When you say local env, are you talking about an OpenShift cluster or plain Kubernetes? Because OpenShift should already have the CRD, and on plain Kubernetes the Helm chart should disable the monitoring resources based if `env.openshift=false`. What are the Helm values that the operator tries to reconcile?

@stehessel It's plain kubernetes (i.e. k3s running in colima).

stehessel:
Is it possible that the operator tries to set `env.openshift=4` in the Helm chart even though it's installing to k3s? I'm asking because these lines should disable the monitoring resources if not on OpenShift 4 https://github.com/stackrox/stackrox/blob/master/image/templates/helm/stackrox-central/templates/_init.tpl.htpl#L118-L121.

stehessel:
Installing the CRDs works of course, but the better approach is probably to do something like https://github.com/stackrox/stackrox/blob/eee6f455f94e56b05ed048d9b857a941187cc966/operator/pkg/securedcluster/scanner/auto_sense.go#L63 and disable the monitoring resources explicitly if not running on OpenShift. Although I'm wondering why that is not already done via the Helm chart.

kurlov (approved):

vladbologa:
I'm merging this for now since it only affects a particular (and uncommon) ACS CS dev env configuration. If we'll support vanilla k8s in the operator we'll have to fix this properly.

PR #1224, opened 2023-08-28
fix(fleet-manager): add nil check to deleting mgr

stehessel:
## Description

Add a nil check because when `err != nil`, it is not guaranteed that `resp` is not nil. See https://issues.redhat.com/browse/ROX-19332 for an example of where we ran into a nil pointer dereference because of the missing check.

PS: Sadly the golang standard library does not provide a nil-safe getter for the status code :(


ivan-degtiarenko (commented):
Perhaps we should also apply the same principle to dynamic clients creation: https://github.com/stackrox/acs-fleet-manager/blob/89bf60f86189a74a25d3b6c4ee881785650ff223/internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go#L149-L159

stehessel:
There we don't access the `http.Response`, so we don't have the same problem. `dynamicClientData` is non-nil even on error.

dashrews78 (approved):
LGTM

PR #1233, opened 2023-09-04
ROX-19013 Add gitops to fleetmanager

ludydoo:
## Description

This PR applies the gitops configuration to centrals on the fleetmanager side

## Test manual

- Use a hardcoded gitops config
- Add an override patch to modify the central instance
- Verify that the patch is applied to Central


internal/dinosaur/pkg/gitops/config.go
```diff
@@ -14,8 +14,6 @@ type Config struct {
 
 // CentralsConfig represents the declarative configuration for Central instances defaults and overrides.
 type CentralsConfig struct {
-	// Default configuration for Central instances.
-	Default v1alpha1.Central `json:"default"`
```
ludydoo, line 18: Removed the `config.Default` because we decided in using a hardcoded default for centrals.

internal/dinosaur/pkg/gitops/config_test.go
```diff
@@ -24,9 +24,9 @@ func TestValidateGitOpsConfig(t *testing.T) {
 			},
 			yaml: `
 centrals:
-  default: {}
   overrides:
-  - instanceId: id1
```
ludydoo, line 29: Fixed a typo (`instanceId` -> `instanceIds`)

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -112,33 +108,29 @@ var _ DinosaurService = &dinosaurService{}
 type dinosaurService struct {
 	connectionFactory            *db.ConnectionFactory
 	clusterService               ClusterService
-	iamService                   sso.IAMService
 	dinosaurConfig               *config.CentralConfig
 	awsConfig                    *config.AWSConfig
 	quotaServiceFactory          QuotaServiceFactory
 	mu                           sync.Mutex
 	awsClientFactory             aws.ClientFactory
-	authService                  authorization.Authorization
```
ludydoo, line 121: unused dependency

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -641,21 +633,6 @@ func (k *dinosaurService) List(ctx context.Context, listArgs *services.ListArgum
 	return dinosaurRequestList, pagingMeta, nil
 }
 
-// ListByClusterID returns a list of CentralRequests with specified clusterID
```
ludydoo, line 644: moved to `dataPlaneCentralService`

internal/dinosaur/pkg/gitops/provider.go
```diff
@@ -0,0 +1,147 @@
...
+	"os"
+	"sync/atomic"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/yaml"
+)
+
+// ConfigProvider is the interface for GitOps configuration providers.
+type ConfigProvider interface {
```
SimonBaeumer, line 13: What was your reason to use decorators, and not for example a exchangeable ConfigReader?
Decorators to me mostly are used for extending functionality by wrapping an interface, like in your implementation, but here increasing the complexity of the stack by adding several wrappers.

Example of a config reader with an injected reader:
```
type ConfigProvider interfrace {
  Get() (config, error)
}

type ConfigProviderImpl struct {
  ConfigReader ConfigReader
  lastWorkingConfigCacheEnabled bool // If caching should be configurable
  prometheusMetricsEnabled bool // If metrics should be configurable   
}

func (c *ConfigProviderImpl) Get() (config, error) {
  
  if c.lastWorkingConfigCacheEnabled {
    // some caching
  }

  config, err := c.ConfigReader.Read()

  if c.prometheusMetricsEnabled {
   // some metrics 
  }
  return config, err
}
```

internal/dinosaur/pkg/gitops/provider.go
```diff
@@ -0,0 +1,147 @@
...
+// NewFallbackToLastWorkingConfigProvider returns a new fallbackToLastWorkingConfigProvider
+func NewFallbackToLastWorkingConfigProvider(primary ConfigProvider) ConfigProvider {
+	return &fallbackToLastWorkingConfigProvider{primary: primary}
+}
+
+// Get implements ConfigProvider.Get
+func (p *fallbackToLastWorkingConfigProvider) Get() (Config, error) {
+	config, err := p.primary.Get()
+	if err != nil {
+		return p.getLastWorkingConfig(err)
```
SimonBaeumer, line 102: Could you add a log message before returning that the new gitops config was not applied, instead the currently active one is used?

internal/dinosaur/pkg/gitops/provider.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+// -- fallbackToLastWorkingConfigProvider --
+
+// fallbackToLastWorkingConfigProvider is a ConfigProvider that returns the configuration from the primary
+// ConfigProvider, but if the primary ConfigProvider fails, it returns the last working configuration.
+type fallbackToLastWorkingConfigProvider struct {
+	primary           ConfigProvider
+	lastWorkingConfig atomic.Pointer[Config]
```
SimonBaeumer, line 90: What would happen if fleet-manager restarts when the `lastWorkingConfig` was active?
I am worried that fleet-manager cannot use any config because both configs are not applied or not available.

internal/dinosaur/pkg/gitops/provider.go
```diff
@@ -0,0 +1,147 @@
...
+	delegate ConfigProvider
+}
+
+// NewProviderWithMetrics returns a new providerWithMetrics
+func NewProviderWithMetrics(delegate ConfigProvider) ConfigProvider {
+	return &providerWithMetrics{delegate: delegate}
+}
+
+var (
+	errorCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
```
SimonBaeumer, line 130: Is the error counter reset automatically when I re-apply a valid config?

SimonBaeumer (changes requested):
Good job, thanks for splitting the PRs!

internal/dinosaur/pkg/gitops/provider.go
ludydoo: No, it would just stay at the same level. But the prom query would be something like `increase(counter_total[5m])` ...

internal/dinosaur/pkg/gitops/provider.go
ludydoo: To me it is much easier to test if the different cross cutting concerns (caching, observability) are encapsulated in these decorators

internal/dinosaur/pkg/gitops/provider.go
SimonBaeumer: We need to do something about it, the restart could cause service outages and applying default configurations to all Centrals.
I see two options:

- Reject reconciles, reconciliation is paused until gitops-config is fixed
- Persist last applied gitops-config in database to be persistent

My preferences goes towards the first option to return an error to fleetshard.

ludydoo: Yes the current logic would cause the reconcile would be rejected.

Here https://github.com/stackrox/acs-fleet-manager/blob/b961431de2238d5ad6629297dd6a6bac4f1895b2/internal/dinosaur/pkg/gitops/provider.go#L107

The handler would ultimately return an error. 
 
And here https://github.com/stackrox/acs-fleet-manager/blob/2d12faee6a6781243f2c995a8018d1c759d9ade1/fleetshard/pkg/runtime/runtime.go#L139

The error would reject that reconciliation

ludydoo: Additionally, see this test here

https://github.com/stackrox/acs-fleet-manager/blob/b961431de2238d5ad6629297dd6a6bac4f1895b2/internal/dinosaur/pkg/gitops/provider_test.go#L74

internal/dinosaur/pkg/gitops/provider.go
SimonBaeumer: @ludydoo The usage of decorators is off here. They are typically used to extend functions if the original code of a function should not be modified, i.e. when a plugin extends a core function of a library or framework.
In this case we have full control over the code and do not need this flexibility.
The disadvantages of using decorators here impacting the maintainability:

- While debugging the use of interfaces complicates identifying which implementation is used (similar challenges like the DI container we use today)
-  The construction of the object is weird here: https://github.com/stackrox/acs-fleet-manager/pull/1233/files#diff-2ec80f63d350d567e55163bd0d8e7d857e400aa38d56282e7f36ea7272338508R23-R30
The object has an implicit order in which it must be constructed (decorators wrap core logic) which is not obvious to the reader. 
 - Reading the code is not possible in its execution flow without overhead. I needed to create a mental map of the ordering of the wrapped instances, and then constructing the stack to understand how the code is called. This construction can change in different contexts (i.e. tests) where I need to re-map the ordering of decorators.

Decorators are a powerful but come at a high price of abstractions. A simple reader interface which is injected into the ConfigProvider would do the same job with less abstraction.
The cross-cutting concerns could still be tested by using an empty reader (or any other reader implementation).

internal/dinosaur/pkg/gitops/provider.go
SimonBaeumer: Thanks for explaining, this makes sense!

internal/dinosaur/pkg/gitops/provider.go
SimonBaeumer: In that case the counter would provide wrong information when the counter increase continuously, reporting 100 errors while there are 0 errors. It creates an implicit requirement on the writer of a prom query to get usable data.
Is it possible to decrease the counter when the error was reset, basically having a boolean value?
In doubt consulting SRE how to handle this metric.
caveat: I am not very familiar with Prometheus metrics, if this is considered best-practices can you share an article/doc with me?

SimonBaeumer (changes requested):
Summary: The ConfigProvider implementation needs a bit of refactoring to be less abstract and Prometheus counter looks like it provides wrong data.

internal/dinosaur/pkg/gitops/provider.go
ludydoo: That's how counters are supposed to work. It's not a problem for querying it with PromQL

https://prometheus.io/docs/concepts/metric_types/

https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Counter

internal/dinosaur/pkg/gitops/README.md
```diff
@@ -0,0 +1,14 @@
+# GitOps Workflow
```
kurlov, line 1: There is a pretty nice doc and schema 🚀

internal/dinosaur/pkg/gitops/service.go
```diff
@@ -0,0 +1,214 @@
+package gitops
+
+import (
+	"encoding/json"
+	"strings"
+	"text/template"
+
+	"github.com/ghodss/yaml"
```
kurlov, line 8: Was it intentional to use this lib?

internal/dinosaur/pkg/gitops/service.go
```diff
@@ -0,0 +1,214 @@
...
+	if err := json.Unmarshal(centralBytes, &result); err != nil {
+		return v1alpha1.Central{}, errors.Wrap(err, "failed to unmarshal Central instance")
+	}
+	return result, nil
+}
+
+// shouldApplyCentralOverride returns true if the given Central override should be applied to the given Central instance.
+func shouldApplyCentralOverride(override CentralOverride, ctx CentralParams) bool {
+	for _, d := range override.InstanceIDs {
+		if d == "*" {
```
kurlov, line 122: When instance ID is equal to `*`?

internal/dinosaur/pkg/gitops/service.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	var writer = new(strings.Builder)
+	if err := tpl.Execute(writer, ctx); err != nil {
+		return "", errors.Wrap(err, "failed to render patch template")
+	}
+	return writer.String(), nil
+}
+
+// defaultTemplate is the default template for Central instances.
+var defaultTemplate = template.Must(template.New("default").Parse(`
```
kurlov, line 160: Let's put it in .yaml file and read from it. Just to make it a little easier to tune later

internal/dinosaur/pkg/gitops/provider.go
SimonBaeumer: A gaudge could do that: https://prometheus.io/docs/concepts/metric_types/#gauge :thinking:

ludydoo: You could, but that's not standard. Check the usages in the repository for counters, they are canonically used to represent the number of errors, the number of requests, etc. e.g. `ReconcilerFailureCount`.

Then, we can craft a query like `increase` to watch the rate at which this number increases. https://prometheus.io/docs/prometheus/latest/querying/functions/#increase

> calculates the increase in the time series in the range vector. Breaks in monotonicity (such as counter resets due to target restarts) are automatically adjusted for

internal/dinosaur/pkg/gitops/service.go
ludydoo: Means that this override will apply to all centrals

internal/dinosaur/pkg/gitops/provider.go
SimonBaeumer: Hm, it still delegates the complexity of the implementation to the writer of the prom query. 
We need an indication that an error happened in the GitOps configuration as feedback for the engineer who has merged a pull-request, we are kind of abusing the alerting system a bit here.
This would make the gauge the better alternative and resulting in a promquery which is simple and idiomatic to use. 
A user of the metric does not know how to work with this data without additional documentation how to interpret it.

Using a counter overall increases the system's complexity in this use-case, making the gauge the simpler, respectively better alternative.

ludydoo: It's incorrect. Gauges are used for metrics that can decrease, such as memory usage, cpu usage, number of connected sensors. Counters are used to measure increasing values, such as the number of errors thrown, the number of requests handled. 

The only valid way to use a gauge in this context would be to calculate the "number of successive gitops errors" - as opposed to the total number of gitops errors. This has little value in this context. 

Using counters to report errors through prometheus is a standard use of this type of metric. 

See 

- https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/stats
- https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/tcp_proxy_filter
- https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats
- https://kubernetes.io/docs/reference/instrumentation/metrics/

internal/dinosaur/pkg/gitops/provider.go
ludydoo: Agreed. Changed the implementation to use a `Reader` and a `Provider` as suggested

kurlov (approved):
🎉

internal/dinosaur/pkg/gitops/provider.go
SimonBaeumer: @ludydoo Thanks for the great explanation, it makes sense to me now. I agree, let's keep the `counter` and align with best pratices as you suggested.

internal/dinosaur/pkg/gitops/provider_test.go
```diff
@@ -0,0 +1,140 @@
...
+				p.reader = &failingReader
+			} else {
+				p.reader = &successfulReader
+			}
+
+			if tc.validationWillFail {
+				p.validationFn = failingValidationFn
+			} else {
+				p.validationFn = successfulValidationFn
+			}
```
SimonBaeumer, line 97: nit: the validationFns and readers can be part of the configured test directly, would remove the if-statements

internal/dinosaur/pkg/gitops/reader_test.go
```diff
@@ -0,0 +1,50 @@
...
+func TestFileReader_Get_FailsIfFileIsInvalidYAML(t *testing.T) {
+	tmpDir := t.TempDir()
+	tmpFilePath := tmpDir + "/config.yaml"
+	file, err := os.Create(tmpFilePath)
+	require.NoError(t, err)
+	_, err = file.WriteString("invalid yaml")
+	require.NoError(t, err)
+	provider := NewFileReader(tmpFilePath)
+	_, err = provider.Read()
+	assert.Error(t, err)
```
SimonBaeumer, line 28: nit: require.NoError

SimonBaeumer (approved):
LGTM :ship:

internal/dinosaur/pkg/gitops/reader_test.go
ludydoo: This one expects an error (`TestFileReader_Get_FailsIfFileIsInvalidYAML`) - if the yaml is invalid

PR #1234, opened 2023-09-04
ROX-19013: Add gitops to fleetshard

ludydoo:
## Description
Uses the `centralYAML` from the gitops config to reconcile the centrals. 

## Test manual

- Use a hardcoded gitops config
- Add an override patch to modify the central instance
- Verify that the patch is applied to Centraly lint binary test test/integration

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -255,14 +255,12 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 func (r *CentralReconciler) getInstanceConfig(remoteCentral *private.ManagedCentral) (*v1alpha1.Central, error) {
 	var central *v1alpha1.Central
 	var err error
-	if features.GitOpsCentrals.Enabled() {
-		central, err = r.getInstanceConfigWithGitops(remoteCentral)
-		if err != nil {
+	if r.shouldUseGitopsConfig(remoteCentral) {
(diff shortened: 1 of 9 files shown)


SimonBaeumer (changes requested):
Please add an automated test for the gitops config when it is enabled.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -252,6 +253,34 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 }
 
 func (r *CentralReconciler) getInstanceConfig(remoteCentral *private.ManagedCentral) (*v1alpha1.Central, error) {
+	var central *v1alpha1.Central
+	var err error
+	if features.GitOpsCentrals.Enabled() {
```
SimonBaeumer, line 258: In case GitOps config is enabled, but the response is empty it should fallback on the legacy version.
With the following release we can remove the backward-compatibility switch. By doing that we support older fleet-manager verisons in case of a rollback.

SimonBaeumer (approved):

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: When `getInstanceConfigWithGitops` returns an error would be more accurate, empty could a valid state

ludydoo: Resolving the conversation after private meeting

PR #1236, opened 2023-09-04
ROX-11640: RHSSO dynamic clients rotation API

ivan-degtiarenko:
## Description
This PR creates an admin API that:
* creates a new client/secret pair via the dynamic client sso.r.c API.
* delete the previous dynamic client.
* updates the client/secret pair in fleet manager DB.

In case of failure to either use new client or delete the previous one, client ID is logged that allows for ACSCS engineer to manually delete stale client.
In case dynamic clients are not configured, `404` status code is returned.

The dynamic clients API client is initialized once, however, it might not be initialized properly if dynamic clients are not configured. Whether dynamic clients are configured is checked on every request.

## Test manual

(You need to set up `SSO_CLIENT_ID_DEFAULT` and `SSO_CLIENT_SECRET_DEFAULT`, as well as specify `--central-idp-client-id` as empty string)

1. `make deploy/bootstrap`
2. `make deploy/dev`
3. Create central
4. Look into `cloud-service-sensible-declarative-configs` secret and see client ID and secret values
5. Call `/rotate-secrets` endpoint
6. See that client ID and secret have changed within `cloud-service-sensible-declarative-configs`

### DIFF
--- a/internal/dinosaur/pkg/handlers/admin_dinosaur.go
+++ b/internal/dinosaur/pkg/handlers/admin_dinosaur.go
@@ -44,6 +44,8 @@ type AdminCentralHandler interface {
 	GetCentralDefaultVersion(w http.ResponseWriter, r *http.Request)
 	// Restore restores a tenant that was already marked as deleted
 	Restore(w http.ResponseWriter, r *http.Request)
+	// RotateSecrets rotates secrets within central
+	RotateSecrets(writer http.ResponseWriter, request *http.Request)
 }
 
 type adminCentralHandler struct {
@@ -427,6 +429,22 @@ func (h adminCentralHandler) GetCentralDefaultVersion(w http.ResponseWriter, r *
 	handlers.Handle(w, r, cfg, http.StatusOK)
 }
 
+func (h adminCentralHandler) RotateSecrets(w http.ResponseWriter, r *http.Request) {
+	cfg := &handlers.HandlerConfig{
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+			id := mux.Vars(r)["id"]
+			ctx := r.Context()
+			centralRequest, err := h.service.Get(ctx, id)
+			if err != nil {
+				return nil, err
+			}
+			svcErr := h.service.RotateCentralRHSSOClient(ctx, centralRequest)
+			return nil, svcErr
+		},
+	}
+	handlers.Handle(w, r, cfg, http.StatusOK)
+}
+
 type gitOpsAdminHandler struct{}
 
 var _ AdminCentralHandler = (*gitOpsAdminHandler)(nil)
@@ -464,5 +482,9 @@ func (g gitOpsAdminHandler) GetCentralDefaultVersion(w http.ResponseWriter, r *h
 }
 
 func (g gitOpsAdminHandler) Restore(w http.ResponseWriter, r *http.Request) {
+	http.Error(w, "not implemented", http.StatusNotImplemented)
+}
+
+func (g gitOpsAdminHandler) RotateSecrets(w http.ResponseWriter, r *http.Request) {
 	http.Error(w, "not implemented", http.StatusNotImplemented)
 }
--- a/internal/dinosaur/pkg/handlers/admin_dinosaur.go
+++ b/internal/dinosaur/pkg/handlers/admin_dinosaur.go
@@ -45,7 +45,7 @@ type AdminCentralHandler interface {
 	// Restore restores a tenant that was already marked as deleted
 	Restore(w http.ResponseWriter, r *http.Request)
 	// RotateSecrets rotates secrets within central
-	RotateSecrets(writer http.ResponseWriter, request *http.Request)
+	RotateSecrets(w http.ResponseWriter, r *http.Request)
 }
 
 type adminCentralHandler struct {
@@ -434,12 +434,12 @@ func (h adminCentralHandler) RotateSecrets(w http.ResponseWriter, r *http.Reques
 		Action: func() (i interface{}, serviceError *errors.ServiceError) {
 			id := mux.Vars(r)["id"]
 			ctx := r.Context()
-			centralRequest, svcErr := h.service.Get(ctx, id)
-			if svcErr != nil {
-				return nil, svcErr
+			centralRequest, err := h.service.Get(ctx, id)
+			if err != nil {
+				return nil, err
 			}
-			svcErr = h.service.RotateCentralRHSSOClient(ctx, centralRequest)
-			return nil, svcErr
+			err = h.service.RotateCentralRHSSOClient(ctx, centralRequest)
+			return nil, err
 		},
 	}
 	handlers.Handle(w, r, cfg, http.StatusOK)
--- a/internal/dinosaur/pkg/handlers/admin_dinosaur.go
+++ b/internal/dinosaur/pkg/handlers/admin_dinosaur.go
@@ -434,11 +434,11 @@ func (h adminCentralHandler) RotateSecrets(w http.ResponseWriter, r *http.Reques
 		Action: func() (i interface{}, serviceError *errors.ServiceError) {
 			id := mux.Vars(r)["id"]
 			ctx := r.Context()
-			centralRequest, err := h.service.Get(ctx, id)
-			if err != nil {
-				return nil, err
+			centralRequest, svcErr := h.service.Get(ctx, id)
+			if svcErr != nil {
+				return nil, svcErr
 			}
-			svcErr := h.service.RotateCentralRHSSOClient(ctx, centralRequest)
+			svcErr = h.service.RotateCentralRHSSOClient(ctx, centralRequest)
 			return nil, svcErr
 		},
 	}
--- a/internal/dinosaur/pkg/rhsso/augment.go
+++ b/internal/dinosaur/pkg/rhsso/augment.go
@@ -1,0 +1,38 @@
+package rhsso
+
+import (
+	"context"
+	"fmt"
+	"github.com/pkg/errors"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/dbapi"
+	"github.com/stackrox/acs-fleet-manager/pkg/client/iam"
+	"github.com/stackrox/acs-fleet-manager/pkg/client/redhatsso/api"
+	"github.com/stackrox/rox/pkg/stringutils"
+)
+
+const (
+	oidcProviderCallbackPath    = "/sso/providers/oidc/callback"
+	dynamicClientsNameMaxLength = 50
+)
+
+func AugmentWithDynamicAuthConfig(ctx context.Context, r *dbapi.CentralRequest, realmConfig *iam.IAMRealmConfig, apiClient *api.AcsTenantsApiService) error {
+	// There is a limit on name length of the dynamic client. To avoid unnecessary errors,
+	// we truncate name here.
+	name := stringutils.Truncate(fmt.Sprintf("acsms-%s", r.Name), dynamicClientsNameMaxLength)
+	orgID := r.OrganisationID
+	redirectURIs := []string{fmt.Sprintf("https://%s%s", r.GetUIHost(), oidcProviderCallbackPath)}
+
+	dynamicClientData, _, err := apiClient.CreateAcsClient(ctx, api.AcsClientRequestData{
+		Name:         name,
+		OrgId:        orgID,
+		RedirectUris: redirectURIs,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "failed to create RHSSO dynamic client for %s", r.ID)
+	}
+
+	r.AuthConfig.ClientID = dynamicClientData.ClientId
+	r.AuthConfig.ClientSecret = dynamicClientData.Secret // pragma: allowlist secret
+	r.AuthConfig.Issuer = realmConfig.ValidIssuerURI
+	return nil
+}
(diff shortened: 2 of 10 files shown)


internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -415,6 +417,22 @@ func (h adminCentralHandler) GetCentralDefaultVersion(w http.ResponseWriter, r *
 	handlers.Handle(w, r, cfg, http.StatusOK)
 }
 
+func (h adminCentralHandler) RotateSecrets(w http.ResponseWriter, r *http.Request) {
```
johannes94, line 420: The name of this handler could be ambiguous. We also store other secrets within a central request of FM DB which are not rotated by this operation. I'd suggest to make it more specific like you did in the service, something like `RotateCentralRHSSOClient'.

internal/dinosaur/pkg/routes/route_loader.go
```diff
@@ -256,6 +256,9 @@ func (s *options) buildAPIBaseRouter(mainRouter *mux.Router, basePath string, op
 	adminCentralsRouter.HandleFunc("/{id}", adminCentralHandler.Update).
 		Name(logger.NewLogEvent("admin-update-central", "[admin] update central by id").ToString()).
 		Methods(http.MethodPatch)
+	adminCentralsRouter.HandleFunc("/{id}/rotate-secrets", adminCentralHandler.RotateSecrets).
```
johannes94, line 259: See comment about the handler name, the route should also reflect that.

internal/dinosaur/pkg/routes/route_loader.go
```diff
@@ -256,6 +256,9 @@ func (s *options) buildAPIBaseRouter(mainRouter *mux.Router, basePath string, op
...
+		Name(logger.NewLogEvent("admin-rotate-central-secrets", "[admin] rotate central secrets by id").ToString()).
+		Methods(http.MethodPatch)
```
johannes94, line 261: I'd prefer http.MethodPost at this point. For me `PATCH` indicates that I partially update the resource with the resource body I send. Which doesn't match to what the endpoint is doing.

johannes94 (changes requested):
Please also update the API [documentation](./openapi/fleet-manager-private-admin.yaml).

internal/dinosaur/pkg/routes/route_loader.go
ivan-degtiarenko: I think `POST` does not apply here because `POST` indicates the creation of a new resource, while `PATCH` indicates partial modification of an existing resource. 

I agree with you that having `PATCH` without a body is weird as well(although If I read correctly does not go against [RFC](https://www.rfc-editor.org/rfc/rfc5789#section-2). IMHO out of all 3 potential methods(`POST`, `PUT`, `PATCH`) I think `PATCH` comes the closest to the standard.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
ivan-degtiarenko: My original idea was that we might want to extend this endpoint with more secret rotation in the future, however, there's no indication it will ever be the case. I'll rename the handler 👍

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -415,6 +417,22 @@ func (h adminCentralHandler) GetCentralDefaultVersion(w http.ResponseWriter, r *
...
+	cfg := &handlers.HandlerConfig{
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+			id := mux.Vars(r)["id"]
+			ctx := r.Context()
+			centralRequest, svcErr := h.service.Get(ctx, id)
```
dhaus67, line 425: Let's keep it simple and refer to it as `err`, there's no shadowing here which would require another name for the error.

internal/dinosaur/pkg/rhsso/augment.go
```diff
@@ -0,0 +1,38 @@
...
+
+const (
+	oidcProviderCallbackPath    = "/sso/providers/oidc/callback"
+	dynamicClientsNameMaxLength = 50
+)
+
+func AugmentWithDynamicAuthConfig(ctx context.Context, r *dbapi.CentralRequest, realmConfig *iam.IAMRealmConfig, apiClient *api.AcsTenantsApiService) error {
+	// There is a limit on name length of the dynamic client. To avoid unnecessary errors,
+	// we truncate name here.
+	name := stringutils.Truncate(fmt.Sprintf("acsms-%s", r.Name), dynamicClientsNameMaxLength)
```
dhaus67, line 21: With the other name change PR you had, shall we also change the name to `acscs` here?

Shouldn't really make a difference since they are anways referred to by client ID.

internal/dinosaur/pkg/routes/route_loader.go
dhaus67: What we're doing here isn't really REST since it's not necessarily related to `CRUD` (you don't really create a resource, or update it per-se, but rather trigger an action).

Looking at the overall industry, `POST` seems to be the common standard to use when triggering actions like revoking things, let's use it instead of `PATCH`.

Both are also open for extending it, like we said in case we want to rotate _more_ secrets than just the sso.r.c dynamic client.

internal/dinosaur/pkg/routes/route_loader.go
dhaus67: Hmm, I think we have two ways to go here:
- Either keep this one "generic" and leave it open for extending it.
- Create "RPC-style" API which has options like `rotate/dynamic-client`, `rotate/dns` or whatever else we might want to rotate for an existing instance in the future.

Having the extensibility in mind, I wouldn't mind keeping it "unspecific" and have the option to extend it further down the road.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -105,6 +109,7 @@ type DinosaurService interface {
 	ListDinosaursWithRoutesNotCreated() ([]*dbapi.CentralRequest, *errors.ServiceError)
 	ListCentralsWithoutAuthConfig() ([]*dbapi.CentralRequest, *errors.ServiceError)
 	VerifyAndUpdateDinosaurAdmin(ctx context.Context, dinosaurRequest *dbapi.CentralRequest) *errors.ServiceError
+	RotateCentralRHSSOClient(ctx context.Context, centralRequest *dbapi.CentralRequest) *errors.ServiceError
```
dhaus67, line 112: Maybe as a follow-up, let's use this opportunity to get rid of the `dinosaurRequest` naming within this file.

pkg/errors/errors.go
```diff
@@ -128,6 +128,14 @@ const (
 	ErrorMaxLimitForServiceAccountsReached       ServiceErrorCode = 115
 	ErrorMaxLimitForServiceAccountsReachedReason string           = "Max limit for the service account creation has reached"
 
+	// RHSSO client rotation is not configured for this particular fleet-manager instance
+	ErrorClientRotationNotConfigured       ServiceErrorCode = 116
+	ErrorClientRotationNotConfiguredReason string           = "RHSSO client rotation is not configured"
```
dhaus67, line 133: Is it really `ClientRotationNotConfigured` or rather `DynamicClientNotConfigured`?

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -143,9 +151,35 @@ func NewDinosaurService(connectionFactory *db.ConnectionFactory, clusterService
...
+		return errors.New(errors.ErrorClientRotationNotConfigured, "RHSSO dynamic configuration is not present")
+	}
+
+	previousAuthConfig := centralRequest.AuthConfig
+	if err := rhsso.AugmentWithDynamicAuthConfig(ctx, centralRequest, k.iamConfig.RedhatSSORealm, k.rhSSODynamicClientsAPI); err != nil {
+		return errors.NewWithCause(errors.ErrorClientRotationFailed, err, "failed to augment auth config")
+
+	}
+	if err := k.Update(centralRequest); err != nil {
+		glog.Errorf("Created new RHSSO dynamic client, but failed to update central record, client ID is %s", centralRequest.AuthConfig.ClientID)
```
dhaus67, line 173: Maybe use `rotated` here so it's easier searchable?

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -143,9 +151,35 @@ func NewDinosaurService(connectionFactory *db.ConnectionFactory, clusterService
...
+		return errors.NewWithCause(errors.ErrorClientRotationFailed, err, "failed to update database record")
+	}
+	if _, err := k.rhSSODynamicClientsAPI.DeleteAcsClient(ctx, previousAuthConfig.ClientID); err != nil {
+		glog.Errorf("Failed to delete RHSSO dynamic client, client ID is %s", centralRequest.AuthConfig.ClientID)
+		return errors.NewWithCause(errors.ErrorClientRotationFailed, err, "failed to delete previous OIDC client")
```
dhaus67, line 178:
```suggestion
		return errors.NewWithCause(errors.ErrorClientRotationFailed, err, "failed to delete previous RHSSO dynamic client")
```

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -143,9 +151,35 @@ func NewDinosaurService(connectionFactory *db.ConnectionFactory, clusterService
...
+	if err := rhsso.AugmentWithDynamicAuthConfig(ctx, centralRequest, k.iamConfig.RedhatSSORealm, k.rhSSODynamicClientsAPI); err != nil {
+		return errors.NewWithCause(errors.ErrorClientRotationFailed, err, "failed to augment auth config")
+
+	}
+	if err := k.Update(centralRequest); err != nil {
+		glog.Errorf("Created new RHSSO dynamic client, but failed to update central record, client ID is %s", centralRequest.AuthConfig.ClientID)
+		return errors.NewWithCause(errors.ErrorClientRotationFailed, err, "failed to update database record")
+	}
+	if _, err := k.rhSSODynamicClientsAPI.DeleteAcsClient(ctx, previousAuthConfig.ClientID); err != nil {
+		glog.Errorf("Failed to delete RHSSO dynamic client, client ID is %s", centralRequest.AuthConfig.ClientID)
```
dhaus67, line 177: (not related to this PR): wonder whether it'd be better to create metrics for this and alerts?

Feels like this is definitely something we want to have and persist it, would also help make life easier in terms of a) discovering it and b) investigating it.

But, probably more of a follow-up thing instead of this PR.

pkg/errors/errors.go
ivan-degtiarenko: It's possible for client rotation not to be configured but dynamic clients to be configured, because in code we give preference to static configuration if it's configured

internal/dinosaur/pkg/handlers/admin_dinosaur.go
johannes94: No need to rename, see comment below.

internal/dinosaur/pkg/routes/route_loader.go
ivan-degtiarenko: `POST` it is

internal/dinosaur/pkg/services/dinosaur.go
ivan-degtiarenko: I'll create a follow-up PR 👍

internal/dinosaur/pkg/services/dinosaur.go
ivan-degtiarenko: Reworked the error message

pkg/errors/errors.go
dhaus67: Should we call it `DynamicClientNotUsed` then? My point is we don't have a configuration for client rotation, and if anyone other than the two of us reads the error message they might be confused as to what is going on / why it isn't working; and which configuration option to set 😅

internal/dinosaur/pkg/services/dinosaur.go
ivan-degtiarenko: For me, the path for discovery and investigation is pretty straightforward:
1. Discovery - ACSCS operator sends request to rotate secrets, gets status code other than `2xx`
2. Investigation - ACSCS operator goes into fleet-manager logs and finds the reason for that failure

Metrics and alerts I think would be needed in case we decide to automate client rotation in the future.

dhaus67: Yeah fair, my thinking is more or less that we extend this to _all_ delete operations and also cover stale clients during instance cleanup (which isn't really something we know currently _unless_ there's an issue, and stuff is quite hard to track back).

For the API, I agree it's pretty straight forward and not required, so feel free to skip this comment for now.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -144,9 +152,35 @@ func NewDinosaurService(connectionFactory *db.ConnectionFactory, clusterService
...
 	}
 }
 
+func (k *dinosaurService) RotateCentralRHSSOClient(ctx context.Context, centralRequest *dbapi.CentralRequest) *errors.ServiceError {
+	realmConfig := k.iamConfig.RedhatSSORealm
+	if k.dinosaurConfig.HasStaticAuth() {
+		return errors.New(errors.ErrorDynamicClientsNotUsed, "RHSSO is configured via static configuration")
+	}
+	if !realmConfig.IsConfigured() {
+		return errors.New(errors.ErrorDynamicClientsNotUsed, "RHSSO dynamic configuration is not present")
```
dhaus67, line 165:
```suggestion
		return errors.New(errors.ErrorDynamicClientsNotUsed, "RHSSO dynamic client configuration is not present")
```

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_auth_config_mgr.go
```diff
@@ -2,12 +2,10 @@ package dinosaurmgrs
 
 import (
 	"context"
-	"fmt"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/rhsso"
```
dhaus67, line 5: Let's format this properly and put it below to the non-standard packages.

dhaus67 (approved):

johannes94 (approved):
LGTM

PR #1240, opened 2023-09-05
fix: use "ams" quota type in e2e tests

0x656b694d:
As per the [quota.md](https://github.com/stackrox/acs-fleet-manager/blob/e35d6ad32823f5f19bbdd1f2d19b36e672f32be1/docs/quota/quota.md):

There are currently two quota backends -
[Quota Management with Account Management Service (AMS)](https://gitlab.cee.redhat.com/service/uhc-account-manager)
and quota management list (default). The CLI flag `--quota-type={ams,quota-management-list}` configures the
quota backend at deploy time of fleet manager. Production instances of fleet manager are configured with
`--quota-type=ams` to verify quota against customer's SKU entitlements.

PR #1248, opened 2023-09-08
ROX-19507: Register grace period start date

0x656b694d:
Add `expired_at` field to the central request.
Set it to current time when there's no more quota entitlement.

Next PR #1250 introduces a procedure to set instances state to _deprovision_ state with completed grace period, which is 2 weeks since `expired_at`.


0x656b694d:

  * main
    * **PR #1240**
    * **PR #1248** 👈

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -213,3 +213,37 @@ func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError
...
+
+	quotaType := types.DinosaurInstanceType(central.InstanceType).GetQuotaType()
+	quotaCosts, err := q.amsClient.GetQuotaCostsForProduct(org.ID(), quotaType.GetResourceName(), quotaType.GetProduct())
+	if err != nil {
+		return false, errors.InsufficientQuotaError("%v: error getting quotas for product %s", err, quotaType.GetProduct())
+	}
+
+	for _, qc := range quotaCosts {
+		for _, rr := range qc.RelatedResources() {
+			// When an SKU entitlement expires in AMS, the allowed value for that quota cost is set back to 0.
```
stehessel, line 232: I'm afraid we have to go down to the cloud account ID level here. So if the billing model is one of the cloud marketplaces, we need `Consumed() <= Allowed() && cloud_account_id in cloud_accounts`. Where `cloud_accounts` is the list of of active cloud accounts returned by `/quota_cost?fetchCloudAccounts=true`.

0x656b694d: Are you saying that quota cost can have `allowed == 0`, but still have entitlement if cloud account is active?

0x656b694d: Or, is it that we also need to check if the account is active in addition to `allowed == 0`?

stehessel: No what I'm saying is that an instance can be expired even if `allowed > 0`, because `cloud_account_id` is no longer part of `cloud_accounts`. This happens if the cloud account that initially registered the instance has unsubscribed, but another account still provides quota.

stehessel: Basically we need:
* For `billing_model=standard`: `consumed <= allowed`
* For `billing_model=marketplace_*`: `consumed <= allowed && cloud_account_id in cloud_accounts`.

`allowed==0` would fail the first condition if any instance is still alive.

0x656b694d: Kafka thinks quota is entitled even if overconsumed. So the reaction could be different between `allowed=0 && consumed>0` and `allowed==1 && consumed==2`.

internal/dinosaur/pkg/services/quota/quota_management_list_service.go
```diff
@@ -101,3 +102,26 @@ func (q QuotaManagementListService) ReserveQuota(dinosaur *dbapi.CentralRequest,
 func (q QuotaManagementListService) DeleteQuota(SubscriptionID string) *errors.ServiceError {
 	return nil // NOOP
 }
+
+// Checks if quota used by the given Central instance is granted to the organisation/user and
+// if it is active, not expired, in the quota management list configuration
+// Note that organisation will always take priority over individual accounts to mimic the behaviour of
+// quota allowance checks during Central creation.
+func (q QuotaManagementListService) IsQuotaEntitlementActive(central *dbapi.CentralRequest) (bool, error) {
+
```
ivan-degtiarenko, line 111: empty line

internal/dinosaur/pkg/services/quota/ams_quota_service.go
0x656b694d: Please take a look at the updated logic:
* it maps the allowed quota costs to cloud and standard models, for convenience;
* for cloud accounts: checks it is present in the list of quota cloud accounts;
* for standard accounts: checks if there is some allowed quota in the list.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,127 @@
...
+		glog.Infof("reconciling grace_from for central instance %q", central.ID)
+
+		// skip update when Central is marked for deletion or is already being deleted
+		if arrays.Contains(constants.GetDeletingStatuses(), central.Status) {
+			glog.Infof("central %q is in %q state, skipping grace_from reconciliation", central.ID, central.Status)
+			continue
+		}
+
+		glog.Infof("checking quota entitlement status for Central instance %q", central.ID)
+		active, ok := subscriptionStatusByOrg[central.OrganisationID]
```
ivan-degtiarenko, line 80: let's rename `ok` to `exists` or `isPresent`

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,127 @@
...
+func (k *GracePeriodManager) reconcileCentralGraceFrom(centrals dbapi.CentralList) serviceErr.ErrorList {
+	glog.Infof("reconciling grace period start date for central instances")
+	var svcErrors serviceErr.ErrorList
+	subscriptionStatusByOrg := map[string]bool{}
+
+	for _, central := range centrals {
+		glog.Infof("reconciling grace_from for central instance %q", central.ID)
+
+		// skip update when Central is marked for deletion or is already being deleted
+		if arrays.Contains(constants.GetDeletingStatuses(), central.Status) {
```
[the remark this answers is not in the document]
0x656b694d: Indeed.

internal/dinosaur/constants/central.go
```diff
@@ -100,3 +100,7 @@ func GetUpdateableStatuses() []string {
 		CentralRequestStatusReady.String(),
 	}
 }
+
+func GetDeletingStatuses() []string {
```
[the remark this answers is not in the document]
0x656b694d: Requires future refactoring.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -943,3 +962,12 @@ func buildResourceRecordChange(recordName string, clusterIngress string, action
 
 	return resourceRecordChange
 }
+
+func (k *dinosaurService) IsQuotaEntitlementActive(centralRequest *dbapi.CentralRequest) (bool, error) {
```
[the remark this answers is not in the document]
0x656b694d: Thanks. Moved the code to the `grace_period_mgr`.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,138 @@
...
+func (k *GracePeriodManager) Stop() {
+	k.StopWorker(k)
+}
+
+// Reconcile ...
+func (k *GracePeriodManager) Reconcile() []error {
+	glog.Infoln("reconciling centrals")
+	var encounteredErrors []error
+
+	centrals, svcErr := k.dinosaurService.ListByStatus(constants.GetDeletingStatuses()...)
```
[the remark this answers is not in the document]
0x656b694d: Brilliant! Fixed.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -213,3 +213,75 @@ func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError
...
+				account.CloudProviderID() == central.CloudProvider {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// IsQuotaEntitlementActive checks if an organisation has a SKU entitlement and that entitlement is still active in AMS.
+func (q amsQuotaService) IsQuotaEntitlementActive(central *dbapi.CentralRequest) (bool, error) {
```
ivan-degtiarenko, line 256:
```suggestion
func (q *amsQuotaService) IsQuotaEntitlementActive(central *dbapi.CentralRequest) (bool, error) {
```

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -213,3 +213,75 @@ func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError
...
+		return false, errors.InsufficientQuotaError("%v: error getting quotas for product %q", err, quotaType.GetProduct())
+	}
+
+	quotasMap, err := mapAllowedQuotaCosts(quotaCosts)
+	if err != nil {
+		svcErr := errors.ToServiceError(err)
+		return false, errors.NewWithCause(svcErr.Code, svcErr, "product %q has no allowed billing models", quotaType.GetProduct())
+	}
+
+	isCloudAccount := central.CloudAccountID != "" && central.CloudProvider == awsCloudProvider
```
ivan-degtiarenko, line 274: I think just `central.CloudAccountID != ""` is enough

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
```diff
@@ -31,7 +31,7 @@ func NewDinosaurCNAMEManager(dinosaurService services.DinosaurService, kafkfConf
 			Reconciler: workers.Reconciler{},
 		},
 		dinosaurService: dinosaurService,
-		dinosaurConfig:  kafkfConfig,
+		dinosaurConfig:  dinosaurConfig,
```
ivan-degtiarenko, line 34: Let's not add more dinosaurs

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr_test.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+	defaultCfg := &config.CentralConfig{
+		Quota: config.NewCentralQuotaConfig(),
+	}
+
+	t.Run("no centrals, no problem", func(t *testing.T) {
+		dinoSvc := withCentrals()
+		quotaSvc, quotaFactory := withEntitlement(true)
+		gpm := NewGracePeriodManager(dinoSvc, quotaFactory, defaultCfg)
```
ivan-degtiarenko, line 48: nit: let's rename to something more telling, even just `mgr`

internal/dinosaur/pkg/services/quota/ams_quota_service.go
0x656b694d: TODO for future refactoring: all other methods need to be changed.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
0x656b694d: Ok.

internal/dinosaur/pkg/workers/dinosaurmgrs/dinosaurs_routes_cname_mgr.go
0x656b694d: TODO for future refactoring: rename dinosaur to central.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,122 @@
...
+func NewGracePeriodManager(dinosaurService services.DinosaurService, quotaServiceFactory services.QuotaServiceFactory, dinosaur *config.CentralConfig) *GracePeriodManager {
+	return &GracePeriodManager{
+		BaseWorker: workers.BaseWorker{
+			ID:         uuid.New().String(),
+			WorkerType: "grace_period_worker",
+			Reconciler: workers.Reconciler{},
+		},
+		dinosaurService:     dinosaurService,
+		quotaServiceFactory: quotaServiceFactory,
+		dinosaurConfig:      dinosaur,
```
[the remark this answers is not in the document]
0x656b694d: TODO for future codebase-wide refactoring.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr_test.go
0x656b694d: ok

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -213,3 +213,80 @@ func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError
 	}
 	return nil
 }
+
+func mapAllowedQuotaCosts(quotaCosts []*amsv1.QuotaCost) (map[amsv1.BillingModel][]*amsv1.QuotaCost, error) {
+	costsMap := make(map[amsv1.BillingModel][]*amsv1.QuotaCost)
+	var foundUnsupportedBillingModel string
```
stehessel, line 219: Why store only the last one? Wouldn't it be more practical to just return the list of unsupported billing models?

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -213,3 +213,80 @@ func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError
...
+	if err != nil {
+		svcErr := errors.ToServiceError(err)
+		return false, errors.NewWithCause(svcErr.Code, svcErr, "product %q has no allowed billing models", quotaType.GetProduct())
+	}
+
+	isCloudAccount := central.CloudAccountID != ""
+
+	entitled :=
+		// Entitlement is active if there's allowed quota for standard billing model...
+		!isCloudAccount && len(quotasMap[amsv1.BillingModelStandard]) > 0 ||
```
stehessel, line 283: Maybe extract the first condition into a separate variable for readability?

stehessel, line 283: Technically there is also the possibility that quota exists, but `allowed < consumed`. I.e. consider the case that 2 subscriptions gave quota for 20 instances, and 15 instances are created. Then one subscription expires, and there is only quota for 10 instances. The customer would then be 5 over the limit.

Note that this is an edge case, and in general we don't care much about the number of Centrals as long as the number is not unreasonably big. So I would be fine with ignoring this check here, and only enforce it on instance creation. Just wanted to point it out so it's a conscious decision to do so.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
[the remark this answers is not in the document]
stehessel: I'd encourage boy scout rules. Maybe a follow up PR to do the renaming so we actually do it?

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,127 @@
...
+
+	quotaService, factoryErr := k.quotaServiceFactory.GetQuotaService(api.QuotaType(k.dinosaurConfig.Quota.Type))
+	if factoryErr != nil {
+		return append(svcErrors, factoryErr)
+	}
+
+	subscriptionStatusByOrg := map[string]bool{}
+
+	for _, central := range centrals {
+		active, exists := subscriptionStatusByOrg[central.OrganisationID]
```
stehessel, line 89: What if an organisation has some Centrals that are expired and some that are entitled? I think we have to check it for every Central individually.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// if quota entitlement is not active and grace_from is not already set, set its value based on the current time and grace period allowance
+	if !isQuotaEntitlementActive && central.GraceFrom == nil {
+		now := time.Now()
+		central.GraceFrom = &now
+		glog.Infof("quota entitlement for central instance %q is no longer active, updating grace_from to %q", central.ID, now.Format(time.RFC1123Z))
+		return k.dinosaurService.Update(central)
+	}
+
+	glog.Infof("no grace_from changes needed for central %q, skipping update", central.ID)
```
stehessel, line 125: Not sure we need this log, could create some spam.

0x656b694d: ok

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
0x656b694d: The code was taken from the Kafka implementation. But your argument is valid for our case.
Here we call OMS to query for `quotacost(orgID)`. I'll introduce a cache to not abuse the OMS service.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
0x656b694d: Dumb copy-paste. Fixed.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
0x656b694d: I had such a check in the commit history, which would only log a message, that nobody would probably read. It had been suggested to remove the logging.

internal/dinosaur/pkg/migrations/20230911090600_add_grace_from_to_central_request.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"github.com/stackrox/acs-fleet-manager/pkg/db"
+	"gorm.io/gorm"
+)
+
+func addGraceFromFieldToCentralRequests() *gormigrate.Migration {
+	type CentralRequest struct {
+		db.Model
+		GraceFrom *time.Time `json:"grace_from"`
```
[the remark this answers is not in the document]
0x656b694d: Yes. I'm not sure I'll implement migration testing in this or in a parallel PR.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,127 @@
...
+	workers.BaseWorker
+	dinosaurService services.DinosaurService
+}
+
+// NewGracePeriodManager creates a new grace period manager
+func NewGracePeriodManager(dinosaurService services.DinosaurService) *DinosaurManager {
+	return &DinosaurManager{
+		BaseWorker: workers.BaseWorker{
+			ID:         uuid.New().String(),
+			WorkerType: "grace_period_worker",
```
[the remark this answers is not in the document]
0x656b694d: There could be various reasons why instances may run in grace period. All cases will be covered by this worker.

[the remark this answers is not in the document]
ivan-degtiarenko: @0x656b694d what could be the reasons?

pkg/client/ocm/cache.go
```diff
@@ -0,0 +1,58 @@
...
+		ts:   make(map[K]time.Time),
+	}
+}
+
+// Add a value to the cache. Cleans the cache up from expired data.
+// Returns the provided value.
+func (c *cacheImpl[K, V]) Add(key K, value V) V {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+	for k, ts := range c.ts {
```
ivan-degtiarenko, line 35: nit: right now average, best and worst write complexity of this cache is `O(n)`.  Perhaps we could introduce a minimal time that needs to pass between two cleanups?

pkg/client/ocm/cache.go
```diff
@@ -0,0 +1,58 @@
...
+// Cache provides methods to store and get values from a cache.
+type Cache[K comparable, V any] interface {
+	Add(key K, value V) V
+	Get(key K) (V, bool)
+}
+
+type cacheImpl[K comparable, V any] struct {
+	age  time.Duration
+	mux  sync.Mutex
+	data map[K]V
```
ivan-degtiarenko, line 17: nit: by uniting `V` and `time.Time` in a single struct we could reduce amount of map lookups

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,119 @@
...
+func (k *GracePeriodManager) reconcileCentralGraceFrom(centrals dbapi.CentralList) serviceErr.ErrorList {
+	var svcErrors serviceErr.ErrorList
+
+	quotaService, factoryErr := k.quotaServiceFactory.GetQuotaService(api.QuotaType(k.centralConfig.Quota.Type))
+	if factoryErr != nil {
+		return append(svcErrors, factoryErr)
+	}
+
+	for _, central := range centrals {
+		active, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(central, types.DinosaurInstanceType(central.InstanceType))
```
ivan-degtiarenko, line 88: 1. Why do we check that quota is defined, but return whether quota is active? AFAIU quota can be defined but be inactive(`consumed` > `allowed`).
2. Since `CheckIfQuotaIsDefinedForInstanceType` is called in `DetectInstanceType` which is called [before](https://github.com/stackrox/acs-fleet-manager/blob/d48925a3f952a067c83b2fa68eac82f66e500c7b/internal/dinosaur/pkg/services/dinosaur.go#L264-L264) `reserveQuota`, new hack that we introduced won't work as it is executed within `reserveQuota`

pkg/client/ocm/cache.go
```diff
@@ -0,0 +1,63 @@
+package ocm
+
+import (
+	"sync"
+	"time"
+)
+
+// Cache provides methods to store and get values from a cache.
+type Cache[K comparable, V any] interface {
+	Add(key K, value V) V
```
ivan-degtiarenko, line 10: I don't see big practical reason to return value here

pkg/client/ocm/cache.go
```diff
@@ -0,0 +1,63 @@
...
+// Returns the value and whether it exists in the cache.
+func (c *cacheImpl[K, V]) Get(key K) (V, bool) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+	rec, ok := c.data[key]
+	if ok && time.Since(rec.ts) > c.age {
+		delete(c.data, key)
+		ok = false
+	}
+	return rec.value, ok
```
ivan-degtiarenko, line 62: is it possible to make generic accept only pointer types and return `nil` here?

pkg/client/ocm/client.go
```diff
@@ -709,12 +714,16 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
...
-	quotaCostClient := organizationClient.Organization(organizationID).QuotaCost()
+	quotaCostList, ok := c.quotaCostCache.Get(organizationID)
+	if !ok {
+		organizationClient := c.connection.AccountsMgmt().V1().Organizations()
+		quotaCostClient := organizationClient.Organization(organizationID).QuotaCost()
 
-	quotaCostList, err := quotaCostClient.List().Parameter("fetchRelatedResources", true).Send()
-	if err != nil {
-		return nil, fmt.Errorf("retrieving relatedResources from the QuotaCosts service: %w", err)
+		qcl, err := quotaCostClient.List().Parameter("fetchRelatedResources", true).Parameter("fetchCloudAccounts", true).Send()
```
ivan-degtiarenko, line 722: I don't see under which conditions this value will be different from `quotaCostList`

pkg/client/ocm/cache.go
0x656b694d: Generic could operate with a pointer to the generic parameter. But why?

pkg/client/ocm/cache.go
0x656b694d: Syntactic sugar, see the use: https://github.com/stackrox/acs-fleet-manager/blob/54f591b36233fb837def3aea0e800015d05cbcdc/pkg/client/ocm/client.go#L726

pkg/client/ocm/client.go
0x656b694d: Do you mean it could be
```go
var err error
quotaCostList, err = quotaCostClient.List()...
```
?

pkg/client/ocm/cache.go
ivan-degtiarenko: Right now it's possible to return something like `<outdated value>, false` which assumes that the caller must check `ok` flag.

pkg/client/ocm/client.go
ivan-degtiarenko: Yes

pkg/client/ocm/client.go
```diff
@@ -709,12 +714,16 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
...
+		if err != nil {
+			return nil, fmt.Errorf("retrieving relatedResources from the QuotaCosts service: %w", err)
+		}
+		quotaCostList = c.quotaCostCache.Add(organizationID, qcl)
```
ivan-degtiarenko, line 726: Github does not let me comment on unmodified lines, but I think here
https://github.com/stackrox/acs-fleet-manager/pull/1248/files#diff-8bc7204e8ebade976e69f40acf5a0b8ad416e9d4bd53050bfcf5104217330c0aR733
we don't have to append the same quota cost multiple times if `resourceName` and `product` matches

pkg/client/ocm/cache.go
```diff
@@ -0,0 +1,63 @@
...
+
+type cacheImpl[K comparable, V any] struct {
+	age         time.Duration
+	mux         sync.Mutex
+	data        map[K]record[V]
+	lastCleanUp time.Time
+}
+
+// NewCache constructs a key-value cache that stores values up to age period.
+func NewCache[K comparable, V any](age time.Duration) *cacheImpl[K, V] {
```
ivan-degtiarenko, line 27: My IDE tells me:
```
Exported function with the unexported return type
```

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -235,3 +217,46 @@ func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError
 	}
 	return nil
 }
+
+func mapAllowedQuotaCosts(quotaCosts []*amsv1.QuotaCost) (map[amsv1.BillingModel][]*amsv1.QuotaCost, error) {
+	costsMap := make(map[amsv1.BillingModel][]*amsv1.QuotaCost)
+	var foundUnsupportedBillingModels []string
+	for _, qc := range quotaCosts {
+		// When an SKU entitlement expires in AMS, the allowed value for that quota cost is set back to 0.
+		// If the allowed value is 0 and consumed is greater than this, that denotes that the SKU entitlement
```
ivan-degtiarenko, line 226: We don't check that `consumed > 0` here

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -235,3 +217,46 @@ func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError
...
+			bm := amsv1.BillingModel(rr.BillingModel())
+			if _, isCompatibleBillingModel := supportedAMSBillingModels[rr.BillingModel()]; isCompatibleBillingModel {
+				costsMap[bm] = append(costsMap[bm], qc)
+			} else {
+				foundUnsupportedBillingModels = append(foundUnsupportedBillingModels, rr.BillingModel())
+			}
+
+		}
+	}
+	if len(costsMap) == 0 && len(foundUnsupportedBillingModels) > 0 {
```
ivan-degtiarenko, line 241: Why do we fail if we find an unsupported billing model? Can we just ignore them? I think that would be more error-proof approach considering quota configuration is external to `fleet-manager`

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -39,53 +38,36 @@ var supportedAMSBillingModels = map[string]struct{}{
...
+	quotasMap, err := mapAllowedQuotaCosts(quotaCosts)
 	if err != nil {
-		return false, fmt.Errorf("retrieving quota costs for product %s, organization ID %s, resource type %s: %w", quotaType.GetProduct(), organizationID, quotaType.GetResourceName(), err)
+		svcErr := errors.ToServiceError(err)
+		return false, errors.NewWithCause(svcErr.Code, svcErr, "product %q has no allowed billing models", quotaType.GetProduct())
 	}
 
-	var foundUnsupportedBillingModel string
+	isCloudAccount := central.CloudAccountID != ""
+	standardAccountIsActive := !isCloudAccount && len(quotasMap[amsv1.BillingModelStandard]) > 0
```
ivan-degtiarenko, line 60: What will happen if org has several subscriptions with billing model standard and one of them expires? Since we only check that `len(quotasMap[amsv1.BillingModelStandard]) > 0`, would that mean that customer will be able to have more instances that current subscriptions allow them?

Same question applies to non-standard billing model case.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -39,53 +38,36 @@ var supportedAMSBillingModels = map[string]struct{}{
...
-//   - Matches the given input quotaType
-//   - Contains at least one AMS RelatedResources whose billing model is one
-//     of the supported Billing Models specified in supportedAMSBillingModels
-//   - Has an "Allowed" value greater than 0
-//
-// An error is returned if the given organizationID has a QuotaCost
-// with an unsupported billing model and there are no supported billing models
-func (q amsQuotaService) hasConfiguredQuotaCost(organizationID string, quotaType ocm.DinosaurQuotaType) (bool, error) {
-	quotaCosts, err := q.amsClient.GetQuotaCostsForProduct(organizationID, quotaType.GetResourceName(), quotaType.GetProduct())
+	quotasMap, err := mapAllowedQuotaCosts(quotaCosts)
```
ivan-degtiarenko, line 53: `quotasMap` does not really tell what is the key and what is the value in this map

pkg/client/ocm/client.go
0x656b694d: Ok!

internal/dinosaur/pkg/services/quota/ams_quota_service.go
0x656b694d: Yup. Not sure what would be the difference if we checked.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
0x656b694d: We fail because there's no supported model. We decide on entitlement based on that, so I'd rather not swallow errors.

ivan-degtiarenko: So the error then is not that there is unsupported billing model, but rather that there are no supported billing models with active SKU entitlement. Let's make error message reflect that

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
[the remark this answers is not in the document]
0x656b694d: Contract expiration, subscription expiration, overconsumption due to quota change, account deleted, etc?

internal/dinosaur/pkg/services/quota/ams_quota_service.go
ivan-degtiarenko: Still named the same

internal/dinosaur/pkg/services/quota/ams_quota_service.go
ivan-degtiarenko: Not addressed

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -39,53 +38,36 @@ var supportedAMSBillingModels = map[string]struct{}{
...
-		return false, errors.OrganisationNotFound(dinosaur.OrganisationID, err)
+		return false, errors.OrganisationNotFound(central.OrganisationID, err)
 	}
 
-	hasQuota, err := q.hasConfiguredQuotaCost(org.ID(), instanceType.GetQuotaType())
+	quotaType := instanceType.GetQuotaType()
+	quotaCosts, err := q.amsClient.GetQuotaCostsForProduct(org.ID(), quotaType.GetResourceName(), quotaType.GetProduct())
 	if err != nil {
-		return false, errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("failed to get assigned quota of type %v for organization with external id %v and id %v", instanceType.GetQuotaType(), dinosaur.OrganisationID, org.ID()))
+		return false, errors.InsufficientQuotaError("%v: error getting quotas for product %q", err, quotaType.GetProduct())
```
ivan-degtiarenko, line 50: Why is this an insufficient quota error and not general error?

internal/dinosaur/pkg/services/quota/quota_management_list_service.go
```diff
@@ -25,10 +25,10 @@ func (q QuotaManagementListService) CheckIfQuotaIsDefinedForInstanceType(dinosau
 	org, orgFound := q.quotaManagementList.QuotaList.Organisations.GetByID(orgID)
 	userIsRegistered := false
 	if orgFound && org.IsUserRegistered(username) {
-		userIsRegistered = true
+		userIsRegistered = org.GetMaxAllowedInstances() > 0
```
ivan-degtiarenko, line 28: Let's not hide another condition under `userIsRegistred` variable

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,132 @@
...
+func (k *GracePeriodManager) Stop() {
+	k.StopWorker(k)
+}
+
+// Reconcile ...
+func (k *GracePeriodManager) Reconcile() []error {
+	glog.Infoln("reconciling grace period start date for central instances")
+	var encounteredErrors []error
+
+	centrals, svcErr := k.centralService.ListByStatus(
```
ivan-degtiarenko, line 61: Consider situation where another state would be added - how would the person who adds it know that it should be added here as well? Let's create some method that will return these statuses in `internal/dinosaur/constants/central.go`

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
```diff
@@ -0,0 +1,132 @@
...
+		instanceType   string
+	}
+
+	quotaCostCache := make(map[quotaCostCacheKey]bool)
+	for _, central := range centrals {
+		key := quotaCostCacheKey{central.OrganisationID, central.CloudAccountID, central.InstanceType}
+		defined, inCache := quotaCostCache[key]
+		if !inCache {
+			var svcErr *serviceErr.ServiceError
+			defined, svcErr = quotaService.CheckIfQuotaIsDefinedForInstanceType(central, types.DinosaurInstanceType(central.InstanceType))
```
ivan-degtiarenko, line 99: The naming is confusing - the method is called `CheckIfQuotaIsDefinedForInstanceType` and returns `defined` variable, but what it **actually** returns is `isQuotaEntitlementActive`. Why don't we rename the method accordingly?

pkg/client/ocm/client.go
```diff
@@ -711,21 +709,27 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
 	var res []*amsv1.QuotaCost
 	organizationClient := c.connection.AccountsMgmt().V1().Organizations()
 	quotaCostClient := organizationClient.Organization(organizationID).QuotaCost()
+	const pageSize = 100
```
ivan-degtiarenko, line 712: let's move `const` to the top of the file

pkg/client/ocm/cache.go
0x656b694d: The code has been removed.

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -196,12 +196,12 @@ func Test_AMSCheckQuota(t *testing.T) {
 				},
 				Owner: tt.args.owner,
 			}
-			sq, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(dinosaur, types.STANDARD)
+			standard_active, err := quotaService.IsQuotaActive(dinosaur, types.STANDARD)
```
ivan-degtiarenko, line 199: Let's not name variables with `_`

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -925,9 +929,245 @@ func Test_amsQuotaService_CheckIfQuotaIsDefinedForInstanceType(t *testing.T) {
...
 			quotaService, _ := quotaServiceFactory.GetQuotaService(api.AMSQuotaType)
-			res, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(tt.args.dinosaurRequest, tt.args.dinosaurInstanceType)
+			res, err := quotaService.IsQuotaActive(tt.args.dinosaurRequest, tt.args.dinosaurInstanceType)
 			gomega.Expect(err != nil).To(gomega.Equal(tt.wantErr))
 			gomega.Expect(res).To(gomega.Equal(tt.want))
 		})
 	}
 }
+
+func Test_amsQuotaService_CheckQuotaEntitlement(t *testing.T) {
```
ivan-degtiarenko, line 939:
```suggestion
func Test_amsQuotaService_IsQuotaActive(t *testing.T) {
```

internal/dinosaur/constants/central.go
```diff
@@ -91,12 +91,15 @@ func (k CentralStatus) CompareTo(k1 CentralStatus) int {
...
-	return []string{
-		CentralRequestStatusPreparing.String(),
-		CentralRequestStatusProvisioning.String(),
-		CentralRequestStatusFailed.String(),
-		CentralRequestStatusReady.String(),
+// ActiveStatuses are the centrals statuses, that precede deprovisioning.
+var ActiveStatuses = func() []CentralStatus {
+	active := []CentralStatus{}
+	d := ordinals[CentralRequestStatusDeprovision.String()]
+	for status, ordinal := range ordinals {
```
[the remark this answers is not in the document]
0x656b694d: Not sure we want to list statuses explicitely, as we may miss a later added status. Moreover, not sure we even want to limit the list of statuses for which we set the `expired_at`.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -39,54 +39,38 @@ var supportedAMSBillingModels = map[string]struct{}{
...
+		return false, errors.OrganisationNotFound(central.OrganisationID, err)
 	}
 
-	hasQuota, err := q.hasConfiguredQuotaCost(org.ID(), instanceType.GetQuotaType())
+	quotaType := instanceType.GetQuotaType()
+	quotaCosts, err := q.amsClient.GetQuotaCostsForProduct(org.ID(), quotaType.GetResourceName(), quotaType.GetProduct())
 	if err != nil {
-		return false, errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("failed to get assigned quota of type %v for organization with external id %v and id %v", instanceType.GetQuotaType(), dinosaur.OrganisationID, org.ID()))
+		return false, errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf(
+			"failed to get assigned quota of type %q for organization with external id %q and id %q",
```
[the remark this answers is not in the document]
0x656b694d: Won't hurt showing both either, I guess.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -39,54 +39,38 @@ var supportedAMSBillingModels = map[string]struct{}{
...
+			string(quotaType), central.OrganisationID, org.ID()))
```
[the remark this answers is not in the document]
0x656b694d: It would, indeed.

pkg/client/ocm/client.go
```diff
@@ -629,21 +626,28 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
...
+	req := quotaCostClient.List().Parameter("fetchRelatedResources", true).Parameter("fetchCloudAccounts", true).Size(pageSize)
+	for page := 1; ; page++ {
+		quotaCostList, err := req.Page(page).Send()
+		if err != nil {
+			return nil, fmt.Errorf("retrieving relatedResources from the QuotaCosts service: %w", err)
```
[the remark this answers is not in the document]
0x656b694d: `fmt.Errorf` is closer above, that's why.

pkg/client/ocm/client.go
```diff
@@ -629,21 +626,28 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
 	var res []*amsv1.QuotaCost
 	organizationClient := c.connection.AccountsMgmt().V1().Organizations()
 	quotaCostClient := organizationClient.Organization(organizationID).QuotaCost()
+	const pageSize = 100
+	req := quotaCostClient.List().Parameter("fetchRelatedResources", true).Parameter("fetchCloudAccounts", true).Size(pageSize)
+	for page := 1; ; page++ {
+		quotaCostList, err := req.Page(page).Send()
+		if err != nil {
+			return nil, fmt.Errorf("retrieving calling QuotaCosts service: %w", err)
```
ivan-degtiarenko, line 634: I think you meant something else here

internal/dinosaur/constants/central.go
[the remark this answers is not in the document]
ivan-degtiarenko: Listing statuses implicitly does not work here:
1. It does not work now, because in the only non-test usage we have to append another status that has too big of an ordinal. The only usage of this function already has to adjust the result.
2. It won't work in the future, because there's no guarantee our state machine will be sequential. What if we add status `Paused` - would it go before or after `Ready`/`Deprovision`? I think the question whether we expire `paused` instances would better be handled explicitly and not accidentally by choosing the wrong(or right) ordinal

> Moreover, not sure we even want to limit the list of statuses for which we set the `expired_at`.

Setting/unsetting (remember we do both depending on quota allowance) `expired_at` for instances in `Deprovision`/`Deleting` state would not change anything for them at best, at worst - create a confusing record in the database of central request that was deleted due to lack of quota allowance but has no `expired_at` set.
Let's work only with instances we can actually delete in expiration manager.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
ivan-degtiarenko: In that case let's remove the part about consumed

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -925,9 +929,245 @@ func Test_amsQuotaService_CheckIfQuotaIsDefinedForInstanceType(t *testing.T) {
...
+	standardCentral := &dbapi.CentralRequest{
+		InstanceType: "standard",
+	}
+
+	cloudCentral := &dbapi.CentralRequest{
+		InstanceType:   "standard",
+		CloudProvider:  awsCloudProvider,
+		CloudAccountID: "cloudAccountID",
+	}
+	const not_allowed = 0
```
ivan-degtiarenko, line 949: let's not name variables or constants with `_`

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -925,9 +929,245 @@ func Test_amsQuotaService_CheckIfQuotaIsDefinedForInstanceType(t *testing.T) {
...
+
+func makeCloudTestQuotaCost(resourceName string, organizationID string, allowed int, consumed int, t *testing.T) *v1.QuotaCost {
+	cloudAccount := v1.NewCloudAccount().CloudAccountID("cloudAccountID").CloudProviderID(awsCloudProvider)
+	rrbq := v1.NewRelatedResource().BillingModel(string(v1.BillingModelMarketplaceAWS)).Product(string(ocm.RHACSProduct)).ResourceName(resourceName).Cost(1)
+	qcb, err := v1.NewQuotaCost().Allowed(allowed).Consumed(consumed).OrganizationID(organizationID).RelatedResources(rrbq).CloudAccounts(cloudAccount).Build()
+	require.NoError(t, err)
+	return qcb
+}
+
+func makeOrganizationFromExtID(externalId string) (*v1.Organization, error) {
```
ivan-degtiarenko, line 1170:
```suggestion
func makeOrganizationFromExternalID(externalId string) (*v1.Organization, error) {
```

We can allow ourselves 5 more letters 😉

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -925,9 +929,245 @@ func Test_amsQuotaService_CheckIfQuotaIsDefinedForInstanceType(t *testing.T) {
...
+			g.Expect(err != nil).To(gomega.Equal(tt.wantErr))
+			if tt.wantErr {
+				g.Expect(err.Error()).To(gomega.Equal(tt.wantErrMsg), err.Error())
+			}
+			g.Expect(got).To(gomega.Equal(tt.want))
+		})
+	}
+}
+
+func makeStandardTestQuotaCost(resourceName string, organizationID string, allowed int, consumed int, t *testing.T) *v1.QuotaCost {
```
ivan-degtiarenko, line 1155:
```suggestion
func makeStandardTestQuotaCost(resourceName, organizationID string, allowed, consumed int, t *testing.T) *v1.QuotaCost {
```

pkg/client/ocm/client.go
```diff
@@ -629,21 +626,28 @@ func (c client) GetQuotaCostsForProduct(organizationID, resourceName, product st
 	var res []*amsv1.QuotaCost
 	organizationClient := c.connection.AccountsMgmt().V1().Organizations()
 	quotaCostClient := organizationClient.Organization(organizationID).QuotaCost()
+	const pageSize = 100
+	req := quotaCostClient.List().Parameter("fetchRelatedResources", true).Parameter("fetchCloudAccounts", true).Size(pageSize)
+	for page := 1; ; page++ {
```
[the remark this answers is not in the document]
0x656b694d: I temporarily set page size to 2 and added logging to the item processing functor in my local binary:
```go
		items++
		glog.V(10).Infof("fetched item %d", items)
```
And observed the following in the log, ensuring we're fetching all pages:
```console
$ kubectl logs fleet-manager-66d8fcd4df-ldnhd| grep fetched
Defaulted container "fleet-manager" out of: fleet-manager, db-migrate (init)
I1116 18:56:06.737468      13 client.go:641] fetched item 1
I1116 18:56:06.738223      13 client.go:641] fetched item 2
I1116 18:56:06.875013      13 client.go:641] fetched item 3
I1116 18:56:06.875026      13 client.go:641] fetched item 4
I1116 18:56:07.001512      13 client.go:641] fetched item 5
I1116 18:56:07.001549      13 client.go:641] fetched item 6
I1116 18:56:07.129293      13 client.go:641] fetched item 7
I1116 18:56:07.129342      13 client.go:641] fetched item 8
I1116 18:56:07.255719      13 client.go:641] fetched item 9
I1116 18:56:07.255738      13 client.go:641] fetched item 10
I1116 18:56:07.393687      13 client.go:641] fetched item 11
I1116 18:56:07.393729      13 client.go:641] fetched item 12
...
```
We may notice that the log timestamps are close in groups of 2 records, which is the size of the page.

internal/dinosaur/constants/central.go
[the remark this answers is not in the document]
0x656b694d: Ok.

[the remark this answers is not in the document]
0x656b694d: FTR. The statuses have ordinals exactly for the purpose of ordering. So if we pause a ready instance, it would be "Paused Ready" status, I guess. In this case iterating over statuses by ordinals have full sense.
The Failed status is an exceptional state, so handled exceptionally.

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
0x656b694d: O_k.

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
0x656b694d: I remember receiving a request to do the opposite. So I keep types for each parameter since. Sorry.

internal/dinosaur/pkg/services/quota/ams_quota_service_test.go
```diff
@@ -925,9 +929,245 @@ func Test_amsQuotaService_CheckIfQuotaIsDefinedForInstanceType(t *testing.T) {
...
+	rrbq := v1.NewRelatedResource().BillingModel(string(v1.BillingModelStandard)).Product(string(ocm.RHACSProduct)).ResourceName(resourceName).Cost(1)
+	qcb, err := v1.NewQuotaCost().Allowed(allowed).Consumed(consumed).OrganizationID(organizationID).RelatedResources(rrbq).Build()
+	require.NoError(t, err)
+	return qcb
+}
+
+func makeCloudTestQuotaCost(resourceName string, organizationID string, allowed int, consumed int, t *testing.T) *v1.QuotaCost {
```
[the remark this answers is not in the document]
0x656b694d: Same.

internal/dinosaur/pkg/workers/dinosaurmgrs/grace_period_mgr.go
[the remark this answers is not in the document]
0x656b694d: Renamed dinosaurs in the file.

internal/dinosaur/pkg/workers/dinosaurmgrs/expiration_date_mgr.go
```diff
@@ -0,0 +1,131 @@
+package dinosaurmgrs
```
SimonBaeumer, line 1: This could a service/manager which provides these function and being reconciled in the ready dinosaur manager. By doing this the ready worker finally gets a purpose and we don't need to introduce a new worker for a single field.

SimonBaeumer (changes requested):

pkg/client/ocm/paginated_request.go
```diff
@@ -0,0 +1,42 @@
...
+
+type paginatedRequest[RequestType any, ResponseType any] interface {
+	Size(int) RequestType
+	Page(int) RequestType
+	Send() (ResponseType, error)
+}
+
+func fetchPages[RQ paginatedRequest[RQ, RS], RS paginatedResponse[I], I pageItem[Data], Data any](
+	request RQ, pageSize int, maxPages int, f func(Data) bool) error {
+
```
ivan-degtiarenko, line 22: empty line

pkg/client/ocm/paginated_request.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+		keepGoing := true
+		response.Items().Each(func(data Data) bool {
+			keepGoing = f(data)
+			return keepGoing
+		})
+		if !keepGoing || response.Size() < pageSize {
+			break
+		}
+		if page == maxPages {
```
ivan-degtiarenko, line 37: I would rather put this check at the beginning of the cycle and replace with `page == maxPages + 1` so we would not break on the `maxPage` if it has `pageSize` elements

pkg/client/ocm/paginated_request.go
0x656b694d: Right, to separate function parameters from the body.

pkg/client/ocm/paginated_request.go
0x656b694d: `page` will never be equal to `maxPages+1` inside the loop.
Having this check here gives a chance to break normally in case the `page[maxPages]` is the last page.

internal/dinosaur/pkg/workers/dinosaurmgrs/expiration_date_mgr.go
0x656b694d: As agreed, let's refactor the worker as a separate effort.

pkg/client/ocm/paginated_request.go
ivan-degtiarenko: Why this function is different from others that it requires such separation?

pkg/client/ocm/paginated_request.go
```diff
@@ -0,0 +1,46 @@
...
+			return pkgerrors.New("too many pages")
+		}
+	}
```
ivan-degtiarenko, line 44: I think I need to correct my suggestion. I think it is still correct **not to fail** when `maxPages` page has `pageSize` elements.

```suggestion
	for page := 1; page <= maxPages + 1; page++ {
		response, err := req.Page(page).Send()
		if err != nil {
			return pkgerrors.Wrapf(err, "error retrieving page %d", page)
		}
		keepGoing := true
		response.Items().Each(func(data Data) bool {
			keepGoing = f(data)
			return keepGoing
		})
		if !keepGoing || response.Size() < pageSize {
			break
		}
		if page == maxPages + 1 &&  response.Size() > 0 {
			return pkgerrors.New("too many pages")
		}
	}
```

0x656b694d: While `response.Size() == pageSize` the result is considered incomplete. The logic of `maxPages` is to query not more than `maxPages`. Therefore, the function doesn't fetch `maxPages+1` page in the hope that last page was the last one.
Anyway, the `maxPages` feature is there mostly to prevent endless fetching, so there's no big difference if we set it to 1000 or 1001 in the real usecase.
I refactored the function to better indicate the error condition.

pkg/client/ocm/paginated_request.go
0x656b694d: Too long signature.

ivan-degtiarenko: I meant body from function parameters. I don't think we usually do that

pkg/client/ocm/paginated_request.go
```diff
@@ -0,0 +1,45 @@
...
+	Items() Items
+}
+
+type paginatedRequest[RequestType any, ResponseType any] interface {
+	Size(int) RequestType
+	Page(int) RequestType
+	Send() (ResponseType, error)
+}
+
+// fetchPages calls OCM API with size and page request parameters, allowing for
```
ivan-degtiarenko, line 20: If function is generic, the description should be generic as well

pkg/client/ocm/paginated_request.go
```diff
@@ -0,0 +1,45 @@
...
+// paged access to the data. Example request type: *amsv1.QuotaCostListRequest,
+// with *amsv1.QuotaCost as the data in the response list.
+// Iteration stops when all or maxPages pages are retrieved, or f returns false.
```
ivan-degtiarenko, line 23: That does not explain the error logic

pkg/client/ocm/paginated_request.go
0x656b694d: Agree. Updated.

pkg/client/ocm/paginated_request.go
0x656b694d: The code is less readable without this empty line. We usually should do that in such cases.

pkg/client/ocm/paginated_request.go
0x656b694d: Fixed.

PR #1256, opened 2023-09-13
Add kind cluster type

SimonBaeumer:
## Description

Add kind cluster type, additional logic was necessary to load images into kind for development.

## Test manual

- use kind to deploy fleet-manager via `make deploy/bootstrap && make deploy/dev`


Makefile
```diff
@@ -526,6 +518,10 @@ image/build: GOOS=linux
 image/build: IMAGE_REF ?= "$(external_image_registry)/$(image_repository):$(image_tag)"
 image/build: fleet-manager fleetshard-sync
 	DOCKER_CONFIG=${DOCKER_CONFIG} $(DOCKER) build -t $(IMAGE_REF) -f Dockerfile.hybrid .
+	$(DOCKER) tag $(IMAGE_REF) $(SHORT_IMAGE_REF)
+ifeq ("$(CLUSTER_TYPE)","kind")
+	kind load docker-image $(SHORT_IMAGE_REF)
+endif
```
ludydoo, line 524: I think this would be cleaner if it would be in `deploy/dev/fleetshard-sync` and `deploy/dev/fleetmanager`. Seems like the target `image/build` should just build the image and not "push" it

dev/env/defaults/cluster-type-minikube/env
```diff
@@ -6,3 +6,4 @@ export INHERIT_IMAGEPULLSECRETS_DEFAULT="true" # pragma: allowlist secret
 export INSTALL_OLM_DEFAULT="true"
 export ENABLE_EXTERNAL_CONFIG_DEFAULT="true"
 export AWS_AUTH_HELPER_DEFAULT="aws-saml"
+export IMAGE_REGISTRY=quay.io/rhacs-eng
```
ludydoo, line 9: Do you know why `IMAGE_REGISTRY` is required here and not in `cluster-type-kind/env` ?

Makefile
```diff
@@ -598,8 +594,8 @@ image/push/internal: docker/login/internal
 .PHONY: image/push/internal
 
 # build and push the image to an OpenShift cluster's internal registry
-# namespace used in the image repository must exist on the cluster before running this command. Run `make deploy/project` to create the namespace if not available.
-image/build/push/internal: image/build/internal image/push/internal
+# namespace used in the image repository must exist on the cluster before running this command. Run `make deploy/project` to create the   if not available.
+image/build/push/internal: image/push/internal
```
ludydoo, line 598: Can you explain why `image/build/internal` is removed?

Makefile
```diff
@@ -903,18 +899,14 @@ deploy/bootstrap:
 	./dev/env/scripts/bootstrap.sh
 .PHONY: deploy/bootstrap
 
-deploy/dev-fast: GOOS=linux
```
ludydoo, line 906: Can you explain the consequence of removing `GOOS=linux` ?

Makefile
```diff
@@ -903,18 +899,14 @@ deploy/bootstrap:
...
-deploy/dev-fast: deploy/dev-fast/fleet-manager deploy/dev-fast/fleetshard-sync
+# Deploy local images fast for development
+deploy/dev-fast: image/build deploy/dev-fast/fleet-manager deploy/dev-fast/fleetshard-sync
```
ludydoo, line 903: Wasn't `image-build` executed before?

ludydoo (changes requested):
It works on my machine. Though, I am not yet clear as to why the Makefile targets were changed in this way. Do you mind expanding on that and explaining your reasoning?

Makefile
```diff
@@ -45,21 +45,13 @@ SHORT_IMAGE_REF = "$(IMAGE_NAME):$(image_tag)"
 PROBE_SHORT_IMAGE_REF = "$(PROBE_IMAGE_NAME):$(image_tag)"
 
 # Default namespace for local deployments
-NAMESPACE ?= fleet-manager-${USER}
```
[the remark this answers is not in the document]
SimonBaeumer: As discussed offline:
Experimental at the moment as it is not documented and not used in scripts.
If we see engineers using the variables it must be reverted.

Makefile
SimonBaeumer: As discussed offline: We'll keep loading the image here. If it is was a push to a registry it would be more logic, here we can assume when `CLUSTER_TYPE=kind` is set that the build is executed locally.
Extracting it into another target either:
A) increases complexity by introducing a new Make target for loading an image
B) duplicating the code to load the image

Makefile
```diff
@@ -598,8 +594,8 @@ image/push/internal: docker/login/internal
 .PHONY: image/push/internal
 
 # build and push the image to an OpenShift cluster's internal registry
-# namespace used in the image repository must exist on the cluster before running this command. Run `make deploy/project` to create the namespace if not available.
```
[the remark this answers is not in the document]
SimonBaeumer: fixed

dev/env/defaults/cluster-type-minikube/env
SimonBaeumer: This should not be here, removed it

Makefile
SimonBaeumer: Make takes care that `image-build` is only executed once.

Makefile
SimonBaeumer: Reducing duplicate code. `GOOS=linux`  is already set on `image/build`.

Makefile
SimonBaeumer: It is already removed in main branch, the target does not exist anymore. It was a legacy.

PR #1283, opened 2023-09-18
ROX-19718: Add flag for fleetmanager gitops

ludydoo:
## Description

This PR exposes a new flag on fleet-manager to specify the path to the gitops configuration file. 
If the flag is not specified in the command-line args, it is assumed that no gitops configuration is used.

It also enables gitops by default on for development environments.   

- [ ] CI and all relevant tests are passing~~

## Test manual

1. Run as usual using `make deploy/dev`
2. Change the gitops configmap in dev/env/manifests/fleet-manager/04-gitops-config.yaml
3. Run `make deploy/dev/update-gitops-config` to update the configMap on k8s
4. Observe that the centrals are patched according to the config


dev/env/manifests/fleet-manager/02-fleet-manager-deployment.yaml
```diff
@@ -82,6 +82,9 @@ spec:
               name: secrets
             - mountPath: /config
               name: config
+            - mountPath: /gitops-config
```
SimonBaeumer, line 85: Mount should be optional as in production fleet-manager templates

internal/dinosaur/pkg/gitops/module.go
```diff
@@ -0,0 +1,25 @@
...
+
+// NewModule ...
+func NewModule() *Module {
+	return &Module{
+		ConfigPath: "",
+	}
+}
+
+// AddFlags ...
+func (s *Module) AddFlags(fs *pflag.FlagSet) {
```
SimonBaeumer, line 18: Why using a flag here for adding the mount path which is hardcoded in the Deployments? Do you expect the path to change, if yes, why? :thinking: 
imho a const for the path would be sufficient in this case.

SimonBaeumer (changes requested):

dev/env/manifests/fleet-manager/02-fleet-manager-deployment.yaml
ludydoo: Done

internal/dinosaur/pkg/gitops/module.go
ludydoo: It's from experience, I'm used to have the ability to specify the path to a config file from a flag or environment variable. Like `nginx -c /my/nginx-config.conf` or `PGDATA=/my/postgres/files postgres`.

The value is different (currently) between dev environment/ci and stage/prod. It is defined on dev/ci and not defined on stage/prod because we are not yet using gitops on stage/prod. 

Is it a blocker?

SimonBaeumer: > Is it a blocker?

No, it's not a blocker

SimonBaeumer (approved):

kurlov (approved):

PR #1285, opened 2023-09-19
ROX-16664: add KeyGenerator interface and implementations

johannes94:
## Description
This PR adds a `KeyGenerator` interface and respective `AES256KeyGenerator` and `KMSDataKeyGenerator` implementations of the interface.

This implementations will later be used to generate a encryption key to provide to a managed central tenant in order to encrypt secrets stored in its DB.

## Test manual

CI and local unit tests are sufficient.

```
# To run tests locally run:
```

### DIFF
--- a/fleetshard/pkg/cipher/aes256_cipher.go
+++ b/fleetshard/pkg/cipher/aes256_cipher.go
@@ -65,3 +65,20 @@ func (a AES256Cipher) Decrypt(ciphertext []byte) ([]byte, error) {
 
 	return plaintext, nil
 }
+
+// AES256KeyGenerator implements the KeyGenerator interface for AES256 keys
+type AES256KeyGenerator struct{}
+
+// Generate generates a Key compatible for AES256 ciphers
+func (g AES256KeyGenerator) Generate() ([]byte, error) {
+	return generateAESKey(32)
+}
+
+func generateAESKey(size int) ([]byte, error) {
+	key := make([]byte, size)
+	if _, err := io.ReadFull(rand.Reader, key); err != nil {
+		return nil, fmt.Errorf("generating random bytes for AES key: %w", err)
+	}
+
+	return key, nil
+}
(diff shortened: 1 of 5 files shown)


ludydoo:
Do we need to generate the keys? I thought KMS could be used to encrypt/decrypt directly, without us ever having access to key material

johannes94:
> Do we need to generate the keys? I thought KMS could be used to encrypt/decrypt directly, without us ever having access to key material
@ludydoo 

There are two reasons for that.

1. Central will not implement a KMS integration for now

- The "central" component itself has no strict dependency to AWS like we have for our cloud service, so core workflows decided to not implement KMS integration for this feature because specifying a key is a more general implementation that all customers could potentially use. The interface for us to use will only be specifying a AES256 key in a K8s secret. That's why we need the key material in this case.

2. KMS API encryption/decryption is limited to 4KB. 

I learned this during the secret encryption I've implemented for secret backups. In case you need to encrypt more than that you have to generate a data key and implement encryption in your own application. https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#data-keys. In this case central-ca was to big to be encrypted via API, so we had to go the data key route.

For notifiers we would have the same issue because sometimes the credentials for notifiers can be certificates.

ludydoo:
Also interesting: https://github.com/hashicorp/go-kms-wrapping

johannes94:
> BTW, is [ROX-16663](https://issues.redhat.com//browse/ROX-16663) the right jira ticket? (from the title) I'm trying to get more context on the encryption feature.

You're right the actual ticket is [ROX-16664](https://issues.redhat.com/browse/ROX-16664). 

> For 2, it would be possible to use envelope encryption instead of sending the whole block to KMS. https://docs.aws.amazon.com/wellarchitected/latest/financial-services-industry-lens/use-envelope-encryption-with-customer-master-keys.html

IIUC this is essentially what is implemented in the secret backup feature.

Encryption:
1. Generate a data key
2. Encrypt data key with KMS key, encrypted payload with data key
3. Store encrypted data key with encrypted payload

Decryption:
1. Get encrypted data key and encrypted payload
2. Decrypt data key with KMS key
3. Decrypt payload with data key

vladbologa (approved):
LGTM

PR #1287, opened 2023-09-20
ROX-19735: PR suggestions

ludydoo:


dev/env/manifests/fleet-manager/04-gitops-config.yaml
```diff
@@ -5,8 +5,23 @@ metadata:
   namespace: "$ACSMS_NAMESPACE"
 data:
   config.yaml: |
+    rhacsOperators:
+      crd:
+        baseURL: https://raw.githubusercontent.com/stackrox/stackrox/{{ .GitRef }}/operator/bundle/manifests/
+        gitRef: 4.1.1
```
ludydoo, line 11: Do we really need a template here? Why not just specify the url directly

dev/env/manifests/fleet-manager/04-gitops-config.yaml
```diff
@@ -5,8 +5,23 @@ metadata:
...
+      operators:
+        - gitRef: 4.1.1
+          image: "quay.io/rhacs-eng/stackrox-operator:4.1.1"
```
ludydoo, line 14: I would favor adding the `labelSelector` here

fleetshard/pkg/central/operator/config.go
```diff
@@ -17,25 +16,12 @@ func parseConfig(content []byte) (OperatorConfigs, error) {
 	return out, nil
 }
 
-// GetConfig returns the rhacs operator configurations
```
ludydoo, line 20: Unused

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -372,30 +370,6 @@ func (r *CentralReconciler) applyCentralConfig(remoteCentral *private.ManagedCen
 	r.applyProxyConfig(central)
 	r.applyDeclarativeConfig(central)
 	r.applyAnnotations(central)
-	return r.applyLabelSelector(remoteCentral, central)
-}
-
-func (r *CentralReconciler) applyLabelSelector(remoteCentral *private.ManagedCentral, central *v1alpha1.Central) error {
```
ludydoo, line 378: label is applied using the gitops configMap

dev/env/manifests/fleet-manager/04-gitops-config.yaml
SimonBaeumer: Most of the times it will only be necessary to adjust the GitRef instead of the whole URL. I've used a template in this case because `%s` seemed not obvious enough if a new engineer not familiar with the config looks at the file.
Example: `https://raw.githubusercontent.com/stackrox/stackrox/%s/operator/bundle/manifests/`

dev/env/manifests/fleet-manager/04-gitops-config.yaml
SimonBaeumer: What would you prefer, `centralVersionSelector` or naming the label?

```
  operators:
    - gitRef: 4.1.1
      centralVersionSelector: 4.1.1
      image: "quay.io/rhacs-eng/stackrox-operator:4.1.1"
```

vs naming the label explicitly as a key, it makes it very explicit how the correspond to each other:

```
  operators:
    - gitRef: 4.1.1
      rhacs.redhat.com/version-selector: 4.1.1
      image: "quay.io/rhacs-eng/stackrox-operator:4.1.1"
```
```

File: fleetshard/pkg/central/operator/config.go [hunk 5]
```diff
@@ -17,25 +16,12 @@ func parseConfig(content []byte) (OperatorConfigs, error) {
...
 // Validate validates the operator configuration and can be used in different life-cycle stages like runtime and deploy time.
 func Validate(path *field.Path, configs OperatorConfigs) field.ErrorList {
-	manager := ACSOperatorManager{}
-	manifests, err := manager.RenderChart(configs)
-	if err != nil {
+	if _, err := RenderChart(configs); err != nil {
 		return field.ErrorList{
 			field.Forbidden(path, fmt.Sprintf("could not render operator helm charts, got invalid configuration: %s", err.Error())),
 		}
-	} else if len(manifests) == 0 {
```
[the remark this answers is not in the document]
SimonBaeumer: It's both, I've intended to catch templating errors. When an empty config is provided it is valid, when config contains operators it is invalid.

```
} else if len(configs.OperatorConfigs) > 0&& len(manifests) == 0 {
```

fleetshard/pkg/central/reconciler/reconciler.go
SimonBaeumer: How is the default case handled? Does the default Central add the label to all instances outside of a group?

SimonBaeumer (commented):
Thanks for providing these fixes @ludydoo!

dev/env/manifests/fleet-manager/04-gitops-config.yaml
ludydoo: Perhaps the simplest would be to align with the helm values structure?

fleetshard/pkg/central/reconciler/reconciler.go
ludydoo: Currently not, but we could either
- Specify a path with `instanceIds: ["*"]` where the label selector is set
- Change the default central template. 
- Other?

fleetshard/pkg/central/operator/config.go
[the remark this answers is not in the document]
SimonBaeumer: Done and added tests

dev/env/manifests/fleet-manager/04-gitops-config.yaml
ludydoo: Seems like the current helm template uses `labelSelector` as a key

SimonBaeumer: What is the simplest for the user? I think this is the most idiomatic `rhacs.redhat.com/version-selector: 4.1.1`  value.

PR #1289, opened 2023-09-20
Call /current_account before reserving quota

ivan-degtiarenko:
## Description
This PR adds a call to `/current_account` [API endpoint](https://api.openshift.com/?urls.primaryName=Accounts%20management%20service#/default/get_api_accounts_mgmt_v1_current_account).

The reason behind this hack is that when an organization is just created, it might take up to 30 minutes for AMS to populate quota data—calling `/current_account` forces AMS to populate this data earlier.

Since this is optimization and not a requirement to reserve quota, the errors are logged and not returned to the customer.

It is possible to optimize the performance by calling `/current_account` only once per user or organization. However, I would suggest to do it in a follow-up.

## Test manual

1. `make deploy/bootstrap`
2. Specify `CM_SERVICE_CLIENT_ID_DEFAULT` and `OCM_SERVICE_CLIENT_SECRET_DEFAULT`
3. Modify `fleet_manager_command` in https://github.com/stackrox/acs-fleet-manager/blob/35c22dffb47f1ed57ad968bc932c594e53ba5e26/dev/env/scripts/lib.sh to
```
local fleet_manager_command="/usr/local/bin/fleet-manager serve --force-leader --api-server-bindaddress=0.0.0.0:8000 --health-check-server-bindaddress=0.0.0.0:8083 --kubeconfig=/secrets/kubeconfig --enable-central-external-certificate=$ENABLE_CENTRAL_EXTERNAL_CERTIFICATE --central-domain-name='$CENTRAL_DOMAIN_NAME' --gitops-config-path='/gitops-config/config.yaml' --quota-type=ams --ams-base-url=https://api.openshift.com --enable-ocm-mock=false"
```
4. POST new central:
```
{"name":"test-on-cluster","cloud_provider":"standalone","region":"standalone","multi_az":true}
```
and get `Cloud account ID is not set up properly` error. 
5. Look into the logs and see there's no errors related to calling `/current_account`

### DIFF
--- a/internal/dinosaur/pkg/services/quota/ams_quota_service.go
+++ b/internal/dinosaur/pkg/services/quota/ams_quota_service.go
@@ -171,16 +171,11 @@ func (q amsQuotaService) ReserveQuota(ctx context.Context, dinosaur *dbapi.Centr
 		return "", errors.NewWithCause(errors.ErrorGeneral, err, "Error reserving quota")
 	}
 
-	userToken, err := authentication.TokenFromContext(ctx)
-	if err != nil {
-		// TODO
-		return "", errors.FailedClusterAuthorization(err)
-	}
-	_, err = q.amsClient.GetCurrentAccount(userToken.Raw)
-	if err != nil {
-		// TODO
-		return "", errors.FailedClusterAuthorization(err)
-	}
+	// The reason to call /current_account here is how AMS functions.
+	// In case customer just created the account, AMS might miss information about their quota.
+	// Calling /current_account endpoint results in this data being populated.
+	// Since this is a non-requirement for successful quota reservation, errors are logged but ignored here.
+	q.callCurrentAccount(ctx)
 
 	resp, err := q.amsClient.ClusterAuthorization(cb)
 	if err != nil {
@@ -191,6 +186,21 @@ func (q amsQuotaService) ReserveQuota(ctx context.Context, dinosaur *dbapi.Centr
 		return resp.Subscription().ID(), nil
 	}
 	return "", errors.InsufficientQuotaError("Insufficient Quota")
+}
+
+func (q amsQuotaService) callCurrentAccount(ctx context.Context) {
+	userToken, err := authentication.TokenFromContext(ctx)
+	if err != nil {
+		glog.Warningf("Couldn't extract user token from context: %w", err)
+		return
+	}
+	if userToken == nil {
+		return
+	}
+	_, err = q.amsClient.GetCurrentAccount(userToken.Raw)
+	if err != nil {
+		glog.Warningf("Call to  current account endpoint was unsuccessful: %w", err)
+	}
 }
 
 func (q amsQuotaService) verifyCloudAccountInAMS(dinosaur *dbapi.CentralRequest, orgID string) *errors.ServiceError {
--- a/internal/dinosaur/pkg/services/quota/ams_quota_service.go
+++ b/internal/dinosaur/pkg/services/quota/ams_quota_service.go
@@ -2,7 +2,9 @@
 package quota
 
 import (
+	"context"
 	"fmt"
+	"github.com/openshift-online/ocm-sdk-go/authentication"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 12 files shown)


rukletsov (approved):
This implements the workaround suggested by the OCM team so let's try it out. After this is merged, could you please test the happy path in the integration env, i.e., a new account subscribed for trial can immediately create an instance?

pkg/client/ocm/client.go
```diff
@@ -74,24 +76,19 @@ type AMSClient Client
...
-		MetricsSubsystem("api_outbound")
-
+	builder := getBaseConnectionBuilder(baseURL)
 	if !ocmConfig.EnableMock {
 		// Create a logger that has the debug level enabled:
-		logger, err := sdkClient.NewGoLoggerBuilder().
-			Debug(ocmConfig.Debug).
-			Build()
+		var err error
+		builder, err = addLoggerToConnectionBuilder(ocmConfig.Debug, builder)
```
0x656b694d, line 88:
```suggestion
		logger, err := getLogger(ocmConfig.Debug)
```

pkg/client/ocm/client.go
```diff
@@ -74,24 +76,19 @@ type AMSClient Client
...
 		if err != nil {
-			return nil, nil, fmt.Errorf("creating logger for OCM client connection: %w", err)
+			return nil, nil, err
 		}
```
0x656b694d, line 91:
```suggestion
			builder.Logger(logger)
		}
```

pkg/client/ocm/client.go
0x656b694d: Not sure the function requires extraction.

internal/dinosaur/pkg/services/quota/ams_quota_service.go
```diff
@@ -169,6 +171,12 @@ func (q amsQuotaService) ReserveQuota(dinosaur *dbapi.CentralRequest, instanceTy
 		return "", errors.NewWithCause(errors.ErrorGeneral, err, "Error reserving quota")
 	}
 
+	// The reason to call /current_account here is how AMS functions.
+	// In case customer just created the account, AMS might miss information about their quota.
+	// Calling /current_account endpoint results in this data being populated.
+	// Since this is a non-requirement for successful quota reservation, errors are logged but ignored here.
+	q.callCurrentAccount(ctx)
```
stehessel, line 178: Shouldn't we call this before we call
```
org, err := q.amsClient.GetOrganisationFromExternalID(dinosaur.OrganisationID)
	if err != nil {
		return "", errors.OrganisationNotFound(dinosaur.OrganisationID, err)
	}
```
earlier in the function?

pkg/client/ocm/client.go
```diff
@@ -742,3 +756,25 @@ func (c *client) GetCustomerCloudAccounts(organizationID string, quotaIDs []stri
...
+	if err != nil {
+		return nil, fmt.Errorf("couldn't build modified OCM connection: %w", err)
+	}
+	defer modifiedConnection.Close()
+	response, err := modifiedConnection.AccountsMgmt().V1().CurrentAccount().Get().Send()
+	if err != nil {
+		return nil, fmt.Errorf("unsuccessful call to current account endpoint: %w", err)
+	}
+
+	currentAccount := response.Body()
```
stehessel, line 778: Do we need to close the response body here?

ivan-degtiarenko: Both `response` and `currentAccount` are `ocm-sdk-go` library types and don't have `Close()` method, so I would say no :)

internal/dinosaur/pkg/services/quota/ams_quota_service.go
stehessel: We have seen the 404 error in the past for new accounts. Now AMS has changed something on their end and perhaps we only see 403 errors now, but how confident are you that all edge cases have been tested? 

I'd just move the workaround to the start, because logically it seems like `GetOrganisationFromExternalID` could fail if the organisation has not yet been reconciled - and therefore does not exist yet in AMS.

stehessel (approved):

PR #1378, opened 2023-10-18
ROX-19904: Add OCM cluster id override

kovayur:
## Description
For deploying the addon to a dev osd cluster it is convenient to use `acs-dev-dp-01` config, but it has a dummy cluster id that doesn't match real ocm cluster id. This change allows to load `acs-dev-dp-01` config and override the cluster id with the environment variable `OVERRIDE_CLUSTER_ID`

## Test manual
```
./dp-terraform/ocm/install_addon.sh dev acs-dev-dp-01
```

### DIFF
--- a/dp-terraform/ocm/install_addon.sh
+++ b/dp-terraform/ocm/install_addon.sh
@@ -106,8 +106,11 @@ load_external_config "audit-logs/${CLUSTER_NAME}" AUDIT_LOGS_
 echo "Loading external config: cluster-${CLUSTER_NAME}"
 load_external_config "cluster-${CLUSTER_NAME}" CLUSTER_
 
+# Allows to load an external cluster config (e.g. acs-dev-dp-01) and apply it to a different cluster with override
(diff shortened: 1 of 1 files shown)


dp-terraform/ocm/install_addon.sh
```diff
@@ -106,8 +106,11 @@ load_external_config "audit-logs/${CLUSTER_NAME}" AUDIT_LOGS_
 echo "Loading external config: cluster-${CLUSTER_NAME}"
 load_external_config "cluster-${CLUSTER_NAME}" CLUSTER_
 
+# Allows to load an external cluster config (e.g. acs-dev-dp-01) and apply it to a different cluster with override
```
ebensh, line 109: Sounds more like this is the output or target cluster?

ebensh (approved):

dp-terraform/ocm/install_addon.sh
kovayur: Correct, `CLUSTER_ID` is the output of the parameter store call.
Instead of passing it directly to ocm, now we can override it

PR #1418, opened 2023-10-26
ROX-18504: GitOps Cleanup

ludydoo:


cmd/fleet-manager/main_test.go
```diff
@@ -38,7 +38,7 @@ func TestInjections(t *testing.T) {
 
 	var bootList []environments.BootService
 	env.MustResolve(&bootList)
-	Expect(len(bootList)).To(Equal(7))
+	Expect(len(bootList)).To(Equal(6))
```
ludydoo, line 41: Seems to be because we are removing the DefaultVersionService from the `di` container

SimonBaeumer (approved):
BAAAAAAAAAM :wastebasket:

PR #1431, opened 2023-11-02
Add pprof server to fleet-manager and unlimited evals in testing&develop

SimonBaeumer:
## Description
 - Add pprof server to fleet-manager on port 6060
 - Add name parameter to `create-central.sh`
 - Allow more instance for eval instances in development
 - Increase data-plane capacity in development

## Test manual

 - /

### DIFF
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -3,7 +3,6 @@ package services
 import (
 	"context"
 	"fmt"
-	"github.com/stackrox/acs-fleet-manager/pkg/environments"
 	"sync"
 	"time"
 
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stackrox/acs-fleet-manager/pkg/client/iam"
 	dynamicClientAPI "github.com/stackrox/acs-fleet-manager/pkg/client/redhatsso/api"
 	"github.com/stackrox/acs-fleet-manager/pkg/client/redhatsso/dynamicclients"
+	"github.com/stackrox/acs-fleet-manager/pkg/environments"
 
 	dinosaurConstants "github.com/stackrox/acs-fleet-manager/internal/dinosaur/constants"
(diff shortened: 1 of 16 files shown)


internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -3,6 +3,7 @@ package services
 import (
 	"context"
 	"fmt"
+	"github.com/stackrox/acs-fleet-manager/pkg/environments"
```
ludydoo, line 6: This import should be in the block below no?

pkg/server/profiler/profiler_server.go
```diff
@@ -0,0 +1,59 @@
...
+		pprofServerInstance = &PprofServer{}
+	})
+	return pprofServerInstance
+}
+
+func (p *PprofServer) Stop() {
+}
+
+func (p *PprofServer) Listen() (net.Listener, error) {
+	return nil, nil
```
ludydoo, line 40: Why not use the listener here for the pprof server ?

templates/service-template.yml
```diff
@@ -1479,3 +1479,17 @@ objects:
...
+      name: fleet-manager-pprof
+      labels:
+        app: fleet-manager
+        port: pprof
+    spec:
+      selector:
+        app: fleet-manager
+      ports:
+        - port: 6060
+          targetPort: 6060
```
ludydoo, line 1495: I'm not sure about creating a service for it. Because fleet-manager runs with multiple replicas, we never know which one we hit. We should just do a port forward and use that rather than creating a service for it ?

kovayur (commented):
Looks good, I echo Ludovic's review

pkg/server/profiler/profiler_server.go
SimonBaeumer: Because Listen-func is used only for `Listen` and not for starting the server. Only the API Server uses the `Listen` function.

See docs here on the funs for more info: https://github.com/stackrox/acs-fleet-manager/blob/ac4af1502d2dee370156557c21d2746da2970bf1/pkg/server/api_server.go#L158-L187

internal/dinosaur/pkg/services/dinosaur.go
SimonBaeumer: Sí, done

ludydoo (approved):

PR #1434, opened 2023-11-03
Render managed centrals in parallel

ludydoo:


internal/dinosaur/pkg/handlers/data_plane_dinosaur.go
```diff
@@ -80,12 +81,19 @@ func (h *dataPlaneDinosaurHandler) GetAll(w http.ResponseWriter, r *http.Request
...
 			for i := range centralRequests {
-				converted, err := h.presenter.PresentManagedCentral(centralRequests[i])
-				if err != nil {
-					return nil, errors.GeneralError("failed to convert central request to managed central: %v", err)
-				}
-				managedDinosaurList.Items = append(managedDinosaurList.Items, converted)
+				index := i
+				g.Go(func() error {
+					var err error
+					managedDinosaurList.Items[index], err = h.presenter.PresentManagedCentral(centralRequests[index])
```
SimonBaeumer, line 90: I see multiple challenges with this approach:

**1. Race conditions:**

This could lead to a race condition when to goroutines try to allocate the same address at the same time. The `index` is AFAIK not guaranteed to allocated the memory in an order.

You can still do this in two different ways:
 - We now the sice of the slice ahead of time, allocating the memory in the `make` state should do the trick to address the indexes according to their address
 - My favorite is using a channel, send the results to a channel and iterate over it, every result is appended to the result slice. This prevents shared memory access.

**2. Excessive go routine creation**
A limit is necessary for go routines to not create an unlimited amount of them. 

**3. Go routine stacking and cancellation:**
When a go routine is blocked it is never cancelled or timed out. Fleetshard-Sync will timeout and continuously try to get Centrals. An additional call to `GetAll` triggers again a Go routine for every active Central.

ludydoo: > This could lead to a race condition when to goroutines try to allocate the same address at the same time. The index is AFAIK not guaranteed to allocated the memory in an order.

I don't think that's true, the index is declared within the loop. It will only be scoped to the goroutine. 

> We now the sice of the slice ahead of time, allocating the memory in the make state should do the trick to address the indexes according to their address

That's already the case:
```
managedDinosaurList.Items = make([]private.ManagedCentral, len(centralRequests))
```

> Go routine stacking and cancellation:

If the context is canceled, the errgroup will stop

ludydoo: I added a max parallelism of 50

SimonBaeumer: > I don't think that's true, the index is declared within the loop. It will only be scoped to the goroutine.

The memory allocation in the slice is not bound to the index. But this would only apply if you hadn't allocated the memory beforehand, I've not seen that :+1: 

> If the context is canceled, the errgroup will stop

Uh, interesting I did not know that. TIL again.

SimonBaeumer (approved):

PR #1447, opened 2023-11-07
disable OpenShift monitoring for internal Centrals

stehessel:
## Description

This changes disables OpenShift monitoring for internal Centrals. This is done for these reasons:
* Probe instances are cleaned up immediately, so there is no need for any monitoring.
* Since probe instances are created very frequently, the collected metrics ramp up Prometheus storage.
* Prevent sending useless metrics to Telemeter once that is enabled.

Note that we still collect kube metrics, this just disables the metrics that are emitted by Central itself and collected by OpenShift monitoring.

## Test manual

CI

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -286,6 +286,7 @@ func (r *CentralReconciler) applyCentralConfig(remoteCentral *private.ManagedCen
 	r.applyProxyConfig(central)
 	r.applyDeclarativeConfig(central)
 	r.applyAnnotations(central)
+	r.applyMonitoring(remoteCentral, central)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -359,6 +360,22 @@ func (r *CentralReconciler) applyTelemetry(remoteCentral *private.ManagedCentral
 	central.Spec.Central.Telemetry = telemetry
 }
 
+func (r *CentralReconciler) applyMonitoring(remoteCentral *private.ManagedCentral, central *v1alpha1.Central) {
```
ludydoo, line 363: We are trying to move the reconciliation logic away from fleetshard and onto fleetmanager as much as possible. Can you move this logic within the 
internal/dinosaur/pkg/gitops/default_central.yaml tamplate?

stehessel: How can I do the internal Central check in the template?

stehessel: Nvm `isInternal` is exposed in the template. Removed the fleetshard code and replaced with snippet in template.

internal/dinosaur/pkg/gitops/default_central_test.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func TestDefaultCentral(t *testing.T) {
-	p := getDummyCentralParams()
+	p := getDummyCentralParams(false)
```
ludydoo, line 16: would it be possible to just ad-hoc set isInternal to true when needed for testing ? 
```suggestion
	p := getDummyCentralParams()
	p.IsInternal = false / true
```

internal/dinosaur/pkg/gitops/default_central_test.go
```diff
@@ -102,7 +102,119 @@ func TestDefaultCentral(t *testing.T) {
...
+func TestInternalCentral(t *testing.T) {
+	p := getDummyCentralParams(true)
+	central, err := renderDefaultCentral(p)
+	assert.NoError(t, err)
+
+	exposeEndpointEnabled := v1alpha1.ExposeEndpointEnabled
+	autoScalingEnabled := v1alpha1.ScannerAutoScalingEnabled
+	scannerComponentEnabled := v1alpha1.ScannerComponentEnabled
+
+	wantCentral := v1alpha1.Central{
```
ludydoo, line 133: Here you could extract a global `wantCentralForDummyParams`, and use
```
p := getDummyCentralParams()
p.IsInternal.= true
gotCentral, err := renderDefaultCentral(p)
require.NoError(t, err)
wantCentral := wantCentralForDummyParams.DeepCopy()
wantCentral.Spec.Monitoring.Enabled = false
assert.Equal(wantCentral, gotCentral)
```

ludydoo (approved):

PR #1454, opened 2023-11-09
ROX-20736: move observability to infra nodes

ludydoo:


ludydoo:
Closing as it is not possible to schedule any workloads on infra nodes :(

PR #1467, opened 2023-11-14
Increase prom memory on prod

ludydoo:


stehessel (approved):

ludydoo:
@SimonBaeumer I think the pprof is making tests fail (`TestProfProfiler: dial tcp [::1]:6060: connect: connection refused`). Have you seen this before?

PR #1517, opened 2023-12-05
chore: drop cpu limits from egress proxy

stehessel:
## Description

Noticed that we still set cpu limits for the egress proxy. We already removed cpu limits from other deployments because it leads to unnecessary throttling. On example of that is in https://redhat.pagerduty.com/incidents/Q39LMR8IUSF7XG?utm_campaign=channel&utm_source=slack.


ludydoo (approved):

kurlov (approved):

ludydoo:
whoops, I think I addressed this in https://github.com/stackrox/acs-fleet-manager/pull/1518

stehessel:
superseded by #1518

PR #1518, opened 2023-12-07
Adjust egress proxy resources and replicas

ludydoo:
Bumped the memory of egress proxy
Reduced from 3 to 2 replicas

### DIFF
--- a/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
+++ b/fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
@@ -1,12 +1,11 @@
 egressProxy:
   image: ubuntu/squid:5.2-22.04_beta
-  replicas: 3
+  replicas: 2
   resources:
     requests:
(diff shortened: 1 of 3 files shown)


fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
```diff
@@ -1,12 +1,11 @@
 egressProxy:
   image: ubuntu/squid:5.2-22.04_beta
-  replicas: 3
```
stehessel, line 3: Why the change from 3 to 2? Does it matter that we'll have less throughput performance with only 2 replicas?

ludydoo: The 3 replicas are overkill. They are idling 99% of the time and just consuming memory:

Screenshot 2023-12-07 at 2 55 29 PM
Screenshot 2023-12-07 at 2 56 14 PM

PR #1520, opened 2023-12-08
ROX-20479: service registration

ludydoo:
We have a race condition with the gitops config.

The gitops config is mounted using a `configMap` volume. When the configmap changes, it will be hot-reloaded into the pod. Though, there is a significant delay (couple seconds) for those changes to be reflected. 

Fleetshard will hit a random (load-balanced) fleetmanager pod. So, when the config changes, there will be a few polls that will return different data, because the configmap has been reloaded in some, but not all, fleetmanager pods. This will happen until all fleet-manger pods have hot-reloaded the new configmap version.

Solution: 
1. Introduce a leader election mechanism to elect a single fleetmanager pod
2. When a fleetmanager becomes leader, it annotates its own POD with "fleet-manager-active: true". If it loses the leader election, it sets "fleet-manager-active: false".
3. An additional OpenShift route is introduced (fleet-manager-active) targeting only the active fleet-manager pod.
4. Fleetshard is configured to reach fleetmanager-active route. 

Using this, fleetshard will only reach 1 single fleet-manager, and the gitops config will be consistent. 

Inspired by https://github.com/hashicorp/vault/blob/main/serviceregistration/kubernetes/retry_handler.go

NOTE: This PR doesn't change the current behavior. A progressive rollout is required:
- Step 1. A new route will need to be introduced on app-interface
- Step 2. The (generated) route URL will need to be propagated to dp-terraform


pkg/serviceregistration/leader.go
```diff
@@ -0,0 +1,157 @@
...
+				value, ok := pod.Labels[activeLabel]
+				if ok && value == isActive {
+					return nil
+				}
+			}
+
+			glog.Infof("[serviceregistration] updating pod %s/%s labels: %s=%s", l.namespaceName, l.podName, activeLabel, isActive)
+			patch := `{"metadata":{"labels":{"` + activeLabel + `":"` + isActive + `"}}}`
+
+			_, err = l.client.CoreV1().Pods(l.namespaceName).Patch(ctx, l.podName, types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{})
```
SimonBaeumer, line 147: Use `.Pods(namespace).Update()` instead of a formatted patch string?

```
glog.Infof("[serviceregistration] updating pod %s/%s labels: %s=%s", l.namespaceName, l.podName, activeLabel, isActive)
pod.Labels[activeLabel] = isActive
_, err = l.client.CoreV1().Pods(l.namespaceName).Update(ctx, pod, metav1.UpdateOptions{})
```

SimonBaeumer (changes requested):
The approach and code looking good to me, but it introduces a second leader election to the existing in [leader_election_mgr.go](https://github.com/stackrox/acs-fleet-manager/blob/ac4af1502d2dee370156557c21d2746da2970bf1/pkg/workers/leader_election_mgr.go).
The env variable `RHACS_LEADER_ELECTION_ENABLED` introduces a new configuration which increases adds a new variable which diverges between environments like testing/dev/prod.

pkg/workers/leader_election_mgr.go
```diff
@@ -1,224 +0,0 @@
-// Package workers ...
-package workers
```
SimonBaeumer, line 2: :fire: :fire: :fire: :fire: :fire:

pkg/workers/leader_election_mgr_test.go
```diff
@@ -1,185 +0,0 @@
-package workers
```
SimonBaeumer, line 1: :fire: :fire: :fire: :fire: :fire: :fire:

.github/workflows/ci.yaml
```diff
@@ -56,6 +56,7 @@ jobs:
     env:
       # TODO make sure that the secrets are configured for your repository
       OCM_ENV: integration
+      E2E: "true"
```
SimonBaeumer, line 59: This looks a bit like an accident. Is this change correct?

SimonBaeumer (approved):
:fire: :fire: :fire: :ship:

.github/workflows/ci.yaml
ludydoo: Thanks for the review! This is required, because during those tests there are no kubernetes clusters available, and we have to mock the kubernetes apiserver. Check pkg/serviceregistration/service.go, there's a switch to create a fake vs real client during those tests

PR #1521, opened 2023-12-08
ROX-21070: Declare Data Plane Cluster configuration in gitops config

kovayur:
## Description
Install addon from Fleet Manager: Part 3.
Add the Data Plane clusters configuration to gitops config in order to update the addons via fleet manager 

## Test manual

```
# To run tests locally run:
```

### DIFF
--- a/internal/dinosaur/pkg/gitops/provider_test.go
+++ b/internal/dinosaur/pkg/gitops/provider_test.go
@@ -177,3 +177,98 @@ centrals:
 	assert.Equal(t, 2, validationFnCalls)
 
 }
+
+func TestProviderGet_DataPlaneClusters(t *testing.T) {
+	successfulValidation := func(config Config) error {
+		return nil
+	}
+	type tc struct {
+		name            string
+		file            string
+		expectedConfigs []DataPlaneClusterConfig
+	}
+
+	tcs := []tc{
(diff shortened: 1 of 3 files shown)


internal/dinosaur/pkg/gitops/provider_test.go
```diff
@@ -177,3 +177,98 @@ centrals:
...
+							Parameters: map[string]string{
+								"acscsEnvironment": "test",
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	for _, tc := range tcs {
```
SimonBaeumer, line 255: This assignment could cause a race condition when tests are run in parallel.
The `tc` needs to be reassigned inside the for-loop.
```
for _, tc := range tcs {
  testCase := tc
  t.Run(testCase.name, func(t *testing.T) {
  // and so on...
```

SimonBaeumer (approved):

internal/dinosaur/pkg/gitops/provider_test.go
kovayur: Does it make sense without setting `t.Parallel()` ? It looks like a preemptive measure to me. At least I haven't encountered such a pattern in stackrox before.

PR #1555, opened 2024-01-08
ROX-21596: add PVC for alertmanager

stehessel:
## Description

Add a PVC to the alertmanager deployment. This should prevent silences getting lost after pod restarts. Also did a small refactoring of the Prometheus PVC by introducing values for it.


SimonBaeumer (approved):
Great change for our work-life-balance, thank you! :heart:

PR #1557, opened 2024-01-09
ROX-20479: fix leader retry

ludydoo:
The logic used to retry in the leaderelection used apimachinery exponential backoff. But this actually does not retry infinitely. Implemented a simple exponential backoff to retry forever.

### DIFF
--- a/pkg/serviceregistration/leader.go
+++ b/pkg/serviceregistration/leader.go
@@ -131,9 +131,9 @@ func (l *leaderWorker) updateWithRetry(ctx context.Context) {
 			if err := l.update(ctx); err == nil {
 				return
 			}
-			// sleep min(1000, 10 * 2 ^ attempt milliseconds)
-			sleepDuration := math.Min(baseSleepMillis*math.Pow(2, float64(attempt)), maxSleepMillis)
-			time.Sleep(time.Duration(sleepDuration) * time.Millisecond)
+			// sleep min(1000, 10 * 2 ^ attempt) milliseconds
+			timer := time.NewTimer(time.Duration(math.Min(baseSleepMillis*math.Pow(2, float64(attempt)), maxSleepMillis)) * time.Millisecond)
+			<-timer.C
 			attempt++
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/serviceregistration/leader.go
```diff
@@ -118,42 +118,58 @@ func (l *leaderWorker) run(ctx context.Context) error {
...
-			pod, err := l.client.CoreV1().Pods(l.namespaceName).Get(ctx, l.podName, metav1.GetOptions{})
-			if err != nil {
-				glog.Errorf("[serviceregistration] error getting pod: %v", err)
-				return errors.Wrap(err, "failed to get pod")
+			if err := l.update(ctx); err == nil {
+				return
 			}
+			// sleep min(1000, 10 * 2 ^ attempt milliseconds)
+			sleepDuration := math.Min(baseSleepMillis*math.Pow(2, float64(attempt)), maxSleepMillis)
+			time.Sleep(time.Duration(sleepDuration) * time.Millisecond)
```
SimonBaeumer, line 136: `time.Sleep` is blocking and cannot be interrupt, e.g. when the Pod terminates. `time.Ticker` can be interrupted (see https://gobyexample.com/tickers)

SimonBaeumer (changes requested):
Is the problem that the exponential backoff by Kubernetes interprets a failed leader-election try as a failure? 

I am concerned that we miss errors in the leader election, e.g. all pods fail to become a leader and we are not notified (maybe probe would catch it?). 

The code looks good otherwise :+1:

ludydoo:
> Is the problem that the exponential backoff by Kubernetes interprets a failed leader-election try as a failure?
> 
> I am concerned that we miss errors in the leader election, e.g. all pods fail to become a leader and we are not notified (maybe probe would catch it?).
> 
> The code looks good otherwise 👍

Thanks :) The issue with the current logic is that the Kubernetes retry will not retry forever, but only 5 times :/ . I saw this error on integration, when fleetmanager was missing some api role to update its label. But it would only retry 5 times then stop. I think probe would catch it, because it wouldn't start any of the workers

SimonBaeumer (approved):

PR #1549, opened 2024-01-02
feat: central labels

0x656b694d:
## Description

Added `labels` column to the `central_requests` table, which may hold a (json) key-value dictionary with random json values.
API:
* show instance labels: `GET /v1/admin/centrals/<id>`
* get label JSON value: `GET /v1/admin/centrals/<id>/labels/<label>`
* add/modify a label: `PATCH /v1/admin/centrals/<id>/labels/<label>`, provided JSON value in the body
* delete a label: `DELETE /v1/admin/centrals/<id>/labels/<label>`

## Checklist (Definition of Done)

- [ ] Unit and integration tests added
- [ ] Added test description under `Test manual`
- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)
- [ ] CI and all relevant tests are passing
- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.
- [ ] Add secret to app-interface Vault or Secrets Manager if necessary
- [ ] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)
- [ ] Check AWS limits are reasonable for changes provisioning new resources

## Test manual

```console
# get unexisting label
$ curl -H "$AUTH" "http://localhost:8000/api/rhacs/v1/admin/centrals/cl75fu8bs0bd6k334h0g/labels/a"
{"id":"7","kind":"Error","href":"/api/rhacs/v1/errors/7","code":"RHACS-MGMT-7","reason":"No such label","operation_id":"cma3ajtfh90d6k3upjf0"}

# create label `a` with integer value `42`
$ curl -H "$AUTH" "http://localhost:8000/api/rhacs/v1/admin/centrals/cl75fu8bs0bd6k334h0g/labels/a" -X PATCH -d 42
{"a":42}

# get central request with labels
$ curl -H "Authorization: Bearer $STATIC_TOKEN_ADMIN" "http://localhost:8000/api/rhacs/v1/admin/centrals/cl75fu8bs0bd6k334h0g"                       
[...]

PR #1564, opened 2024-01-15
feat: support for central trait `preserved`

parametalol:
Make use of #1549: do not deprovision centrals which have "preserved" trait.
+ E2E tests for the traits.

### DIFF
--- a/e2e/e2e_central_traits_test.go
+++ b/e2e/e2e_central_traits_test.go
@@ -44,7 +44,12 @@ var _ = Describe("central traits", Ordered, func() {
 	})
 
 	It("should manage traits", func() {
-		central, _, err := adminAPI.CreateCentral(ctx, true, private.CentralRequestPayload{})
+		central, _, err := adminAPI.CreateCentral(ctx, true, private.CentralRequestPayload{
+			CloudProvider: dpCloudProvider,
+			MultiAz:       true,
+			Name:          newCentralName(),
+			Region:        dpRegion,
+		})
 		Expect(err).Should(Succeed())
 		id := central.Id
 
@@ -94,7 +99,12 @@ var _ = Describe("central traits", Ordered, func() {
 	})
 
 	It("should preserve preserved", func() {
-		central, _, err := adminAPI.CreateCentral(ctx, true, private.CentralRequestPayload{})
+		central, _, err := adminAPI.CreateCentral(ctx, true, private.CentralRequestPayload{
+			CloudProvider: dpCloudProvider,
+			MultiAz:       true,
+			Name:          newCentralName(),
+			Region:        dpRegion,
+		})
 		Expect(err).Should(Succeed())
 		Eventually(assertCentralRequestProvisioning(ctx, client, central.Id)).
 			WithTimeout(waitTimeout).
--- a/e2e/e2e_central_traits_test.go
+++ b/e2e/e2e_central_traits_test.go
@@ -115,7 +115,7 @@ var _ = Describe("central traits", Ordered, func() {
 		Expect(err).Should(Succeed())
 
 		resp, err := client.PublicAPI().DeleteCentralById(ctx, central.Id, true)
-		Expect(resp.StatusCode).To(Equal(http.StatusBadRequest))
+		Expect(resp.StatusCode).To(Equal(http.StatusConflict))
 		Expect(err).To(HaveOccurred())
 
 		_, err = adminAPI.DeleteDbCentralById(ctx, central.Id)
--- a/e2e/e2e_central_traits_test.go
+++ b/e2e/e2e_central_traits_test.go
@@ -2,18 +2,17 @@ package e2e
 
 import (
 	"context"
+	"net/http"
 	"os"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	"github.com/pkg/errors"
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/constants"
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/admin/private"
 	"github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager"
 	fmImpl "github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager/impl"
-	serviceErrors "github.com/stackrox/acs-fleet-manager/pkg/errors"
 )
 
 var _ = Describe("central traits", Ordered, func() {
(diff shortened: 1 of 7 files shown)


0x656b694d:

* [main](../tree/main)
  * **PR #1549**
    * **PR #1564** 👈

internal/dinosaur/pkg/services/dinosaur_test.go
```diff
@@ -175,6 +176,43 @@ func Test_dinosaurService_Get(t *testing.T) {
 	}
 }
 
+func Test_dinosaurService_Delete_Presered(t *testing.T) {
```
stehessel, line 179:
```suggestion
func Test_dinosaurService_Delete_Preserved(t *testing.T) {
```

stehessel:
I think we need an e2e test here as well that actually executes the query.

0x656b694d:
> I think we need an e2e test here as well that actually executes the query.

I added e2e tests for generic traits and this `preserved` trait specifically.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -60,6 +62,10 @@ type CNameRecordStatus struct {
 	Status *string
 }
 
+var errPreservedCentral = func(id string) *errors.ServiceError {
+	return errors.BadRequest("central %q has %s trait", id, constants.CentralTraitPreserved)
```
rukletsov, line 66: Should be 409 I think. The error message (probably in the caller given there are two) should also include the remediation hint, I guess it will be removing the trait or using force.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -492,6 +498,10 @@ func (k *dinosaurService) RegisterDinosaurDeprovisionJob(ctx context.Context, id
 	}
 	metrics.IncreaseCentralTotalOperationsCountMetric(dinosaurConstants.CentralOperationDeprovision)
 
+	if arrays.Contains(dinosaurRequest.Traits, constants.CentralTraitPreserved) {
```
rukletsov, line 501: Why do we need a check here in addition to the one during deletion?

0x656b694d: This is for the case when the function is called by the HTTP handler: the check allows for returning the error immediately, instead of marking the instance for deletion and fail to delete it later.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -60,6 +63,11 @@ type CNameRecordStatus struct {
 	Status *string
 }
 
+var errPreservedCentral = func(id string) *errors.ServiceError {
+	return errors.NewErrorFromHTTPStatusCode(http.StatusConflict,
+		"central %q has %s trait, remove the trait to enable deletion", id, constants.CentralTraitPreserved)
```
stehessel, line 68: Can you update the openapi spec to include 409 for delete?

internal/dinosaur/pkg/services/dinosaur_test.go
```diff
@@ -175,6 +177,69 @@ func Test_dinosaurService_Get(t *testing.T) {
...
+		dinosaurConfig: &config.CentralConfig{
+			CentralLifespan: config.NewCentralLifespanConfig(),
+		},
+	}
+	preserved := buildCentralRequest(func(centralRequest *dbapi.CentralRequest) {
+		centralRequest.Traits = append(centralRequest.Traits, constants.CentralTraitPreserved)
+	})
+	svcErr := k.Delete(preserved, false)
+	assert.Equal(t, http.StatusConflict, svcErr.HTTPCode)
+	{
```
stehessel, line 192: This block feels like it should be extracted into a separate function.

e2e/e2e_central_traits_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+		traits, _, err = adminAPI.GetCentralTraits(ctx, id)
+		Expect(err).ToNot(HaveOccurred(), "no error on having multiple traits")
+		Expect(traits).To(BeEquivalentTo([]string{"test-trait", "test-trait-1"}), "should have only two traits")
+
+		_, err = adminAPI.GetCentralTrait(ctx, id, "test-trait")
+		Expect(err).ToNot(HaveOccurred(), "no error on checking for existing trait")
+
+		_, err = adminAPI.GetCentralTrait(ctx, id, "test-trait-2")
+		Expect(err).To(HaveOccurred(), "error on checking for non-existing trait")
```
stehessel, line 88: Could assert 404 here.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -575,6 +589,11 @@ func (k *dinosaurService) DeprovisionExpiredDinosaurs() *errors.ServiceError {
 // If the force flag is true, then any errors prior to the final deletion of the CentralRequest will be logged as warnings
 // but do not interrupt the deletion flow.
 func (k *dinosaurService) Delete(centralRequest *dbapi.CentralRequest, force bool) *errors.ServiceError {
```
stehessel, line 591: Is the `force` option exposed by the console UI? It feels like we are mixing up different concepts a bit. What deletion scenarios are we trying to prevent?

1. User deletes their own instance. Do we want to stop anyone from doing this? They will be confused what's going on because the traits are managed via the admin API. Traits will be generally unknown to users, so why expose them to users?
2. (Accidental) deletion by admins? Admins control traits, but admin API ignores traits.
3. Unwanted/early expiry of instances. Wouldn't it more sense to prevent expiry in the first place, rather than stop the deprovision of an already expired instance?

0x656b694d: No, `force` is not exposed. In fact, `force=true` if only an admin deletes the central via `/centrals/db/{id}` endpoint. I was thinking that this would be the only way to delete a preserved instance, including users not being able to do so.

Your 3rd point makes sense. The trait should be called `perpetual` then. We could actually have both traits, but I think `perpetual` is more interesting.

ivan-degtiarenko: IMHO both admin and user API calls to delete instances should be respected despite `preserved` trait. If user wants to delete an instance, they should be able to do that without additional steps (they already have to type in instance name in the UI to confirm). If an admin wants to delete a user instance, they most likely have a good reason for it.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -512,6 +524,7 @@ func (k *dinosaurService) DeprovisionDinosaurForUsers(users []string) *errors.Se
 		Model(&dbapi.CentralRequest{}).
 		Where("owner IN (?)", users).
 		Where("status NOT IN (?)", dinosaurDeletionStatuses).
+		Where("traits IS NULL OR ? != ALL (traits)", dinosaurConstants.CentralTraitPreserved).
```
ludydoo, line 527: By curiosity, what does `? != ALL (traits)` mean?

0x656b694d: This is a specific syntax for column of type array: https://www.postgresql.org/docs/current/arrays.html#ARRAYS-SEARCHING

PR #1569, opened 2024-01-15
ROX-21701: add PATCH for Central name

stehessel:
## Description

Add PATCH path for changing the Central name of a tenant. We've had a customer request to rename their Central instance, and are currently unable to do so.

## Test manual

CI e2e test


e2e/e2e_suite_test.go
```diff
@@ -143,6 +143,19 @@ func isDNSEnabled(routesEnabled bool) (bool, string, string) {
...
 }
 
+func assertCentralRequestName(ctx context.Context, client *fleetmanager.Client, id string, name string) func() error {
+	return func() error {
+		centralRequest, _, err := client.PublicAPI().GetCentralById(ctx, id)
+		if err != nil {
+			return err
+		}
+		if centralRequest.Name != name {
+			return fmt.Errorf("expected centralRequest name %s, got %s", name, centralRequest.Name)
```
dhaus67, line 153: super nit: on the formatting, since this is request / response data we print out, `%q` might be more obvious to detect e.g. empty return values.

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -272,3 +274,31 @@ func (h adminCentralHandler) PatchExpiredAt(w http.ResponseWriter, r *http.Reque
...
+				return nil, errors.NewWithCause(errors.ErrorBadRequest, err, "Unmarshalling request body: %s", err.Error())
+			}
+			if updateNameRequest.Name == "" {
+				return nil, errors.New(errors.ErrorBadRequest, "No name provided")
+			}
+			if updateNameRequest.Reason == "" {
+				return nil, errors.New(errors.ErrorBadRequest, "No reason provided")
+			}
+			id := mux.Vars(r)["id"]
+			glog.Warningf("Setting name to %q for central %q: %s", updateNameRequest.Name, id, updateNameRequest.Reason)
```
dhaus67, line 296: Why is this a "warn" log instead of just informational?

internal/dinosaur/pkg/handlers/admin_dinosaur.go
```diff
@@ -272,3 +274,31 @@ func (h adminCentralHandler) PatchExpiredAt(w http.ResponseWriter, r *http.Reque
...
+	cfg := &handlers.HandlerConfig{
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+			bodyBytes, err := io.ReadAll(r.Body)
+			if err != nil {
+				return nil, errors.NewWithCause(errors.ErrorBadRequest, err, "Reading request body: %s", err.Error())
+			}
+			updateNameRequest := private.CentralUpdateNameRequest{}
+			if err := json.Unmarshal(bodyBytes, &updateNameRequest); err != nil {
+				return nil, errors.NewWithCause(errors.ErrorBadRequest, err, "Unmarshalling request body: %s", err.Error())
+			}
```
dhaus67, line 288: Given that we seem to duplicate this logic of reading the body and unmarshalling the contents into a given struct, let's factor this out and create a helper for this.

johannes94 (changes requested):
Did you test that against an actual running fleetshard-sync / k8s cluster?

The reason why I'm asking is that, if I recall correctly the name of the tenant is used as the name of the central custom resource created by fleetshard-sync.

The problem about that:

If the name of the tenant changes, fleetshard-sync reconciler would pick up the tenant and query the namespace for a central CR having that name. Since it wouldn't find the old CR after the name change it would create a new one. Now we would have 2 CRs in one namespace, which will cause operator reconciliation conflicts.

johannes94 (approved):
We've addressed my previous comment in a private conversation. We think we can work around the concerns by:
- pause reconciling before changing the name
- rename through API
- delete the namespace
- unpause reconcile
- let fleetshard-sync restore the NS from backup

We can test that process on the integration environment.

I'm fine with merging if the e2e test is green.

kurlov (approved):

PR #1602, opened 2024-01-18
ROX-21779 Add envoy active listener

ludydoo:
Openshift ingress does not forward SNI headers, so envoy is unable to distinguish requests coming from fleet-manager-envoy vs fleet-manager-active services, and unable to resolve the correct certificate. 

The solution is to add a separate listener on a different port, to server the fleet-manager-active service and use the correct certificate.

### DIFF
--- a/templates/service-template.yml
+++ b/templates/service-template.yml
@@ -859,87 +859,15 @@ objects:
                 address: 0.0.0.0
                 port_value: 9001
             listener_filters:
-            - name: tls_inspector
-              typed_config:
-                "@type": type.googleapis.com/envoy.extensions.filters.listener.tls_inspector.v3.TlsInspector
             filter_chains:
-            - filter_chain_match:
-                server_names:
-                  - "fleet-manager-active"
-                  - "fleet-manager-active.${NAMESPACE}"
-                  - "fleet-manager-active.${NAMESPACE}.svc"
-                  - "fleet-manager-active.${NAMESPACE}.svc.cluster"
-                  - "fleet-manager-active.${NAMESPACE}.svc.cluster.local"
-                  - "romndkjdq62p7sr.api.integration.openshift.com"
-              transport_socket:
-                name: envoy.transport_sockets.tls
-                typed_config:
-                  "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.DownstreamTlsContext
-                  full_scan_certs_on_sni_mismatch: true
-                  common_tls_context:
-                    tls_certificates:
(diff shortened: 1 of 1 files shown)


templates/service-template.yml
```diff
@@ -859,87 +859,15 @@ objects:
                 address: 0.0.0.0
                 port_value: 9001
             listener_filters:
-            - name: tls_inspector
-              typed_config:
-                "@type": type.googleapis.com/envoy.extensions.filters.listener.tls_inspector.v3.TlsInspector
             filter_chains:
-            - filter_chain_match:
```
ludydoo, line 866: This whole filter chain was an attempt at using sni-based certificate selection, which doesn't work with openshift ingress

templates/service-template.yml
```diff
@@ -859,87 +859,15 @@ objects:
                 address: 0.0.0.0
                 port_value: 9001
             listener_filters:
-            - name: tls_inspector
```
ludydoo, line 862: TLS inspector was also added to perform SNI-based routing. No longer needed

templates/service-template.yml
```diff
@@ -859,87 +859,15 @@ objects:
...
-                      "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
-                  - name: envoy.filters.http.router
-                    typed_config:
-                      "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
-
             - transport_socket:
                 name: envoy.transport_sockets.tls
                 typed_config:
                   "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.DownstreamTlsContext
-                  full_scan_certs_on_sni_mismatch: true
```
ludydoo, line 936: This was added to try selecting the proper certificate

templates/service-template.yml
```diff
@@ -859,87 +859,15 @@ objects:
...
                   common_tls_context:
                     tls_certificates:
                     - certificate_chain: {filename: "/secrets/tls/tls.crt"}
                       private_key: {filename: "/secrets/tls/tls.key"}
-                    - certificate_chain: {filename: "/secrets/active-tls/tls.crt"}
-                      private_key: {filename: "/secrets/active-tls/tls.key"}
```
ludydoo, line 942: This listener doesn't use this certificate, it was added to try resolving the correct cert

johannes94 (approved):
Looks reasonable. I'm looking at this envoy config for the first time may be worth to get another review, but as I understand the only difference here is the TLS cert served by the new listener.

PR #1617, opened 2024-01-23
fix: update for secret backup for tenants

johannes94:
## Description

The check intended to prevent fleet-manager from overwriting secrets if they are already stored broke the update logic in case new secrets were added.

The check was still useful because it protected FM from overwriting secrets with an empty map if no secrets are reported by FS, so I only changed the logic to validate the incoming request instead of the database state. 

~~- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`~~

## Test manual

```
# To run tests locally run:
make ocm/setup
```

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -786,7 +786,7 @@ func (r *CentralReconciler) collectReconciliationStatus(ctx context.Context, rem
 	}
 
 	// Only report secrets if Central is ready, to ensure we're not trying to get secrets before they are created.
-	// Only report secrets once. Ensures we don't overwrite initial secrets with corrupted secrets
+	// Only report secrets if not all secrets are already stored to ensures we don't overwrite initial secrets with corrupted secrets
 	// from the cluster state.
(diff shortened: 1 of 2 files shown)


vladbologa (approved):
Looks good! I think that in addition to this, we should modify the code comment on the `fleetshard-sync` side here: https://github.com/stackrox/acs-fleet-manager/blob/main/fleetshard/pkg/central/reconciler/reconciler.go#L789

The comment there says that we report secrets only once, which is no longer the case.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -786,7 +786,7 @@ func (r *CentralReconciler) collectReconciliationStatus(ctx context.Context, rem
 	}
 
 	// Only report secrets if Central is ready, to ensure we're not trying to get secrets before they are created.
-	// Only report secrets once. Ensures we don't overwrite initial secrets with corrupted secrets
+	// Only report secrets if not all secrets are already stored to ensures we don't overwrite initial secrets with corrupted secrets
```
vladbologa, line 789:
```suggestion
	// Only report secrets if not all secrets are already stored to ensure we don't overwrite initial secrets with corrupted secrets
```

vladbologa (approved):

PR #1619, opened 2024-01-25
ROX-21946: FS set central-tls OwnerReference if not already set

johannes94:
## Description

After restore of central-tls secret the owner reference is not set to the central CR again. This blocks the operator from automatically rotating leaf certificates for central, since it won't touch secrets not owned by it.

With this PR fleetshard-sync, will set the owner reference of the secret to the CR it creates, which should fix the issue.

## Test manual

CI and added e2e tests should be sufficient

```
# To run tests locally run:
make ocm/setup
```

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -922,8 +922,13 @@ func (r *CentralReconciler) ensureCentralTLSSecretHasOwnerReference(ctx context.
 		return nil
 	}
 
+	centralCR := &v1alpha1.Central{}
+	if err := r.client.Get(ctx, ctrlClient.ObjectKeyFromObject(central), centralCR); err != nil {
+		return fmt.Errorf("getting current central CR from k8s: %w", err)
+	}
+
 	secret.OwnerReferences = []metav1.OwnerReference{
-		*metav1.NewControllerRef(central, central.GroupVersionKind()),
+		*metav1.NewControllerRef(centralCR, centralCR.GroupVersionKind()),
 	}
 
 	if err := r.client.Update(ctx, secret); err != nil {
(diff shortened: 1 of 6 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -627,7 +631,8 @@ func (r *CentralReconciler) reconcileCentral(ctx context.Context, remoteCentral
 
 	centralExists := true
 	existingCentral := v1alpha1.Central{}
-	centralKey := ctrlClient.ObjectKey{Namespace: remoteCentralNamespace, Name: remoteCentralName}
+	centralKey := ctrlClient.
```
stehessel, line 634: Why this change? The formatting before seems more clear and the line length is not longer than other lines.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -897,6 +902,39 @@ func (r *CentralReconciler) encryptSecrets(secrets map[string]*corev1.Secret) (m
 
 }
 
+// ensureCentralTLSSecretHasOwnerReference is used to make sure the central-tls secret has it's
+// owner reference properly set after a restore operation so that the automatic cert rotation
+// in the operator is working
+func (r *CentralReconciler) ensureCentralTLSSecretHasOwnerReference(ctx context.Context, remoteCentral *private.ManagedCentral, central *v1alpha1.Central) error {
```
stehessel, line 908: Any reason that this function is specific to the `central-tls` secret? I would suggest to make it a bit more generic and pass the secret name as an argument.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -897,6 +902,39 @@ func (r *CentralReconciler) encryptSecrets(secrets map[string]*corev1.Secret) (m
...
+	secret, err := r.getSecret(remoteCentral.Metadata.Namespace, k8s.CentralTLSSecretName)
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
```
stehessel, line 911: Does the previous reconcile step ensure that the secret exists? Otherwise we may fail to ensure that the owner reference is set if the secret is created after this function is called.

stehessel (approved):

PR #1629, opened 2024-01-31
ROX-20872: fix fleetshard reconciler race condition

ludydoo:
## Description

Fixing of the fleetshard cache bug. 

How to reproduce: 
1. Reconcile central v1
2. Reconcile centra v2
3. Deployment is unhealthy (restarting)
4. Reconcile central v1
5. Reconcile central v2 (ignored/skipped)

The reason that the reconcile loop at step 5 is ignored is that the hash is not stored at step 4, even if the central CR was updated. At step 5, the reconciler will be skipped because the hash (v2) matches the last stored hash (step 2). 

The effect is that v1 is deployed, but the reconciler thinks it reconciled v2.

The fix is to unset the hash if the reconciler exited early. 

As an added precaution, trigger the reconcile loop if central was reconciled more than 15 minutes ago.

### DIFF
--- a/fleetshard/pkg/central/reconciler/reconciler.go
+++ b/fleetshard/pkg/central/reconciler/reconciler.go
@@ -187,7 +187,9 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 	remoteCentralNamespace := remoteCentral.Metadata.Namespace
 
 	if remoteCentral.Metadata.DeletionTimestamp != "" {
-		return r.reconcileInstanceDeletion(ctx, &remoteCentral, central)
+		status, err := r.reconcileInstanceDeletion(ctx, &remoteCentral, central)
+		shouldUpdateCentralHash = err == nil
+		return status, err
 	}
 
 	namespaceLabels := map[string]string{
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -146,16 +154,29 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
 	}
-	needsReconcile := r.needsReconcile(changed, central, remoteCentral.Metadata.SecretsStored)
+
+	shouldUpdateCentralHash := false
+	defer func() {
+		if shouldUpdateCentralHash {
+			r.lastCentralHash = centralHash
+			r.lastCentralHashTime = time.Now()
+		} else {
+			r.lastCentralHash = [16]byte{}
```
SimonBaeumer, line 173: When the Central hash should not be updated, delete it? I would have assumed it just doesn't set the last central hash instead of removing it.

SimonBaeumer (approved):
Nice job on the testing site and debugging :rocket:

fleetshard/pkg/central/reconciler/reconciler.go
ludydoo: That was the source of the bug, because when the deployment is not ready, the hash is not updated, even though the central CR has been updated.

PR #1637, opened 2024-02-06
Improve logging and error handling

ludydoo:
Improving logging and error handling
- converting `fmt.Errorf("...: %w)` to `errors.Wrap`
- prefixing reconciler logs with `central 'id' (<namespace>/<name>): `
- quoting names/namespaces in errors and logs 
- use `logf("%v", key)` instead of `logf("%s/%s", namespace, name)`
- in errors, use `failed to... ` instead of `<action>`. Eg. `failed to get secret` instead of `getting secret`


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -156,14 +155,16 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
 	}
 	defer atomic.StoreInt32(r.status, FreeStatus)
 
+	namespace := remoteCentral.Metadata.Namespace
+
 	centralHash, err := r.computeCentralHash(remoteCentral)
 	if err != nil {
-		return nil, errors.Wrap(err, "computing central hash")
+		return nil, errors.Wrap(err, "failed to compute central hash")
```
johannes94, line 162: I've checked stackrox/stackrox and it looks like in general people are using both methods with the `<action>ing` and `failed to...` I like the latter more so no objection for this change.

About the errors.Wrap. I'm ok with the change but I think we might not need to wrap everything.

IMHO a good practice should be to wrap errors returned from other package functions or methods.

If we wrap in every error we see, and split into multiple functions we might end up with huge stack traces which make debugging harder for example:

```
func wrapsOnce(err error) error {
	return errors.Wrap(err, "wrapsOnce here")
}

func wrapsTwice(err error) error {
	e := wrapsOnce(err)
	return errors.Wrap(e, "wrapsTwice here")
}

func TestWrappyWrap(t *testing.T) {
	err := fmt.Errorf("bla bli blub %s", "didub")
	e1 := wrapsTwice(err)

	t.Errorf("Error 1: %+v", e1)
}
```

Would print:

```
    /Users/johannes/go/src/github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler/reconciler_test.go:67: Error 1: bla bli blub didub
        wraps1 here
        github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler.wrapsOnce
        	/Users/johannes/go/src/github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler/reconciler_test.go:53
        github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler.wrapsTwice
        	/Users/johannes/go/src/github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler/reconciler_test.go:57
        github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler.TestWrappyWrap
        	/Users/johannes/go/src/github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler/reconciler_test.go:64
        testing.tRunner
        	/usr/local/go/src/testing/testing.go:1576
        runtime.goexit
        	/usr/local/go/src/runtime/asm_amd64.s:1598
        wrapsTwice here
        github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler.wrapsTwice
        	/Users/johannes/go/src/github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler/reconciler_test.go:58
        github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler.TestWrappyWrap
        	/Users/johannes/go/src/github.com/stackrox/acs-fleet-manager/fleetshard/pkg/central/reconciler/reconciler_test.go:64
        testing.tRunner
        	/usr/local/go/src/testing/testing.go:1576
        runtime.goexit
        	/usr/local/go/src/runtime/asm_amd64.s:1598
```

Notice the 2 stack traces in the log above. If we add to many nested wraps we'll quickly end with java like stack traces of thousands of lines with only the top 20 lines being relevant.

ludydoo: I see your point. Though we currently have a mix of wrapping and not-wrapping same-package errors. E.g. 

```
	if err := r.ensureChartResourcesExist(ctx, remoteCentral); err != nil {
		return nil, errors.Wrapf(err, "unable to install chart resource for central %s/%s", central.GetNamespace(), central.GetName())
	}

	if err = r.reconcileCentralDBConfig(ctx, &remoteCentral, central); err != nil {
		return nil, err
	}
```

Would you suggest to never wrapping same-package errors and always wrapping other-package errors? But if an error is thrown in the same method via 2 different callers, we wouldn't be able to distinguish how the error was actually thrown (from where).

johannes94: Hmm yes, I noticed that inconsistency. 
 
 I guess it makes sense to either wrap in the same package or use `errors.Errorf` when creating a new error so that we also have stack traces for "in package" errors.
 
Given your second argument it seems reasonable to leave it as is for now and re-asses once we think to long stack traces are a problem when debugging.

ludydoo: You were right about same-package errors. I thought that the stacktrace would only be visible if the errors is wrapped at every level. I wrongly thought that if error is thrown in a nested subfunction, that we couldn't see the stacktrace for the whole "chain". But a quick test:


```

func main() {
	err := func1()
	fmt.Printf("%+v", err)
}

func func1() error {
	return func2()
}

func func2() error {
	return func3()
}

func func3() error {
	return errors.Wrap(errors.New("error message"), "err", )
}
```

shows

```
error message
main.func3
        scratch_19.go:22
main.func2
        scratch_19.go:18
main.func1
        scratch_19.go:14
main.main
        scratch_19.go:9
runtime.main
        proc.go:250
runtime.goexit
        asm_arm64.s:1172
err
main.func3
        scratch_19.go:22
main.func2
        scratch_19.go:18
main.func1
        scratch_19.go:14
main.main
        scratch_19.go:9
runtime.main
        proc.go:250
runtime.goexit
        asm_arm64.s:1172
```

(I thought that the stacktrace would only include `func3`.

PR #1677, opened 2024-02-22
ROX-22557: Count `expired-at` non-null changes

0x656b694d:
## Description

A new Fleet-manager Prometheus metric, showing the total number of expired centrals.
The idea is to create an alert on a spike of such events.

## Test manual

Initial run:

```console
$ curl -s http://localhost:8080/metrics | grep expired
# HELP acs_fleet_manager_expired_centrals_count total number of expired centrals
# TYPE acs_fleet_manager_expired_centrals_count gauge
acs_fleet_manager_expired_centrals_count 0
```

1. Created locally an eval central, managed by quota-management-list;
2. Patched quota-management-list config to allow 0 instances;
3. Restarted fleet-manager to trigger expiration worker;
4. Checked the fleet-manager metrics:

    ```console
    curl -s http://localhost:8080/metrics | grep expired
    # HELP acs_fleet_manager_expired_centrals_count total number of expired centrals
    # TYPE acs_fleet_manager_expired_centrals_count gauge
    acs_fleet_manager_expired_centrals_count 1
    ```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -740,6 +740,7 @@ func init() {
 	prometheus.MustRegister(centralOperationsTotalCountMetric)
 	prometheus.MustRegister(centralStatusSinceCreatedMetric)
 	prometheus.MustRegister(CentralStatusCountMetric)
+	prometheus.MustRegister(centralExpirationSetCountMetric)
 
 	// metrics for reconcilers
 	prometheus.MustRegister(reconcilerDurationMetric)
@@ -808,6 +809,7 @@ func Reset() {
 	centralOperationsTotalCountMetric.Reset()
 	centralStatusSinceCreatedMetric.Reset()
 	CentralStatusCountMetric.Reset()
+	centralExpirationSetCountMetric.Reset()
 
 	reconcilerDurationMetric.Reset()
 	reconcilerSuccessCountMetric.Reset()
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -27,6 +27,7 @@ const (
 	LabelStatus            = "status"
 	LabelClusterID         = "cluster_id"
 	LabelClusterExternalID = "external_id"
+	LabelOrganisationID    = "organisation_id"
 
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -326,6 +330,20 @@ func UpdateCentralPerClusterCountMetric(clusterID string, clusterExternalID stri
 	centralPerClusterCountMetric.With(labels).Set(float64(count))
 }
 
+// create a new counterVec for when expiration timestamp is set to a central
+var centralExpirationSetCountMetric = prometheus.NewCounterVec(
```
stehessel, line 334: Why do you use a counter instead of a gauge? A counter will not decrease if instances are "un-expired" again.

stehessel (commented):
Are you trying to define a metric for how many instances are expired in general? Or a metric that tracks the expired status of instances individually (presumably to detect flapping)?

0x656b694d:
> Are you trying to define a metric for how many instances are expired in general? Or a metric that tracks the expired status of instances individually (presumably to detect flapping)?

@stehessel, the plan is to define an alert on high rate of such expiration events, to spot potential mass extinction. The value of the metric doesn't tell much.

Implementing a metric, that counts the actual number of expired instances, would require a special process to periodically query DB specifically for that. Would you suggest that instead?

pkg/metrics/metrics.go
0x656b694d: Because this metric does not count the number of expired instances.

0x656b694d:
@stehessel, I changed the code to count the total number of expired centrals. This might make more sense indeed.

pkg/metrics/metrics.go
0x656b694d: No that's a different metric.

stehessel:
> Implementing a metric, that counts the actual number of expired instances, would require a special process to periodically query DB specifically for that. Would you suggest that instead?

Yes from the metrics perspective this makes more sense imo. The issue with the previous approach is that the metrics will have a high cardinality. One time series per instance + probe instances will generate a lot of metric series.

PR #1690, opened 2024-03-05
ROX-22887: Use go 1.20 to build images in CI

ebensh:
Context is in https://issues.redhat.com/browse/ROX-22887

If we don't pin the version to 1.20 in the build step it will break when the runner tries to use 1.21, because our go.mod file is missing the new toolchain directive (see https://tip.golang.org/doc/toolchain).

When we upgrade to using go 1.21 we'll just need to run `go mod tidy` once, commit the changes, and update our CI runners to use go 1.21.


porridge (approved):
❤️

PR #1708, opened 2024-03-12
ROX-22361: Log ACSCS instance state change

ivan-degtiarenko:
## Description

This PR does its best to keep track of central request state change. 
Conversion to string is done so that secrets, routes and auth config are not logged.

`DeprovisionExpiredDinosaurs` is not covered here since query does not return affected rows.

## Test manual

1. TBD

### DIFF
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -694,7 +694,7 @@ func (k *dinosaurService) Update(dinosaurRequest *dbapi.CentralRequest) *errors.
 		Model(dinosaurRequest).
 		Where("status not IN (?)", dinosaurDeletionStatuses) // ignore updates of dinosaur under deletion
 
-	logStateChange("delete request", dinosaurRequest.ID, dinosaurRequest)
+	logStateChange("updates", dinosaurRequest.ID, dinosaurRequest)
 
 	if err := dbConn.Updates(dinosaurRequest).Error; err != nil {
 		return errors.NewWithCause(errors.ErrorGeneral, err, "Failed to update central")
--- a/internal/dinosaur/pkg/services/dinosaur.go
+++ b/internal/dinosaur/pkg/services/dinosaur.go
@@ -709,7 +709,7 @@ func (k *dinosaurService) Updates(dinosaurRequest *dbapi.CentralRequest, fields 
 		Model(dinosaurRequest).
 		Where("status not IN (?)", dinosaurDeletionStatuses) // ignore updates of dinosaur under deletion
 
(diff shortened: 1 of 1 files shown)


internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -181,6 +181,8 @@ func (k *dinosaurService) RotateCentralRHSSOClient(ctx context.Context, centralR
 func (k *dinosaurService) ResetCentralSecretBackup(ctx context.Context, centralRequest *dbapi.CentralRequest) *errors.ServiceError {
 	centralRequest.Secrets = nil // pragma: allowlist secret
 
+	glog.Infof("instance %q state change: reset secrets", centralRequest.ID)
```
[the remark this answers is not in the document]
ivan-degtiarenko: > let's create a function that will log state change, so that we have one place where format is defined. We would need something like:

Done

>it would be good to "glue" (or couple) log calls to one type of internal call. In some cases, logs are called before the actual database update, and in other cases, they are part of the handler function. I would suggest calling logs just before database update calls if that is possible. In that case, it is easier to review, and we are also safer that any DB change is actually logged as a state change.

Unfortunately `dbConn` is from `gorm` and cannot be modified that easily (we can write a wrapper but that would be a lot of extra work). Also, methods like `dbConn.Updates` have very generic parameter types that we would need to cast back to `CentralRequest`.
With the information we want to log, `dinosaur.go` service looks like a proper place.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -689,6 +694,8 @@ func (k *dinosaurService) Update(dinosaurRequest *dbapi.CentralRequest) *errors.
 		Model(dinosaurRequest).
 		Where("status not IN (?)", dinosaurDeletionStatuses) // ignore updates of dinosaur under deletion
 
+	logStateChange("delete request", dinosaurRequest.ID, dinosaurRequest)
```
mtodor, line 697: I assume that the message should be adjusted. It's probably copy/paste from the previous one.

internal/dinosaur/pkg/services/dinosaur.go
```diff
@@ -702,6 +709,8 @@ func (k *dinosaurService) Updates(dinosaurRequest *dbapi.CentralRequest, fields
 		Model(dinosaurRequest).
 		Where("status not IN (?)", dinosaurDeletionStatuses) // ignore updates of dinosaur under deletion
 
+	glog.Infof("instance %q state change: %+v", dinosaurRequest.ID, fields)
```
mtodor, line 712: We should use `logStateChange`.

PR #1709, opened 2024-03-12
Backup manual declarative config

ivan-degtiarenko:
## Description
This PR adds manual declarative config to a list of watched secrets for backup. In case manual declarative config is missing, `fleetshard-sync` skips backing it up.

## Test manual

1. TBD

### DIFF
--- a/fleetshard/pkg/k8s/secret.go
+++ b/fleetshard/pkg/k8s/secret.go
@@ -11,27 +11,29 @@ import (
 )
 
 const (
-	CentralTLSSecretName           = "central-tls"            // pragma: allowlist secret
-	centralDBPasswordSecretName    = "central-db-password"    // pragma: allowlist secret
-	centralEncryptionKeySecretName = "central-encryption-key" // pragma: allowlist secret
+	CentralTLSSecretName              = "central-tls"                              // pragma: allowlist secret
+	centralDBPasswordSecretName       = "central-db-password"                      // pragma: allowlist secret
+	centralEncryptionKeySecretName    = "central-encryption-key"                   // pragma: allowlist secret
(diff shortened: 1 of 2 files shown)


fleetshard/pkg/k8s/secret.go
```diff
@@ -11,36 +11,40 @@ import (
...
 	}
 
 	return &SecretBackup{client: client, secretsToWatch: secretsToWatch} // pragma: allowlist secret
 }
 
 // GetWatchedSecrets return a sorted list of secrets watched by this package
 func (s *SecretBackup) GetWatchedSecrets() []string {
-	secrets := make([]string, len(s.secretsToWatch))
-	copy(secrets, s.secretsToWatch)
+	secrets := make([]string, 0, len(s.secretsToWatch))
```
johannes94, line 44: Can you elaborate on why this implementation is better than the other one using `copy`?

johannes94 (changes requested):
We have an issue with the `areSecretsStored` function of the reconciler here.

https://github.com/stackrox/acs-fleet-manager/blob/129a4ce2b9ac8b363211adfa7ef8510935e09be2/fleetshard/pkg/central/reconciler/reconciler.go#L850-L867

`expectedSecrets` would always be all secrets required and non-required ones. While this is ok for tenants that have all secrets. It would lead to the reconciler cache being skipped and the reconciler reporting all secrets on every reconciliation for clients without the optional secret.

For those clients `expectedSecrets` would never equal `secretsStored` because it would never store the non-existent optional secret.

Actually solving that isn't exactly trivial, because you'd need to know which secret are there on every reconciliation. I think we have a uncached k8s client. This might lead to a lot of API requests to the kube API server, especially on startup of FS. And you can't use the reconciliation cache to limit those requests because it's a condition for the cache.

PR #1763, opened 2024-04-15
ROX-23559: Export Violation Denial Metrics

mclasmeier:
## Description

This PR adds settings for the cloudwatch exporter to also export our custom violation denials metrics. This includes the metric for SELinux AVC denials and the metric for NetworkPolicy ACL denials.

## Test manual

```
# To run tests locally run:
make ocm/setup
```

### DIFF
--- a/dp-terraform/helm/rhacs-terraform/charts/cloudwatch/templates/01-operator-03-config-map.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/cloudwatch/templates/01-operator-03-config-map.yaml
@@ -51,3 +51,16 @@ data:
             - name: Bounce
             - name: Reputation.BounceRate
             - name: Reputation.ComplaintRate
+    static:
+      - name: selinux_avc_denials
+        namespace: SELinux
+        regions:
+          - eu-west-1
+          - us-east-1
+        metrics:
+          - name: AVCDenial
+            nilToZero: true
+            statistics:
+              - SampleCount
+            period: 300
+            length: 300
--- a/dp-terraform/helm/rhacs-terraform/charts/cloudwatch/templates/01-operator-03-config-map.yaml
+++ b/dp-terraform/helm/rhacs-terraform/charts/cloudwatch/templates/01-operator-03-config-map.yaml
@@ -58,7 +58,19 @@ data:
           - eu-west-1
           - us-east-1
         metrics:
-          - name: AVCDenial
+          - name: AVCDenials
+            nilToZero: true
+            statistics:
+              - SampleCount
+            period: 300
+            length: 300
+      - name: network_policy_acl_denials
+        namespace: NetworkPolicy
+        regions:
+          - eu-west-1
+          - us-east-1
+        metrics:
+          - name: ACLDenials
             nilToZero: true
             statistics:
               - SampleCount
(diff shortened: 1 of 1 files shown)


dp-terraform/helm/rhacs-terraform/charts/cloudwatch/templates/01-operator-03-config-map.yaml
```diff
@@ -51,3 +51,28 @@ data:
             - name: Bounce
             - name: Reputation.BounceRate
             - name: Reputation.ComplaintRate
+    static:
+      - name: selinux_avc_denials
+        namespace: SELinux
+        regions:
+          - eu-west-1
```
ebensh, line 58: This is interesting - we use the same template for all data plane clusters, so the RDS section above is going to report everything, regardless of which data plane cluster it's in.

I guess the "right" solution would be to filter down the exported metrics to just the relevant *cluster*, so that each instance of the cloudwatch exporter is only monitoring its statistics. Or have a global monitoring/alerting dashboard for things that are already aggregated like CloudWatch, but that fell apart.

That said, given our small number of regions and clusters, it's probably fine to leave it like this for now. It means that acs-prod-dp-01 and acs-prod-eu-01 (and future acs-prod-dp-02, etc.) would all alert on an SELinux violation occurring in acs-prod-dp-01. We could mitigate it in AlertManager by at least filtering down to the region... or I wonder, maybe we could do the filtering then in Prometheus? Maybe the metric would have a structured key that could be pulled, including the source data plane cluster. Maybe that's the way it already is.

Have you seen the alerting configuration based on this exported information? I'd be curious to know if it's cluster or region specific.

mclasmeier: Not familiar with the semantics yet, but have you seen https://github.com/stackrox/acs-fleet-manager/pull/1763/files/5bbe64225d38dba26bd7461e26ca641b61ff126c#diff-f211b944cbbea7383ed888157a60e007d785e28460edb9f14106252f8e51a555L15?

ebensh: That's more for dynamic discovery of AWS resources, I think. It's grepping through the AWS account for RDS entries with the "DataplaneClusterName" tag set, no?

This would be different - more like extracting the log_line.ClusterName field from the audit log statement, then exporting that as a Dimension(?) - see https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/ExtractBytesExample.html - and having the CloudWatch Exporter export the cluster name to Prometheus as a "label" on the metric.

Then each CloudWatch Exporter would export all the metrics - but we could at least filter the alert down to just this data plane cluster in the alert definition, for example.

mclasmeier: I might be misunderstanding something here, but my remark was meant as a direct response to this:

> we use the same template for all data plane clusters, so the RDS section above is going to report everything, regardless of which data plane cluster it's in.

So, yes, it was RDS specific, but it was my understanding of the code that it is exactly there to restrict the metric export to a specific cluster.

ebensh (approved):

PR #1767, opened 2024-04-16
ROX-23550: Add NetworkPolicy to Tenant Helm chart

ebensh:
## Description
Add NetworkPolicy's to the Fleetshard Sync Tenant Helm chart. Their creation is guarded by the new secureTenantNetwork configuration option.

These policies effectively replace and improve the ACS Operator's default NetworkPolicy's.

Where possible, I have tried to optimize per https://docs.openshift.com/container-platform/4.14/networking/network_policy/about-network-policy.html#nw-networkpolicy-optimize-ovn_about-network-policy

Document discussing the changes: https://docs.google.com/document/d/15-rlsU_wDsv6TX_2hKHNFir82W1T9Ov1SOnRSha1VUY/edit

## Test manual

```
# To run tests locally run:
make ocm/setup
```


fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,171 @@
...
+          kubernetes.io/metadata.name: rhacs-observability
+    ports:
+    - port: 9090
+      protocol: TCP
+    - port: 9091
+      protocol: TCP
+  egress:
+  - to:  # Allow egress to RDS subnet
+    - ipBlock:
+      cidr: 10.1.0.0/16
```
kylape, line 71: this should be indented once to be under the `ipBlock` section.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,171 @@
...
+        matchLabels:
+          app: scanner
+    ports:
+    - port: 8080
+      protocol: TCP
+    - port: 8443
+      protocol: TCP
+  - to:  # Allow egress to external Internet
+    - ipBlock:
+      cidr: 0.0.0.0/0
```
kylape, line 86: same indentation comment

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,171 @@
...
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: rhacs-observability
+    ports:
+    - port: 9090
+      protocol: TCP
+    - port: 9091
+      protocol: TCP
+  egress:
+  - to:  # Allow egress to Scanner-db
```
kylape, line 122: does scanner v2 reach out to central for updated definitions?

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
kylape, line 71: It might be worth a comment here to note that the `EgressFirewall` blocks this CIDR block except for the specific IP that resolves from the tenant's RDS hostname.

kylape (commented):
Did you ever find out if we can disable network policy creation in the operator?  Also, do we need scanner v4 network policies, too?

I'll try to applly these network policies later today and see if I can test them out.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
ebensh: Done

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
ebensh: Asked @RTann for clarity on this one. It's a good point - before Central is open to any ingress on port 8443, but I didn't check what other jobs could be calling out to it. Thanks!

ebensh: Yes, it does. Added in the newest commit. Marking this one as done.

ebensh:
> Did you ever find out if we can disable network policy creation in the operator?

Yes I checked, no we can't, but as discussed in sync call there's a way to roll out a GitOps override while we add the flag (not blocked by ACS release).

> Also, do we need scanner v4 network policies, too?

I was going to add them in a future PR but have rolled them in here too. PTAL.

> I'll try to applly these network policies later today and see if I can test them out.

Sounds good.

kovayur (approved):

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,294 @@
...
+  - to:  # Allow egress to RDS subnet, EgressFirewall will limit to its specific instance because the IP address is not static and EgressFirewall can do DNS resolution
+    - ipBlock:
+        cidr: 10.1.0.0/16
+    ports:
+      - port: 5432
+        protocol: TCP
+  - to:  # Allow egress to Scanner
+    - podSelector:
+        matchLabels:
+          app: scanner
```
vladbologa, line 83: Are you sure that it doesn't need egress to scanner-v4 components?

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,294 @@
...
+  - from:  # Allow ingress from observability to scrape metrics
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: rhacs-observability
+    ports:
+    - port: 9090
+      protocol: TCP
+    - port: 9091
+      protocol: TCP
+  egress:
```
vladbologa, line 73: I think we also need to allow egress to:
- k8s API (kubernetes.default.svc.cluster.local.:443)
- audit logging, the URL is configurable, see https://github.com/stackrox/acs-fleet-manager/blob/76f1b5e69be1feb11350d6c2b4c89352a01e7cde/fleetshard/config/config.go#L64

vladbologa (changes requested):
In addition to my other comments, I think that at:

https://github.com/stackrox/acs-fleet-manager/blob/68166244938ba8b8e2f5d7c678cb158a396103b2/fleetshard/pkg/central/reconciler/reconciler.go#L313

We should have something like:
```golang
if !r.secureTenantNetwork {
    r.applyProxyConfig(central)
}
```

This applies if this PR is merged before https://github.com/stackrox/acs-fleet-manager/pull/1768. Else Central traffic will go through the egress-proxy, which is not allowed by these network policies.

It's also not clear to me what happens if `secureTenantNetwork` is enabled while the operator still deploys its own NetworkPolicies.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
ebensh: Done.

ebensh: Added

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
ebensh: Added audit logging to central egress (port 8888).

Having troubles with the k8s API - it looks like the IPs there are not stable. We could add open its *range* (if it's clear) and then block in the EgressFirewall, as an option.

See e.g. https://pauldally.medium.com/accessing-kubernetes-api-server-when-there-is-an-egress-networkpolicy-af4435e005f9

ebensh:
> In addition to my other comments, I think that at:
> 
> https://github.com/stackrox/acs-fleet-manager/blob/68166244938ba8b8e2f5d7c678cb158a396103b2/fleetshard/pkg/central/reconciler/reconciler.go#L313
> 
> We should have something like:
> 
> ```go
> if !r.secureTenantNetwork {
>     r.applyProxyConfig(central)
> }
> ```
> 
> This applies if this PR is merged before #1768. Else Central traffic will go through the egress-proxy, which is not allowed by these new network policies.

Good point; added in latest commit.

> It's also not clear to me what happens if `secureTenantNetwork` is enabled while the operator still deploys its own NetworkPolicies.

The policies are strictly additive. The main concern is that the ACS Operator doesn't deploy *any* egress policies. This means that once the Fleetshard Sync starts writing its policies, all egress *not* explicitly specified will be denied.

I have other PRs to:
1. Use gitops to effectively null out the ACS Operator's NetworkPolicy spec's, which will help confirm our policies here are sufficient, and
2. Add the secureTenantNetwork flag to the OCM Addon flags for the Addons. This will let us control the rollout of the secureTenantNetwork by environment, so we can test the interactions in ACSCS Integration before it hits stage/prod. (Separated code behavior and config push)
3.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,312 @@
...
+  - to:  # Allow egress to RDS subnet, EgressFirewall will limit to its specific instance because the IP address is not static and EgressFirewall can do DNS resolution
+    - ipBlock:
+        cidr: 10.1.0.0/16
```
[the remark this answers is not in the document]
vladbologa: This is the CIDR of the VPC where we keep the tenant RDS databases. It's peered to the VPC of the OpenShift cluster.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
vladbologa: @knobunc We're having trouble figuring out how to allow access to the k8s API, can you please help?

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,312 @@
+# If zero NetworkPolicy's exist for Ingress or Egress, then all traffic is by
```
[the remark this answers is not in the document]
ebensh: Thank you for the clearer explanation - I've applied your suggestion and moved the text into the conditional template.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,312 @@
...
+# default *allowed*. By explicitly creating an *empty* policy for Ingress and
+# Egress targeting all pods in the namespace, we are saying "deny all traffic
+# unless another NetworkPolicy allows it".
+# Following https://docs.openshift.com/container-platform/4.14/networking/network_policy/about-network-policy.html#nw-networkpolicy-optimize-ovn_about-network-policy
+# also combine the deny-all with allow-dns policies.
+# Note that OpenShift has two: (internal) DNS and ExternalDNS.
+# TODO: Double check this allows KubeDNS, OpenShift Internal DNS, OpenShift External DNS
```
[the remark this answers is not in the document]
ebensh: Thanks for clarifying this. Simplified the text and removed the TODO.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,312 @@
...
+    ports:
+      - port: 443
+        protocol: TCP
+      - port: 6443
+        protocol: TCP
+  - to:  # Allow egress to external Internet
+    - ipBlock:
+        cidr: 0.0.0.0/0
+      except:
+        {{- include "localNetworkCidrRanges" . | nindent 8 }}
```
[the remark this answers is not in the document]
ebensh: > Is this an attempt to prevent access to pods?

Yes. Part of Central's functionality is its integrations - we want it to be able to make outgoing connections to external services. At the same time, we don't want Central (via malicious configuration of integrations, for example) to be able to connect to our own internal network.

I was hoping there was a way to represent a "network egress" similar to the network ingress and create a network policy allowing only that, but I couldn't find one. This attempts to invert and allow everything except internal access.

> We may need to think this through a little more because people could set the cluster up on their own ip space and not use a reserved one.

To be clear the customer cannot set up *this* cluster, we own and operate it. Given that, is this still a concern?

> Or want to talk to something in their corporate network in a reserved network.

Is this possible? How would they have an IP address in those ranges registered with DNS?

ebensh:
To be explicit I'm waiting on @vladbologa and @knobunc approval before merging :)

Looking at the CI to try to figure out if it's just flakes or a test I missed updating now.

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,311 @@
...
+            - { key: app, operator: In, values: [scanner, scanner-v4-indexer, scanner-v4-matcher] }
+    ports:
+    - port: 8080
+      protocol: TCP
+    - port: 8443
+      protocol: TCP
+  - to:  # Allow egress to audit-logs-aggregator
+      - podSelector:
+          matchLabels:
+            app=audit-logs-aggregator
```
vladbologa, line 93:
```suggestion
            app: audit-logs-aggregator
```

vladbologa (changes requested):

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
ebensh: Applied

vladbologa (approved):
LGTM!

One more thing, what do you think about surrounding `fleetshard/pkg/central/charts/data/tenant-resources/templates/egress-proxy.yaml`
with
`
{{- if not .Values.secureTenantNetwork }}
{{ end }}
`
so that we don't even install the egress-proxy and its network policies when `secureTenantNetwork` is enabled?

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
```diff
@@ -0,0 +1,314 @@
...
+        protocol: TCP
+  - to:  # Allow egress to Kube API for debug snapshot generation, mTLS setup for OpenShift monitoring, telemetry
+    - podSelector:
+        matchLabels:
+          app: openshift-kube-apiserver
+    ports:
+      - port: 443
+        protocol: TCP
+      - port: 6443
+        protocol: TCP
```
vladbologa, line 108: Are these rules OK? Without a namespace selector, I think that it would allow traffic to `audit-logs-aggregator` and `openshift-kube-apiserver` only if they are in the same namespace as the tenant, which is not the case.

vladbologa (changes requested):

fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy.yaml
ebensh: Thanks for catching this; yes. I audited the other `to:` stanzas as well to make sure they checked out.

fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
```diff
@@ -9,3 +9,5 @@ egressProxy:
       memory: 275Mi
 labels: {}
 annotations: {}
+secureTenantNetwork: false
```
vladbologa, line 12: Instead of hardcoding this value, you could set the value from the code here fleetshard/pkg/central/reconciler/reconciler.go:1730

And so you'd propagate the value from dp-terraform.

Otherwise this remains false when you set `secureTenantNetwork: true` via the Addon.

vladbologa (changes requested):

fleetshard/pkg/central/charts/data/tenant-resources/values.yaml
ebensh: Is this not what https://github.com/stackrox/acs-fleet-manager/pull/1767/files#diff-8b2c412ebc3fa158d46c2dfb0136cfef521abb4fb2dda0a77b403a32afd9c58eR1744 is doing?

vladbologa: Yes it is, I'll see myself out. :)

vladbologa (approved):

PR #1768, opened 2024-04-16
ROX 23555: Remove egress-proxy

ebensh:
## Description
Removes the egress-proxy, as it has been replaced by NetworkPolicies.

## Test manual

```
# To run tests locally run:
make ocm/setup
```


fleetshard/pkg/central/reconciler/testdata/tenant-resources/templates/dummy.yaml
```diff
@@ -1,5 +1,5 @@
-apiVersion: v1
-kind: Service
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
```
vladbologa, line 2: Had to change this from `Service` to `NetworkPolicy` because the tenant resource garbage collection mechanism no longer manages objects of type `Service` (see the change to `zzz_managed_resources.go`

fleetshard/config/config.go
```diff
@@ -35,7 +35,6 @@ type Config struct {
 	ServiceAccountTokenFile string        `env:"FLEET_MANAGER_TOKEN_FILE"`
 	CreateAuthProvider      bool          `env:"CREATE_AUTH_PROVIDER" envDefault:"false"`
 	MetricsAddress          string        `env:"FLEETSHARD_METRICS_ADDRESS" envDefault:":8080"`
-	EgressProxyImage        string        `env:"EGRESS_PROXY_IMAGE"`
```
ebensh, line 38: Will deleting this remove the ENV flag and break if we try to set it? Or will the unknown flag be silently ignored?

fleetshard/pkg/central/reconciler/testdata/tenant-resources/templates/dummy.yaml
ebensh: I don't think this will be a problem, but we do need to be careful - this isn't conditional on setting the secureTenantNetwork flag. If we disabled that flag, and this resource exists, all network traffic will be blocked for all tenants.

Can we use a non-NetworkPolicy resource? A ConfigMap or Secret?

vladbologa: Due to the way tenant resource deletion works, we can only use types that are mentioned in `zzz_managed_resources.go`. The types in that file are auto-generated by parsing everything in `fleetshard/pkg/central/charts/data/tenant-resources`.

These types currently are `VerticalPodAutoscaler` and `NetworkPolicy`. If you use any other type, the objects will not get deleted and some tests will fail (e.g. `TestChartResourcesAreAddedAndRemoved`).

The tests themselves are not conditional on `secureTenantNetwork`. But if we will no longer have NetworkPolicies deployed as tenant resources, this dummy file will need to be fixed, and dependent tests too.

fleetshard/config/config.go
vladbologa: I think the order in which these are loaded is in reverse.

`.Values.fleetshardSync.egressProxy.image` in the fleetshard-sync template creates the env var. The value in the env var is then loaded here.

They were both removed, but the helm value is still set (for now) in the gitops config. I think helm will silently ignore that value.

ebensh: ACK; sounds good.

mclasmeier (approved):

PR #1781, opened 2024-05-03
ROX-23550:  Change DNS from kube-dns to openshift-dns

ebensh:


fleetshard/pkg/central/charts/data/tenant-resources/templates/network-policy-default.yaml
```diff
@@ -22,14 +22,8 @@ spec:
...
-              k8s-app: kube-dns
-      ports:
-        - port: 53
-          protocol: UDP
-        - port: 53
-          protocol: TCP
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-dns
```
ludydoo, line 28: No need for ports?

ludydoo (approved):

PR #1801, opened 2024-05-08
ROX-23709: Trust Data Plane OAuth issuers in fleetshard authorization middleware

kovayur:
## Description
This change adds Data Plane OAuth issuers validation in fleetshard authorization middleware.
The old authorization remains for backward compatibility

## Checklist (Definition of Done)

- [ ] Unit and integration tests added
- [ ] Added test description under `Test manual`
- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)
- [ ] CI and all relevant tests are passing
- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.
- [ ] Add secret to app-interface Vault or Secrets Manager if necessary
- [ ] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)
- [ ] Check AWS limits are reasonable for changes provisioning new resources
- [ ] (If applicable) Changes to the dp-terraform Helm values have been reflected in the addon on integration environment

## Test manual

**TODO:** Add manual testing efforts

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup
make verify lint binary test test/integration
```

PR #1789, opened 2024-05-06
ROX-23709: Refactor authentication handler

kovayur:
## Description
This change splits the authentication handler into 3 parts, each dedicated to a different API: private, public, admin.
The goal is to limit the number of trusted issuers for a particular API.

Every part has its own `issuerURIs` which allows to limit the issuers for every API:
- `private` handler represented by the new `DataplaneOIDCIssuers` configuration parameter;
- `admin` handler is configured with the `InternalSSORealm` issuer uri. 
- `default` handler serves the rest (excl. `InternalSSORealm`) 

`JwksURL` and `JwksFile` were moved to from `ServerConfig` to `IAMConfig` to avoid circular dependency between packages.

## Checklist (Definition of Done)

- [ ] Unit and integration tests added
- [ ] Added test description under `Test manual`
- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...)
- [ ] CI and all relevant tests are passing
- [ ] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [ ] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.
- [ ] Add secret to app-interface Vault or Secrets Manager if necessary
- [ ] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)
- [ ] Check AWS limits are reasonable for changes provisioning new resources
- [ ] (If applicable) Changes to the dp-terraform Helm values have been reflected in the addon on integration environment

## Test manual

**TODO:** Add manual testing efforts

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup
make verify lint binary test test/integration
```

PR #1802, opened 2024-05-08
ROX-23709: Load token from file in fleetshard-sync

kovayur:
## Description
This change adds a new `AuthType` called `TOKEN_FILE` in order to load a service account token for authentication in Fleet Manager. This type becomes the default value in the dp-terraform helm chart instead of `RHSSO`.

## Test manual

```
# To run tests locally run:
make ocm/setup
```


kovayur:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1802?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1802** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1802?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1801** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1801?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1789** <a href="https://app.graphite.dev/github/pr/stackrox/acs-fleet-manager/1789?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @kovayur and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -148,3 +154,8 @@ spec:
               - serviceAccountToken:
                   path: aws-token
                   audience: sts.amazonaws.com
+        - name: fleet-manager-token
+          projected:
+            sources:
+              - serviceAccountToken:
+                  path: fleet-manager-token
```
dhaus67, line 161: Let's also set a custom audience here (see [also this comment](https://github.com/stackrox/acs-fleet-manager/pull/1801#issuecomment-2107365258)) to have a) additional verification on FM side and b) avoid the token to be used to access the kube API.

dp-terraform/helm/rhacs-terraform/values.yaml
```diff
@@ -6,8 +6,8 @@ fleetshardSync:
   image:
     repo: "quay.io/app-sre/acs-fleet-manager"
     tag: "main"
-  # Can be either OCM, RHSSO, STATIC_TOKEN. When choosing RHSSO, make sure the clientId/secret is set. By default, uses RHSSO.
-  authType: "RHSSO"
+  # Can be either OCM, RHSSO, STATIC_TOKEN, TOKEN_FILE. When choosing RHSSO, make sure the clientId/secret is set. By default, uses TOKEN_FILE.
+  authType: "TOKEN_FILE"
```
dhaus67, line 10: nit: might be better to use service account token? It's a bit hard to really understand what is backing the "token file" auth type.

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -148,3 +154,8 @@ spec:
               - serviceAccountToken:
                   path: aws-token
                   audience: sts.amazonaws.com
+        - name: fleet-manager-token
```
dhaus67, line 157: This should only be done optional if the auth type is used (same applies for the mount + env variable).

pkg/client/fleetmanager/impl/auth.go
```diff
@@ -94,6 +100,11 @@ func NewStaticAuth(ctx context.Context, opt StaticOption) (Auth, error) {
 	return newAuth(ctx, staticTokenFactory.GetName(), Option{Static: opt})
 }
 
+// NewFileAuth will return Auth that uses a token file to provide authentication for HTTP requests.
+func NewFileAuth(ctx context.Context, opt TokenFileOption) (Auth, error) {
```
dhaus67, line 104: nit: if we keep the name token file auth, this should be named appropriately then (i.e. `NewTokenFileAuth`).

pkg/client/fleetmanager/impl/auth_file.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/stackrox/acs-fleet-manager/pkg/shared"
+)
+
+const (
+	// TokenFileAuthName is the name of the token file auth authentication method
+	TokenFileAuthName = "TOKEN_FILE"
```
dhaus67, line 14: Doesn't need to be exposed I _think_?

pkg/client/fleetmanager/impl/auth_file.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/stackrox/acs-fleet-manager/pkg/shared"
+)
+
+const (
+	// TokenFileAuthName is the name of the token file auth authentication method
```
dhaus67, line 13:
```suggestion
	// TokenFileAuthName is the name of the token file auth authentication method.
```

pkg/client/fleetmanager/impl/auth_file.go
```diff
@@ -0,0 +1,55 @@
...
+type tokenFileAuth struct {
+	file string
+}
+
+// GetName gets the name of the factory.
+func (f *tokenFileAuthFactory) GetName() string {
+	return TokenFileAuthName
+}
+
+// CreateAuth creates a new instance of Auth or returns error
```
dhaus67, line 34:
```suggestion
// CreateAuth creates a new instance of Auth or returns error.
```

pkg/client/fleetmanager/impl/auth_file.go
```diff
@@ -0,0 +1,55 @@
...
+func (f *tokenFileAuthFactory) CreateAuth(_ context.Context, o Option) (Auth, error) {
+	tokenFile := o.TokenFile.File
+	if _, err := os.Stat(tokenFile); err != nil {
+		return nil, fmt.Errorf("failed to read token file %s: %w", tokenFile, err)
```
dhaus67, line 38:
```suggestion
		return nil, fmt.Errorf("failed to read token file %q: %w", tokenFile, err)
```

pkg/client/fleetmanager/impl/auth_file.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	return &tokenFileAuth{
+		file: tokenFile,
+	}, nil
+}
+
+// AddAuth add auth token to the request retrieved from OCM.
+func (o *tokenFileAuth) AddAuth(req *http.Request) error {
+	token, err := shared.ReadFile(o.file)
```
dhaus67, line 48: It's a bit unfortunate that we greedily always read from file here, especially since we do send quite a few requests from fleet-sync to fleet-manager.

On the options to improve this, we can either a) use an expiring cache where the duration is a bit less than the 1 hour b) use the file watcher concept we have in stackrox to read updates on changes.

Either way, I don't think we should be reading on _every_ request here.

pkg/client/fleetmanager/impl/testdata/token
```diff
@@ -0,0 +1 @@
+some-value
```
dhaus67, line 1: super nit: might make sense to just take the sample JWT from `jwt.io` and paste it here to have a more "closer to reality" example.

pkg/client/fleetmanager/impl/auth_file.go
kovayur: The constants of other types are public too.
It can be used in `impl.NewAuth`

pkg/client/fleetmanager/impl/auth_file.go
kovayur: Fixed

pkg/client/fleetmanager/impl/auth_file.go
kovayur: Fixed

pkg/client/fleetmanager/impl/auth_file.go
kovayur: fixed

pkg/client/fleetmanager/impl/auth_file.go
kovayur: Expiring cache added

pkg/client/fleetmanager/impl/testdata/token
kovayur: Done.

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -69,6 +70,7 @@ spec:
               name: fleetshard-sync
               key: "rhsso-service-account-client-secret"
               optional: false
+        {{- end }}
         - name: RHSSO_REALM
           value: {{ .Values.fleetshardSync.redHatSSO.realm }}
         - name: RHSSO_ENDPOINT
```
dhaus67, line 76: I don't think we use realm and endpoint outside of the scope of the RHSSO auth type, so might as well put this under the same condition?

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
```diff
@@ -148,3 +158,11 @@ spec:
               - serviceAccountToken:
                   path: aws-token
                   audience: sts.amazonaws.com
+        {{- if eq "SERVICE_ACCOUNT_TOKEN" .Values.fleetshardSync.authType }}
+        - name: fleet-manager-token
+          projected:
+            sources:
+              - serviceAccountToken:
```
dhaus67, line 165: Sort of unrelated to the changes: while this is set to 3600 seconds (i.e. 1 hour); there are also other settings within the API server that can override the default max time for a token (e.g. if you install Central on infra openshift clusters, the default is set to 24 hours). Any idea what this is for our dataplane clusters? Not that we have the wrong understanding of the _actual_ expiration time of the token.

pkg/client/fleetmanager/impl/auth_service_account_token.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("reading token file %q: %w", a.file, err)
+	}
+	tokenString := strings.TrimSpace(string(tokenBytes))
+	if len(tokenString) == 0 {
+		return nil, fmt.Errorf("empty token file %q", a.file)
+	}
+	return &token{
+		raw:    tokenString,
+		expiry: a.now().Add(a.period),
```
dhaus67, line 101: Instead of having the period set statically, we can also parse the JWT token and get the expiration seconds from it and calculate the expiry this way. It would be more optiomal compared to the additional reads we currently do (but definitely  a bit more complets) - WDYT?

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
kovayur: Done, thanks

dp-terraform/helm/rhacs-terraform/templates/fleetshard-sync.yaml
kovayur: I explicitly set the duration to 3600 seconds

pkg/client/fleetmanager/impl/auth_service_account_token.go
kovayur: I like the idea, however I think the current solution is more reliable: if we parse the token and something goes wrong as we expect, the implementation will degrade to reading a token from the file every time. The examples are:

- The file is not refreshed at 80% of duration (this is the most concerning, since we don't control this);
- Can't parse the token: format changed, claims missing, etc. 

In current implementation we just refresh every 6 minutes, enough for self-recovery, but not too frequent.

dhaus67 (approved):

kovayur:
Had to rollback the default value in the helm chart due to the issues on the local / CI environment. Will follow-up in the next PR

PR #1819, opened 2024-05-21
ROX-24273: garbage collection for tenant-resources chart

ludydoo:

## Test manual

```
# To run tests locally run:
make ocm/setup
```


fleetshard/pkg/central/reconciler/gen/main.go
```diff
@@ -0,0 +1,180 @@
...
+
+func getChartDir() string {
+	return path.Join(path.Dir(getCurrentFile()), "../../charts/data/tenant-resources/templates")
+}
+
+func getReconcilerDir() string {
+	return path.Join(path.Dir(getCurrentFile()), "..")
+}
+
+func generate() error {
```
vladbologa, line 48: My impression is that generating the known GVKs is an optimization (to not go through a lot of objects that are surely not "chart resources" managed by fleetshard), and not something necessarily required. Is my understanding correct?

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -1558,55 +1569,118 @@ func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remot
...
 
+	// Perform garbage collection
+	for _, gvk := range tenantChartResourceGVKs {
+		gvk := gvk
+		var existingObjects unstructured.UnstructuredList
+		existingObjects.SetGroupVersionKind(gvk)
+
+		if err := r.client.List(ctx, &existingObjects,
+			ctrlClient.InNamespace(remoteCentral.Metadata.Namespace),
+			ctrlClient.MatchingLabels{helmChartNameLabel: r.resourcesChart.Name()},
```
vladbologa, line 1606: maybe also check for `labelManagedByFleetshardValue`?

vladbologa (commented):
This is very useful, I think it will also fix [ROX-13199](https://issues.redhat.com/browse/ROX-13199), right?

fleetshard/pkg/central/reconciler/gen/main.go
ludydoo: Yes. We either need to manually maintain the GVKs of objects defined in the chart, or generate them. Otherwise, we would need to enumerate every possible ApiVersion/Kind combination, and list all objects. That would become very expensive

ludydoo:
> This is very useful, I think it will also fix [ROX-13199](https://issues.redhat.com/browse/ROX-13199), right?

Yes it would fix it for the tenant-resources chart, not the operator charts

fleetshard/pkg/central/reconciler/gen/main.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		fileStr := string(fileBytes)
+		lines := strings.Split(fileStr, "\n")
+		for i := 0; i < len(lines)-1; i++ {
+			line1 := lines[i]
+			line2 := lines[i+1]
+			if line1 > line2 {
+				line1, line2 = line2, line1
+			}
```
vladbologa, line 71: You make the assumption that `apiVersion` and `kind` are always in adjacent lines, and might be in the "wrong" order. But it'll miss something like (which could theoretically be possible, right?):
```yaml
apiVersion:
metadata:
kind:
```

Not sure if it's worth the effort to fix it. One simple solution would be to  just look for `apiVersion`, and check that `kind` is on the line before or after. If not, return an error, and the developer who introduced the changes that caused this could fix the yaml code.

Another solution would be to go through all the lines, after we find both `apiVersion` and `kind` mark them as seen. Reset when encountering `---`.

fleetshard/pkg/central/reconciler/gen/main.go
```diff
@@ -0,0 +1,180 @@
...
+			matches := rgx.FindStringSubmatch(line)
+			if len(matches) != 4 {
+				continue
+			}
+			kind := matches[3]
+			group := matches[1]
+			version := matches[2]
+			fmt.Printf("found existing resource: %s/%s/%s\n", kind, group, version)
+			key := [3]string{kind, group, version}
+			seen[key] = seen[key]
```
vladbologa, line 132: Did you mean

```suggestion
			seen[key] = true
```

fleetshard/pkg/central/reconciler/gen/main.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+
+	sorted := make([][3]string, 0, len(seen))
```
vladbologa, line 136: nit: could split this entire function into 3 functions:
- gather the GVKs from the yaml files
- load the existing file
- write the new file

vladbologa (changes requested):

fleetshard/pkg/central/reconciler/gen/main.go
ludydoo: No, this part of the code parses GVKs declared in the existing `zzz_managed_resources.go`. 

If `seen[key]` is true, this means that the resource is present in the chart. 
If `seen[key]` is false, this means that the resource is declared in `zzz_managed_resources.go`, but is no longer in the chart. In other words, this resource existed at some point in the chart, but was removed. We still need the reconciler to be aware of that, so that when it deletes/upgrades a chart release, it will be able to remove those.

In this case, `seen[key] = (object is present in the chart) || false`.

Consider the following:
1. We add a resource of `Kind: Bla` in the chart. 
2. A few months pass. 
3. We remove the `Kind: Bla` resource in the chart. 
4. The operator still needs to know that there was, at some point a `Kind: Bla` resource in the charts so it can cleanup properly.

vladbologa: I understood what the code is doing here, but was misled by `seen[k]` which has 3 possible values, although the name implies two:
* not in map -> was never seen
* `true` -> seen in manifests 
* `false` -> was seen in manifests at some point in the past (this is the part that I didn't get)

How about having a type `Seen` instead of `bool`, with three possible values (`Never` = default, `Now`, `Previously`). Would make the code more verbose but also easier to understand imo.

vladbologa: I'm ok also with your recent changes, don't have to do ^

fleetshard/pkg/central/reconciler/gen/main.go
```diff
@@ -0,0 +1,232 @@
...
+func splitFileIntoResources(fileBytes []byte) []string {
+	fileStr := string(fileBytes)
+	return strings.Split(fileStr, "---")
+}
+
+func findGVKsInFile(fileBytes []byte, seen resourceMap) error {
+	resources := splitFileIntoResources(fileBytes)
+	for _, resource := range resources {
+		gvk, ok, err := findResourceGVK(resource)
+		if err != nil || !ok {
```
vladbologa, line 200: Just to double check that this was the intention here: if the function returns ok `false` and error `nil`, it will return `nil`, so a resource without `apiVersion` or `kind` will be silently discarded.

ludydoo: Yeah that was the intention, but in light of your comment, I changed it to return an error if apiVersion/kind is not found

vladbologa (approved):
lgtm, thanks!

PR #1830, opened 2024-05-27
Improve logging for tenant resources helm chart values

ludydoo:


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -1565,6 +1565,10 @@ func (r *CentralReconciler) ensureChartResourcesExist(ctx context.Context, remot
 		return fmt.Errorf("obtaining values for resources chart: %w", err)
 	}
 
+	if features.PrintTenantResourcesChartValues.Enabled() {
+		glog.Infof("Tenant resources for central %q: %s", remoteCentral.Metadata.Name, vals)
```
kovayur, line 1569: I realise we don't have sensitive values right now, but it would be nice to mask them here

kovayur (approved):

PR #1856, opened 2024-06-03
ROX-23709: Rework local kubernetes issuer resolution

kovayur:
Rework kubernetes issuer resolution mainly for local/dev kubernetes clusters.
An initial implementation didn't work as expected, since the kubernetes issuer discovery requries an authorized service account. Authentication handler from ocm-sdk-go doesn't allow to pass a service account token to retrieve a JWKS file. Therefore, I've introduced a special use-case for such clusters and download JWKs to a temp file.   

There are also a few related changes to enable the SERVICE_ACCOUNT_TOKEN auth type in all environments. 

## Test manual

```
# To run tests locally run:
make ocm/setup
```


dev/env/manifests/fleetshard-sync/02-fleetshard-sync-rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: fleetshard-sync
+    namespace: "$ACSCS_NAMESPACE"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: fleetshard-sync-role
+# // TODO(create-ticket): Replace with least privileges permissions
```
dhaus67, line 24: Let's also do this in this PR and replace the number here.

dev/env/manifests/fleetshard-sync/03-fleetshard-sync-deployment.yaml
```diff
@@ -78,16 +79,23 @@ spec:
...
+              value: "/var/run/secrets/tokens/fleet-manager-token"
           image: "${FLEET_MANAGER_IMAGE}"
           imagePullPolicy: IfNotPresent
           name: fleetshard-sync
           resources: $FLEETSHARD_SYNC_RESOURCES
           volumeMounts:
-            - mountPath: /secrets
-              name: secrets
+            - mountPath: /var/run/secrets/tokens
+              name: fleet-manager-token
```
dhaus67, line 92: nit: set it to readonly as well.

dev/env/manifests/fleetshard-sync/03-fleetshard-sync-deployment.yaml
```diff
@@ -78,16 +79,23 @@ spec:
                 secretKeyRef:
                   name: fleetshard-sync
                   key: "tenant-image-pull-secret"
+            - name: AUTH_TYPE
+              value: "SERVICE_ACCOUNT_TOKEN"
```
dhaus67, line 83: If it's the default going forward, why even set it to being with in the dev environment?

fleetshard/config/config.go
```diff
@@ -19,23 +19,24 @@ const (
...
-	MetricsAddress       string        `env:"FLEETSHARD_METRICS_ADDRESS" envDefault:":8080"`
-	EgressProxyImage     string        `env:"EGRESS_PROXY_IMAGE"`
-	DefaultBaseCRDURL    string        `env:"DEFAULT_BASE_CRD_URL" envDefault:"https://raw.githubusercontent.com/stackrox/stackrox/%s/operator/bundle/manifests/"`
+	FleetManagerEndpoint    string        `env:"FLEET_MANAGER_ENDPOINT" envDefault:"http://127.0.0.1:8000"`
+	StartupTimeout          time.Duration `env:"STARTUP_TIMEOUT" envDefault:"300s"`
+	ClusterID               string        `env:"CLUSTER_ID"`
+	ClusterName             string        `env:"CLUSTER_NAME"`
+	Environment             string        `env:"ENVIRONMENT"`
+	RuntimePollPeriod       time.Duration `env:"RUNTIME_POLL_PERIOD" envDefault:"5s"`
+	AuthType                string        `env:"AUTH_TYPE" envDefault:"SERVICE_ACCOUNT_TOKEN"`
```
dhaus67, line 28: Are we ready to make it the default yet, given that you've uncovered some issues during testing? My worry is that with the next ACSCS release this gets to prod and things don't work _yet_?

I'd prefer to first check whether things work as expected and only then do the breaking change by moving the default value.

pkg/client/iam/config.go
```diff
@@ -67,6 +68,19 @@ type IAMRealmConfig struct {
 	APIEndpointURI   string `json:"api_endpoint_uri"`
 }
 
+// KubernetesIssuer specific to service account issuer discovery.
+// Unlike OIDCIssuers used in cases where unauthorised/anonymous access to the issuer endpoint is not permitted.
```
dhaus67, line 72: This reads like unauthorized/anonymous access is not permitted for OIDC issuers, whereas this only applies to the kubernetes issuer. Maybe name it something along the likes of "The KubernetesIssuer doesn't allow unauthorized..." to make things more clear?

pkg/client/iam/config.go
```diff
@@ -67,6 +68,19 @@ type IAMRealmConfig struct {
...
+// The purpose of this issuer is to support local k8s clusters.
+// see: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery.
+type KubernetesIssuer struct {
+	// Enabled add to the server configuration if true
+	Enabled bool
+	// IssuerURI - uri of the issuer endpoint
+	IssuerURI string
+	// JWKSFile location of the file to which the JWKs content is written
+	JWKSFile string
```
dhaus67, line 81: nit:
```suggestion
	// Enabled add to the server configuration if true.
	Enabled bool
	// IssuerURI - URI of the issuer endpoint.
	IssuerURI string
	// JWKSFile location of the file to which the JWKs content is written,
	JWKSFile string
```

pkg/client/iam/config.go
```diff
@@ -67,6 +68,19 @@ type IAMRealmConfig struct {
 	APIEndpointURI   string `json:"api_endpoint_uri"`
 }
 
+// KubernetesIssuer specific to service account issuer discovery.
+// Unlike OIDCIssuers used in cases where unauthorised/anonymous access to the issuer endpoint is not permitted.
+// The purpose of this issuer is to support local k8s clusters.
+// see: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery.
+type KubernetesIssuer struct {
```
dhaus67, line 75: So since this all is based on the need to have things for the development environment working where one would want to use the local issuer, you could also do the following: 
```
k create clusterrolebinding oidc-reviewer --clusterrole=system:service-
account-issuer-discovery --group=system:unauthenticated
```
will allow anonymous access to the OIDC discovery URL within the cluster - since we control the role bindings in the dev cluster, why not just create the YAMLs and re-use existing OIDCIssuer instead of making this a special case? This will make things easier to maintain because, well, it will be less to maintain in terms of additional configurations.

dhaus67 (changes requested):

dev/env/manifests/fleetshard-sync/02-fleetshard-sync-rbac.yaml
kovayur: Done.

dev/env/manifests/fleetshard-sync/03-fleetshard-sync-deployment.yaml
kovayur: Done.

dev/env/manifests/fleetshard-sync/03-fleetshard-sync-deployment.yaml
kovayur: I have returned the default value as suggested below. Thus, the explicit assignment here should remain.

fleetshard/config/config.go
kovayur: Fair enough, reverted.

pkg/client/iam/config.go
kovayur: I reworded the description, I hope it's clearer now.

pkg/client/iam/config.go
kovayur: Done.

pkg/client/iam/config.go
kovayur: Nevertheless, it's a great suggestion.

dhaus67 (approved):

PR #1826, opened 2024-05-24
ROX-23255: add auth implementation for emailsender API

johannes94:
## Checklist (Definition of Done)

- [x] Unit and integration tests added
- [x] Added test description under `Test manual`
- [ ] Documentation added if necessary (i.e. changes to dev setup, test execution, ...) -> will be added once emailsender is ready to be worked on by others
- [x] CI and all relevant tests are passing
- [x] Add the ticket number to the PR title if available, i.e. `ROX-12345: ...`
- [x] Discussed security and business related topics privately. Will move any security and business related topics that arise to private communication channel.
- [ ] Add secret to app-interface Vault or Secrets Manager if necessary
- [ ] RDS changes were e2e tested [manually](../docs/development/howto-e2e-test-rds.md)
- [ ] Check AWS limits are reasonable for changes provisioning new resources
- [ ] (If applicable) Changes to the dp-terraform Helm values have been reflected in the addon on integration environment

## Test manual

Additional to the added unit tests I tested the API as follows:

```
# Start emailservice
export CLUSTER_ID=test
go run emailsender/cmd/app/main.go

# By default the authentication configuration will allow calling the API with a OCM token of any
# engineer in the ACS organization, so make sure you're logged in to ocm.

# Run
curl localhost:8080/api/v1/acscsemail -H "Authorization: Bearer $(ocm token)" -H "Content-Type: application/json" -XPOST --data-raw '{"to":["test@test.acscs-email.test"],"rawMessage":"dGVzdCBtZXNzYWdlIGNvbnRlbnQ="}' -v
# Should yield a 200 status code

# Run the same without auth header
curl localhost:8080/api/v1/acscsemail -H "Content-Type: application/json" -XPOST --data-raw '{"to":["test@test.acscs-email.test"],"rawMessage":"dGVzdCBtZXNzYWdlIGNvbnRlbnQ="}'
# Should yield a 401 status code
```

```
# To run tests locally run:
make db/teardown db/setup db/migrate
make ocm/setup
make verify lint binary test test/integration
```

PR #1860, opened 2024-06-05
ROX-23255: emailsender read auth cfg from kubernetes

johannes94:
## Description

Follow Up on #1826 .

This adds a function to the AuthConfig of emailsender that allows reading it from it's own Kubernetes service account.

emailsender is expected to be called by central with it's Kubernetes service account. Since emailsender has its own service account as well we can use that information to read the configuration values required for authentication (issuer, audience, jwks_url).

This way we don't need a config file / config map that is different per environment we deploy to, where we have to manually managed the input vars (e.g addon parameters).

## Test manual

- Deployed the emailsender service to kube
- Create a dummy pod central in a dummy namespace rhacs-test with nginx image called central
- Exec into that pod and send a curl to emailsender with the serviceaccount -> works
- Send the curl without authentication -> is blocked

```
# To run tests locally run:
make ocm/setup
```

### DIFF
--- a/pkg/shared/tls.go
+++ b/pkg/shared/tls.go
@@ -1,0 +1,29 @@
+package shared
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"os"
+)
+
+// TLSWithAdditionalCAs returns a tls config with addiotional trusted ca certificates.
+// It uses the systems default certificates and appends the CA certificates in the given files.
+func TLSWithAdditionalCAs(caFiles ...string) (*tls.Config, error) {
+	rootCAs, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load system cert pool: %w", err)
+	}
+
+	for _, caFile := range caFiles {
+		ca, err := os.ReadFile(caFile)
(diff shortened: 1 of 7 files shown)


pkg/shared/tls.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"os"
+)
+
+// TLSWithAdditionalCAs returns a tls config with addiotional trusted ca certificates.
+// It uses the systems default certificates and appends the CA certificates in the given files.
+func TLSWithAdditionalCAs(caFiles ...string) (*tls.Config, error) {
```
johannes94, line 12: @kurlov I copied a lot of this logic from the PR you already review in the core repo. The function itself is still somewhat different, as it errors on failed certificates loads and does not return a full http.Transport like [here](https://github.com/stackrox/stackrox/blob/283cc77f1bee95f1f74ae1d063890bb3e9a002de/central/notifiers/acscsemail/client_impl.go#L76).

kurlov (approved):

kurlov:
Looks great 🎉

PR #1868, opened 2024-06-06
ROX-22549: Add an addon metric to track version mismatch

kovayur:
## Description
The metric is needed when the OCM api returns success after an update, but something is wrong on the cluster and the components cannot be updated. This leads to a situation where all components of the data plane are running but are not up to date.

## Test manual

```
# To run tests locally run:
make ocm/setup
```

### DIFF
--- a/internal/dinosaur/pkg/services/addon.go
+++ b/internal/dinosaur/pkg/services/addon.go
@@ -116,11 +116,11 @@ func (p *AddonProvisioner) Provision(cluster api.Cluster, expectedConfigs []gito
 			continue
 		}
 		if clusterInstallationDifferent(installedOnCluster, versionInstalledInOCM) {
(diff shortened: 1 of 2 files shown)


internal/dinosaur/pkg/services/addon.go
```diff
@@ -114,9 +116,11 @@ func (p *AddonProvisioner) Provision(cluster api.Cluster, expectedConfigs []gito
 			continue
 		}
 		if clusterInstallationDifferent(installedOnCluster, versionInstalledInOCM) {
+			updateClusterAddonMismatchDurationMetric(clusterID, installedOnCluster, versionInstalledInOCM, time.Since(installedInOCM.UpdatedTimestamp()))
```
ebensh, line 119: Instead of doing the duration calculation here, would it make more sense to export the timestamp metric and an in-sync metric to do the difference in Prometheus? This usually makes Prometheus/Grafana behave better when the metrics are monotonically increasing and the delta is calculated there.

kovayur: changed to status enum

ebensh (approved):

PR #1917, opened 2024-06-26
ROX-24040: Set Podman instead of Docker to fix RHEL8 builders

ebensh:
buildx build --push is not yet supported by Podman: https://github.com/containers/buildah/pull/4677
So we switch to single-arch build, then build and push separately to match the other deployments.

Tested:
```
QUAY_USER=ebenshet+ebenshet_quay_robot QUAY_TOKEN=abc123 QUAY_IMAGE_REPOSITORY=ebenshet/myrepo PROBE_QUAY_IMAGE_REPOSITORY=ebenshet/myrepo EMAILSENDER_QUAY_IMAGE_REPOSITORY=ebenshet/myrepo ./build_push_app_interface.sh
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -497,12 +497,12 @@ docker/login: docker/login/fleet-manager
 .PHONY: docker/login
 
 docker/login/fleet-manager:
-	@docker logout quay.io
+	$(DOCKER) logout quay.io || true  # Swallog podman error if not logged in
 	@DOCKER_CONFIG=${DOCKER_CONFIG} $(DOCKER) login -u "${QUAY_USER}" --password-stdin <<< "${QUAY_TOKEN}" quay.io
 .PHONY: docker/login/fleet-manager
 
 docker/login/probe:
-	@docker logout quay.io
+	$(DOCKER) logout quay.io || true  # Swallow podman error if not logged in
 	@DOCKER_CONFIG=${DOCKER_CONFIG} $(DOCKER) login -u "${QUAY_PROBE_USER}" --password-stdin <<< "${QUAY_PROBE_TOKEN}" quay.io
 .PHONY: docker/login/probe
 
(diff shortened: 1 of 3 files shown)


build_push_app_interface.sh
```diff
@@ -36,6 +36,9 @@ IMAGE_REPOSITORY="${QUAY_IMAGE_REPOSITORY:-app-sre/acs-fleet-manager}"
 PROBE_IMAGE_REPOSITORY="${PROBE_QUAY_IMAGE_REPOSITORY:-app-sre/acscs-probe}"
 EMAILSENDER_IMAGE_REPOSITORY="${PROBE_QUAY_IMAGE_REPOSITORY:-app-sre/acscs-emailsender}"
 
+# Support RHEL8 build images for App-Interface ci-ext
+DOCKER=podman
```
kovayur, line 40:
```suggestion
export DOCKER=podman
```
😬

build_push_app_interface.sh
```diff
@@ -36,6 +36,9 @@ IMAGE_REPOSITORY="${QUAY_IMAGE_REPOSITORY:-app-sre/acs-fleet-manager}"
...
+
 source ./scripts/build_setup.sh
 
 # Push the image:
```
kovayur, line 44: Comment for the lines 47, 57:
Do we need remove this for podman or use an alternative?

kovayur:
I'm not sure if `buildx` command is supported for podman

build_push_app_interface.sh
ebensh: Done

build_push_app_interface.sh
ebensh: Not sure which lines /flags you mean, can you please specify?

kovayur: `DOCKER_CONFIG=${DOCKER_CONFIG}` 
We used it only on app-interface and I'm happy to get rid of it.
I don't think it makes any difference to podman.

Makefile
```diff
@@ -497,12 +497,12 @@ docker/login: docker/login/fleet-manager
 .PHONY: docker/login
 
 docker/login/fleet-manager:
-	@docker logout quay.io
+	$(DOCKER) logout quay.io || true  # Swallog podman error if not logged in
```
kovayur, line 500: nit: How about renaming `DOCKER` to `CONTAINER_ENGINE` ?

ebensh:
@kovayur PTAL

ludydoo (approved):

Makefile
```diff
@@ -66,10 +66,6 @@ else
 GOBIN=$(shell $(GO) env GOBIN)
 endif
 
-ifeq ($(IMAGE_PLATFORM),)
```
kovayur, line 69: This will break multi-arch builds.
1. We build arm images on Github actions
2. We build x86 images for openshift internal registry (image/push/internal) on m1

PR #1924, opened 2024-07-01
fix: add a cleanup job for email events

johannes94:
## Description

We don't cleanup emails that we don't longer need to enforce our 24h rolling window for rate limits.

This can lead to a ever growing DB and performance penalties for a long running emailsender DB.

This PR solves this problem by adding a background job that periodically deletes all email sent events older than 2 days.

## Test manual

```
# To run tests locally run:
make db/teardown db/setup
CLUSTER_ID=CLUSTER_ID go run emailsender/cmd/app/main.go

# Create a few test events by directly connecting to DB
make db/psql
# Run this command a few times to create "unexpired" events
INSERT INTO email_sent_by_tenants (tenant_id, created_at) VALUES ('test', NOW()');
# Run this command a few times to create "expired" events
INSERT INTO email_sent_by_tenants (tenant_id, created_at) VALUES ('test', NOW()- INTERVAL '3 DAY');

# Make sure after configured interval for the worker (2 minutes) the expired events are
# deleted, while "new" events are still in the DB
```

### DIFF
--- a/emailsender/cmd/app/main.go
+++ b/emailsender/cmd/app/main.go
@@ -5,10 +5,13 @@ import (
 	"context"
 	"errors"
 	"flag"
-	"github.com/stackrox/acs-fleet-manager/emailsender/pkg/db"
 	"net/http"
 	"os"
 	"os/signal"
+	"time"
+
+	"github.com/stackrox/acs-fleet-manager/emailsender/pkg/db"
+	"github.com/stackrox/acs-fleet-manager/emailsender/pkg/workers"
 
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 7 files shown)


emailsender/cmd/app/main.go
```diff
@@ -58,6 +62,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	cleanupWorker := workers.CleanupEmailSent{
+		DbConn:       dbConnection,
+		Period:       time.Second * 60,
```
kurlov, line 67: Do you think it might be beneficial to extend default period? Perhaps run it every 30 minutes or so. The motivation is that 30 minutes stale records might not affect rate limiter query too much so it might be fine to reduce invokation frequency

kurlov: Also, might be worth to add it to the config

johannes94: Made it configurable and set default period to 5min. I think the overhead of running it every 5 minutes should be ok. Although it could be higher having it at 5mins might help to spot errors easier because the log is more frequently.

kurlov (approved):

PR #1940, opened 2024-07-09
ROX-23263: move from OSCI cluster pool to infra clusters for e2e

johannes94:
## Description

This PR prepares the required changes in this repo for e2e tests to run against cluster create through stackrox infra and respective workflows in the [openshift/release](https://github.com/openshift/release) repository.

This PR adds:
- Wrapper scripts to be able to properly configure, create and destroy stackrox infra cluster via the workflow
- Changes to FM's in cluster oidc implementation, so that it is compatible with the OIDC discovery document returned by infra clusters

To test the changes, I created a PR with a rehearsal job in the openshift/release repository:
https://github.com/openshift/release/pull/54055

As a follow up once this is merged, instead of having a job that runs against the branch in this PR, I will modify our current e2e ci-operator configuration to run a job like above PR but against the main branch.

## Test manual

Successful rehearsal job in openshift/release PR linked above.

```
# To run tests locally run:
make ocm/setup
```


scripts/ci/lib.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# A library of CI related reusable bash functions
+
+ci_export() {
+    if [[ "$#" -ne 2 ]]; then
+        die "missing args. usage: ci_export <env-name> <env-value>"
```
kovayur, line 7: Need to source `log.sh` to use `die` function.

scripts/ci/lib.sh
```diff
@@ -0,0 +1,36 @@
...
+
+    ci_export "$@"
+
+    local env_name="$1"
+    local env_value="$2"
+
+    echo "export ${env_name}=${env_value}" | tee -a "${SHARED_DIR:-/tmp}/shared_env"
+}
+
+info() {
```
kovayur, line 34: Duplicates `log` from `log.sh` to some extent

pkg/client/iam/config.go
```diff
@@ -319,14 +319,31 @@ func (i *KubernetesIssuer) getJwksURI(client *http.Client) (string, error) {
...
 		// kube api-server returns an internal IP, need to override it.
 		jwksURI = i.overrideJwksURIForLocalCluster(jwksURI)
 	}
+
+	jwksURL, err := url.Parse(jwksURI)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to parse jwksURI as net/url: %s", jwksURI)
+	}
+
+	glog.V(5).Infof("jwks_uri before replacement logic: %s", jwksURI)
```
kovayur, line 333: Maybe move that line to the if clause? It is not guaranteed that url can be overridden. In this case it could be confusing.

pkg/client/iam/config.go
```diff
@@ -319,14 +319,31 @@ func (i *KubernetesIssuer) getJwksURI(client *http.Client) (string, error) {
...
+	if netutil.IsIPAddress(jwksURL.Hostname()) && i.IssuerURI == kubernetesIssuer {
+		// in some cases like infra OCP the cluster internal jwks_uri in the discovery document
+		// is a private IP address of the pod running the oidc server. This breaks tls validation.
+		// This override makes sure that in those cases kubernetes.default.svc is used instead of the IP
+		glog.V(5).Infof("Configured issuer is: %s and jwks_uri contains IP, replacing host with internal kubernetes svc", i.IssuerURI)
+		jwksURI = i.overrideJwksURIForInternalCluster(jwksURL)
```
kovayur, line 339: Just thinking out loud: Maybe we need to fallback to this scenario in all cases where we fail to fetch the jwks file? 
1. IP check looks too heuristic, in theory it can be public;
2. It might be necessary in other edge cases;
3. We could expand this fallback to the local cluster edge-case that also looks like a workaround.

It doesn't have to be done right now, just food for thought for future refactoring.

scripts/ci/lib.sh
```diff
@@ -0,0 +1,36 @@
...
+ci_export() {
+    if [[ "$#" -ne 2 ]]; then
+        die "missing args. usage: ci_export <env-name> <env-value>"
+    fi
+
+    local env_name="$1"
+    local env_value="$2"
+
+    if command -v cci-export >/dev/null; then
+        cci-export "$env_name" "$env_value"
```
kovayur, line 14: Are we going to use `cci-export` in CI? Couldn't find any relevant information about it. If not, maybe we could simplify this function..

kovayur (approved):

scripts/ci/lib.sh
johannes94: Changed scripts to use logs.

pkg/client/iam/config.go
johannes94: This was just a log I used to see while debugging what goes into the if statement. removed it.

kovayur (approved):

pkg/client/iam/config.go
johannes94: Sound like it would be good to do some refactoring here. Lets discuss what you have in mind in more detail in slack and do the refactoring in a follow up PR.

PR #1942, opened 2024-07-11
ROX-25276: emailsender auth on rosa

johannes94:
## Description
Use a configurable `jwks_uri` independent from the openid discovery document for emailsenders internal kubernetes authentication.

This proofs to be more reliable since every K8s distribution we use has `https://kubernetes.default.svc/openid/v1/jwks` for internal access to the jwks keys, while it seems that each distribution (crc, minikube, OSD, ROSA, colima, infra OCP) return different URLs. Those URLs all require different handling:
- K8s Pod IP in URL
- K8s Node IP in URL
- K8s svc DNS in URL
- Public facing URL
- Some require authentication, some don't

If we face a K8s distribution that doesn't follow that pattern we can still modify the URL via the new env configuration values.

## Test manual

CI is sufficient.

### DIFF
--- a/emailsender/config/config.go
+++ b/emailsender/config/config.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 	"os"
 	"path"
-	"strings"
 	"time"
 
 	"github.com/caarlos0/env/v6"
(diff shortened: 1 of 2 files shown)


johannes94:
/retest
Cluster creation failed

emailsender/config/config.go
```diff
@@ -230,11 +232,9 @@ func (c *AuthConfig) getOIDCConfig(token string) (oidcConfig, error) {
 	return oidcCfg, nil
 }
 
-func (c *AuthConfig) getJWKS(oidcCfg oidcConfig, token string) ([]byte, error) {
-	// replacing the potentially public facing JWKS url with the cluster internal k8sSvcURL
-	// since we don't want to call the endpoint via ingress but within the cluster
-	jwksPath := jwksPathFromURL(oidcCfg.JwksURI)
-	jwksRequest, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/%s", c.k8sSvcURL, jwksPath), nil)
+func (c *AuthConfig) getJWKS(token string) ([]byte, error) {
+	jwksURL := fmt.Sprintf("%s/%s", c.k8sSvcURL, c.k8sJWKSPath)
```
ebensh, line 236: Nit - won't this have an extra /?
`https://kubernetes.default.svc//openid/v1/jwks`

emailsender/config/config_test.go
```diff
@@ -71,30 +71,6 @@ func TestGetConfigFailureEnabledHTTPSOnly(t *testing.T) {
 	assert.Nil(t, cfg)
 }
 
-func TestJWKSPathFromURL(t *testing.T) {
```
ebensh, line 74: This test would've caught the extra / problem

ebensh (changes requested):

emailsender/config/config_test.go
johannes94: This was testing a deleted private method, but I've added a test to verify the new joining logic.

emailsender/config/config.go
johannes94: Right good catch, one could say this was misconfiguration but I've added a method to make sure we don't produce paths like this.

ebensh (approved):

PR #2009, opened 2024-09-02
ROX-26039: Create tenant argoCd app

ludydoo:
## Description

Adds the support for deploying tenant-resources through ArgoCD rather than the built-in `tenant-resources` chart. The `CentralReconciler` will

- Create an ArgoCD application if the flag is enabled for a tenant
- Delete the ArgoCD application if the flag is disabled (or do nothing)

Screenshot 2024-11-12 at 10 59 54 AM

Environment variables have been added that represent the default `tenant-resources` application source. Currently, it defaults to 

| | |
|----|----|
| **Repo** | `github.com/stackrox/acscs-manifests` |
| **Path** |  `tenant-resources` |
| **Ref** | `HEAD` |

Because this is an opt-in feature, at this point deploying this will have no consequence on the int/stage/prod environments.

The `acscs-manifests` repository is private. But ArgoCD should have access to it thanks to https://github.com/stackrox/acs-fleet-manager-aws-config/pull/257

A further step in the development of the feature is to enable the local development support for ArgoCD: https://github.com/stackrox/acs-fleet-manager/pull/2074/files

## Test manual

```
# To run tests locally run:
make ocm/setup
```


kovayur (approved):
🔥

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -236,8 +245,26 @@ func (r *CentralReconciler) Reconcile(ctx context.Context, remoteCentral private
...
-	if err := r.ensureChartResourcesExist(ctx, remoteCentral); err != nil {
-		return nil, errors.Wrapf(err, "unable to install chart resource for central %s/%s", central.GetNamespace(), central.GetName())
+	if remoteCentral.Spec.ArgoCd.Enabled {
+		if err := r.ensureArgoCdApplicationExists(ctx, remoteCentral); err != nil {
+			return nil, errors.Wrapf(err, "unable to install ArgoCD application for central %s/%s", remoteCentral.Metadata.Namespace, remoteCentral.Metadata.Name)
+		}
+	} else {
+
+		{
+			// This little part would only happen if we enable, then disable ArgoCD for a tenant
```
johannes94, line 255: In my understanding this part would execute on every reconciliation where `remoteCentral.Spec.ArgoCd.Enabled != true`, am I wrong here? If so could you explain why?

Maybe the function would be NOOP in that case but it would go through this code path.

johannes94 (changes requested):
I don't like being the "Process-driven block a PR because of formal issues guy", but for a change like this, I'd expect more information in the PR description. For instance how you've tested the changes and also a functional description of what you expect / not expect to work after this PR.

fleetshard/config/config.go
```diff
@@ -35,13 +35,16 @@ type Config struct {
...
-	ManagedDB             ManagedDB
-	Telemetry             Telemetry
-	AuditLogging          AuditLogging
-	SecretEncryption      SecretEncryption
-	RouteParameters       RouteConfig
-	FleetshardAddonName   string `env:"FLEETSHARD_ADDON_NAME" envDefault:"acs-fleetshard"`
+	TenantImagePullSecret               string `env:"TENANT_IMAGE_PULL_SECRET"`
+	DefaultTenantArgoCdAppSourceRepoURL string `env:"TENANT_ARGOCD_APP_SOURCE_REPO_URL_DEFAULT" envDefault:"https://github.com/stackrox/acscs-manifests.git"`
+	DefaultTenantArgoCdAppSourceRef     string `env:"TENANT_ARGOCD_APP_SOURCE_REF_DEFAULT" envDefault:"HEAD"`
+	DefaultTenantArgoCdAppSourcePath    string `env:"TENANT_ARGOCD_APP_SOURCE_PATH_DEFAULT" envDefault:"tenant-resources"`
```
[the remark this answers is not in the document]
ludydoo: I named it default in the sense of it being the default if there are no per-tenant overrides. You are recommending dropping the DEFAULT here ?

fleetshard/pkg/central/reconciler/reconciler.go
ludydoo: Not every reconciliation, only if the hashSum has been changed. 
This would run if 
1. ArgoCD enabled for a tenant
2. Then ArgoCD disabled for a tenant

This would basically clean up the ArgoCD application

fleetshard/config/config.go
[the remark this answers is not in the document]
johannes94: For this specific env variables I'd recommend to make the naming consistent. The default should be either a prefix or a suffix, for instance: Env Variable: `TENANT_ARGOCD_APP_SOURCE_REF_DEFAULT` should be field  `TenantArgoCdAppSourceRefDefault` not `DefaultTenantArgoCdAppSourceRef`

For the `CentralReconciler` / `CentralReconcilerOptions` I don't see any logic for overriding that values on a per-tenant basis, so I'm questioning if this is going to be added in follow up PRs or if the "Default" prefix is unnecessary.

fleetshard/pkg/central/reconciler/reconciler.go
johannes94: > Not every reconciliation, only if the hashSum has been changed.
True, forgot about that part.

Still I think this code path would run even if ArgoCD was never enabled for a tenant. Not only in the situation described above, which makes the comment misleading.

fleetshard/config/config.go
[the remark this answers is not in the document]
ludydoo: The ability to override the ArgoCD application source (repo/path/ref) will be added later

fleetshard/pkg/central/reconciler/reconciler.go
ludydoo: Ah I see what you mean. Will change to
```
This little part handles the case where we enable, then disable ArgoCD for a tenant
```

fleetshard/config/config.go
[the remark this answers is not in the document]
ludydoo: I'll rename the env var to be consistent

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -1106,6 +1160,12 @@ func (r *CentralReconciler) ensureCentralDeleted(ctx context.Context, remoteCent
 		globalDeleted = globalDeleted && secretDeleted
 	}
 
+	argoCdAppDeleted, err := r.ensureArgoCdApplicationDeleted(ctx, *remoteCentral)
```
johannes94, line 1163: Nit: you might want to guard this line with the argo enabled flag as well. But IIUC it would work as long as the respective CRD is available.

johannes94 (approved):

dev/env/scripts/bootstrap.sh
```diff
@@ -58,12 +58,16 @@ if [[ "$INSTALL_ARGOCD" == "true" ]]; then
     kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml
 elif [[ "$INSTALL_OPENSHIFT_GITOPS" == "true" ]]; then
     log "Installing Openshift GitOps"
+    kubectl create ns openshift-gitops || true
```
kovayur, line 61: Shouldn't the namespace manifest be defined in `${MANIFESTS_DIR}/openshift-gitops` instead ?

dev/env/scripts/bootstrap.sh
```diff
@@ -58,12 +58,16 @@ if [[ "$INSTALL_ARGOCD" == "true" ]]; then
...
     apply "${MANIFESTS_DIR}/openshift-gitops"
 else
     log "One of ArgoCD or OpenShift GitOps must be installed"
     exit 1
 fi
 
+log "Installing acscs-manifests repository secret"
+"${GITROOT}/scripts/import_argocd_repo_secret.sh"
```
kovayur, line 69: Instead of generating a secret you can put it to the `${MANIFESTS_DIR}/openshift-gitops` folder and change the line 62: 
```
chamber exec gitops -b secretsmanager -- apply "${MANIFESTS_DIR}/openshift-gitops"
```

`apply` supports `envsubst`. That's how the secrets propagation works in other components.

kovayur (approved):

PR #2079, opened 2024-10-21
ROX-26277: add cleanup for stale tenants to FS

johannes94:

## Test manual

```
# To run tests locally run:
make ocm/setup
```

### DIFF
--- a/fleetshard/pkg/central/reconciler/central_cr_reconciler.go
+++ b/fleetshard/pkg/central/reconciler/central_cr_reconciler.go
@@ -1,0 +1,157 @@
+package reconciler
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"github.com/stackrox/acs-fleet-manager/fleetshard/pkg/util"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/private"
+	"github.com/stackrox/rox/operator/apis/platform/v1alpha1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CentralCrReconciler provides methods to reconcile the Central CR associated with a tenant
+type CentralCrReconciler struct {
+	client ctrlClient.Client
+}
+
+// NewCentralCrReconciler creates a CentralCrReconciler using given arguments
+func NewCentralCrReconciler(k8sClient ctrlClient.Client) *CentralCrReconciler {
+	return &CentralCrReconciler{client: k8sClient}
+}
+
+// Reconcile create or updates the central CR for the tenant in the cluster so that
+// it fits to the remote central object received.
+func (r *CentralCrReconciler) Reconcile(ctx context.Context, remoteCentral *private.ManagedCentral, central *v1alpha1.Central) error {
+	remoteCentralName := remoteCentral.Metadata.Name
+	remoteCentralNamespace := remoteCentral.Metadata.Namespace
+
+	centralExists := true
+	existingCentral := v1alpha1.Central{}
+	centralKey := ctrlClient.ObjectKey{Namespace: remoteCentralNamespace, Name: remoteCentralName}
+	err := r.client.Get(ctx, centralKey, &existingCentral)
+	if err != nil {
+		if !apiErrors.IsNotFound(err) {
+			return errors.Wrapf(err, "unable to check the existence of central %v", centralKey)
+		}
+		centralExists = false
+	}
+
+	if remoteCentral.Metadata.ExpiredAt != nil {
+		if central.GetAnnotations() == nil {
+			central.Annotations = map[string]string{}
+		}
+		central.Annotations[centralExpiredAtKey] = remoteCentral.Metadata.ExpiredAt.Format(time.RFC3339)
+	}
+
+	if !centralExists {
+		if central.GetAnnotations() == nil {
+			central.Annotations = map[string]string{}
+		}
+		if err := util.IncrementCentralRevision(central); err != nil {
+			return errors.Wrapf(err, "incrementing Central %v revision", centralKey)
+		}
+
+		glog.Infof("Creating Central %v", centralKey)
+		if err := r.client.Create(ctx, central); err != nil {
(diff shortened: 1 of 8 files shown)


johannes94:
/retest
Cluster failure

fleetshard/pkg/central/reconciler/central_cr_reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// NewCentralCrReconciler creates a CentralCrReconciler using given arguments
+func NewCentralCrReconciler(k8sClient ctrlClient.Client) *CentralCrReconciler {
+	return &CentralCrReconciler{client: k8sClient}
+}
+
+// Reconcile create or updates the central CR for the tenant in the cluster so that
+// it fits to the remote central object received.
+func (r *CentralCrReconciler) Reconcile(ctx context.Context, remoteCentral *private.ManagedCentral, central *v1alpha1.Central) error {
```
ludydoo, line 30: I think we should only pass the `*private.ManagedCentral` in `Reconcile`, and have this reconciler build the `*v1alpha1.Central` internally

fleetshard/pkg/central/reconciler/central_cr_reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+			return errors.Wrapf(err, "updating Central %v", centralKey)
+		}
+
+	}
+
+	return nil
+}
+
+// EnsureDeleted deletes the central CR from the cluster
+func (r *CentralCrReconciler) EnsureDeleted(ctx context.Context, namespace, crName string) (bool, error) {
```
ludydoo, line 99: Hm, somehow it doesn't feel right to have this exported method in a Reconciler. Usually reconcilers just have the `Reconcile` method. Could we have some exported method without receiver, like `func EnsureCentralDeleted(...)` ? What do you think?

fleetshard/pkg/runtime/runtime.go
```diff
@@ -226,6 +228,13 @@ func (r *Runtime) Start() error {
 		}
 
 		r.deleteStaleReconcilers(&list)
+
+		if features.ClusterMigration.Enabled() {
+			if err := tenantCleanup.DeleteStaleTenantK8sResources(ctx, &list); err != nil {
```
ludydoo, line 233: Are both `tenantCleanup` and `CentralReconciler` competing to delete the  tenant resources that are either
1. .Metada.DeletionTimestamp != nil
2. Not part of FM api response
?

fleetshard/pkg/central/reconciler/tenant_cleanup.go
```diff
@@ -0,0 +1,104 @@
...
+// DeleteK8sResources deletes all associated resources for a managed central from the cluster.
+// Returns potential errors and a bool indicating whether deletion went through successfully
+func (t *TenantCleanup) DeleteK8sResources(ctx context.Context, namespace string, tenantName string) (bool, error) {
+	// Deleting the NS is not enough to cleanup a tenant as there could be non-namespaced resources
+	// within the tenant resource chart or created by the CR. Because of that we delete chart and CR first
+	// to allow propper cleanup of such resources throug helm/ACS operator before removing the namespace.
+	// If any resources wouldn't be deleted by namespace deletion add them here.
+	globalDeleted := true
+
+	chartReconciler := NewTenantChartReconciler(t.k8sClient, t.secureTenantNetwork)
```
ludydoo, line 77: Should this rather be injected ?  (both `chartReconciler` and `crReconciler`) instead of calling the constructors here?

fleetshard/pkg/central/reconciler/central_cr_reconciler.go
johannes94: Building the `*v1alpha1.Central` internally is an interesting idea. Unfortunately as I was trying to do this I noticed that it makes our control flow more complex.

The "parent" reconciler has many functions that depend on the `*v1alpha.Central` object we would need it internally and externally. If we remove the argument from the CentralCrReconciler, we'd compute it on multiple places, which means functions used to build that object would have to be used by CentralCrReconciler and the parent CentralReconciler in order to keep other code of the parent reconciler functioning. A change like this would make it complex to follow where what data is computed and why as there would be multiple sources of a v1alpha.Central objects. IMHO this make it harder to understand.

fleetshard/pkg/central/reconciler/central_cr_reconciler.go
johannes94: I didn't have the same mental concept of a reconciler in mind. Maybe Reconciler in that case is not the right wording for all those "SubResource" components we have as a "Reconcile" usually handles all states including Deletion whereas current sub components basically are a "Upsert" function + a delete function separated.

I see multiple options:
- Live with how it is and the mental mismatch in "Reconciler" / or find a better word and leave as is
- Refactor the Refactoring to actually have every reconciler implement a "Reconcile" method that does everything
  - I would not like this because it makes using the code from a "not primary reconciler" perspective complex (e.g in cleanup) because you'd have to construct a full resources object to get to the code path of "Reconcile" you intend
- Do what you said and make it not part of the struct, which has other downsides. Like more tedious parameter management, because you have to reconstruct what was previously stored in the struct.

fleetshard/pkg/runtime/runtime.go
johannes94: I don't think so.

CentralReconciler would never be called in the same runtime interval as tenantCleanup. Because reconcile is only called on tenants returned by FM API and cleanup is only called on tenants not returned by FM API.

The only potential conflict that I could imagine is if there was a reconciliation that didn't finish yet (within 30sec) and was started in the previous runtime interval. But even in that case both processes should come to a consistent state eventually. Unless the reconciler would implement an internal loop that recreates the namespace indefinitely.

fleetshard/pkg/central/reconciler/tenant_cleanup.go
johannes94: Changed it to be injected.

johannes94: Changed it to be created in the NewCentralReconciler, and NewTenantCleanup functions instead of injected, because I decided that all other "sub reconcilers" should be private to the reconciler pkg for a clearer interface definition from `runtime` to `reconciler` pkg.

johannes94:
/retest
2 different failures indicating a flaky test.

ludydoo (approved):

PR #2127, opened 2024-12-16
ROX-27073: add cluster migration e2e test

johannes94:
## Description
- Adds a ginkgo e2e test suite for multicluster test
- First test verifies cluster migration between multiple clusters
- Adds a make target to run the tests test/multicluster/e2e
- Moved common logic with other e2e tests to a `testutils` package
- Change admin CentralRequest presenter to actually set the ClusterId, which is required for the test
- See test manual on how to execute this agains 2 infra clusters
- Extends our k8s client pkg to allow loading config for multiple cluster

## Test manual

```
# Start 2 infractl clusters OSD on AWS
# Login to kerberos for secret access
kinit jmalsam

export CLUSTER_TYPE="infra-openshift"
export CLUSTER_1_KUBECONFIG=/Users/johannes/kubes/cluster1
export CLUSTER_2_KUBECONFIG=/Users/johannes/kubes/cluster2
export QUAY_USER=<your user>
export QUAY_TOKEN=<your token>
export ENABLE_CENTRAL_EXTERNAL_CERTIFICATE="true"
export ROUTE53_ACCESS_KEY=<route53-key-id>
export ROUTE53_SECRET_ACCESS_KEY=<route53-secret>
export STATIC_TOKEN=<static-token-from-ci>
export STATIC_TOKEN_ADMIN=<static-admin-token-from-ci>

export INFRA_TOKEN=<your-infra-token>

url=$(infractl artifacts jm-migration-1 --json | jq '.Artifacts[] | select(.Name=="kubeconfig") | .URL' -r)
wget -O $CLUSTER_1_KUBECONFIG $url

url=$(infractl artifacts jm-migration-2 --json | jq '.Artifacts[] | select(.Name=="kubeconfig") | .URL' -r)
wget -O $CLUSTER_2_KUBECONFIG $url

bash scripts/ci/multicluster_tests/deploy.sh

# Verify all services are running like expected
KUBECONFIG=$CLUSTER_1_KUBECONFIG k get pods -n rhacs # fleet-manager and fleetshard-sync must be healthy
KUBECONFIG=$CLUSTER_2_KUBECONFIG k get pods -n rhacs # fleetshard-sync must be healthy, fleet-manager does not exist

make test/multicluster/e2e
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,10 @@
 emailsender-manifests.yaml
 central-chart/
 pids-port-forward
+# temp files created by multicluster e2e tests
+cluster-list.json
+cluster-list2.json
+fm-dataplane-config.yaml
 # Ignore uploaded files in development
 /storage/*
 
--- a/e2e/dns/record_cleanup.go
+++ b/e2e/dns/record_cleanup.go
@@ -1,0 +1,35 @@
+package dns
+
+import (
+	"github.com/aws/aws-sdk-go/service/route53"
+	. "github.com/onsi/gomega"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/public"
+	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/services"
+)
+
+// CleanupCentralRequestRecords deletes all route53 resoruces associated with the centralRequest
+func CleanupCentralRequestRecords(route53Client *route53.Route53, centralRequest public.CentralRequest) {
+	dnsLoader := NewRecordsLoader(route53Client, centralRequest)
+	recordSets := dnsLoader.LoadDNSRecords()
+
+	action := string(services.CentralRoutesActionDelete)
+	changes := []*route53.Change{}
+	for _, rs := range recordSets {
+		c := &route53.Change{
+			Action:            &action,
+			ResourceRecordSet: rs,
+		}
+		changes = append(changes, c)
+	}
+
+	if len(changes) == 0 {
+		return
+	}
+
+	_, err := route53Client.ChangeResourceRecordSets(&route53.ChangeResourceRecordSetsInput{
+		HostedZoneId: dnsLoader.rhacsZone.Name,
+		ChangeBatch:  &route53.ChangeBatch{Changes: changes},
+	})
+
+	Expect(err).ToNot(HaveOccurred())
+}
--- a/e2e/e2e_suite_test.go
+++ b/e2e/e2e_suite_test.go
@@ -14,9 +14,9 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	openshiftRouteV1 "github.com/openshift/api/route/v1"
+	"github.com/stackrox/acs-fleet-manager/e2e/testutil"
 	"github.com/stackrox/acs-fleet-manager/fleetshard/config"
 	"github.com/stackrox/acs-fleet-manager/fleetshard/pkg/k8s"
-	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/constants"
 	"github.com/stackrox/acs-fleet-manager/internal/dinosaur/pkg/api/public"
 	"github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager"
 	"github.com/stackrox/rox/operator/api/v1alpha1"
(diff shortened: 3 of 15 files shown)


.gitignore
```diff
@@ -15,6 +15,10 @@
 emailsender-manifests.yaml
 central-chart/
 pids-port-forward
+# temp files created by multicluster e2e tests
+cluster-list.json
```
ebensh, line 19: Where do these come from / why can we not clean them up after the test?

e2e/dns/record_cleanup.go
```diff
@@ -0,0 +1,35 @@
+package dns
```
ebensh, line 1: Would waiting for "deleting" state instead of "deprovisioning" rely on the FM's cleanup of the routes instead of manually cleaning them up here? (sorry if missing something, not used to looking at the e2e tests as much)

e2e/e2e_suite_test.go
```diff
@@ -32,8 +31,8 @@ var (
 	dnsEnabled            bool
 	routesEnabled         bool
 	route53Client         *route53.Route53
-	waitTimeout           = getWaitTimeout()
-	extendedWaitTimeout   = getWaitTimeout() * 3
+	waitTimeout           = testutil.GetWaitTimeout()
```
ebensh, line 34: Just a meta-comment - adding the `testutil.` in a separate PR if possible would've made reviewing just the cluster migration part easier. It's fast enough to skim by but next time would recommend splitting into 2 PRs

e2e/e2e_test.go
```diff
@@ -224,13 +224,7 @@ var _ = Describe("Central", Ordered, func() {
...
 
 			recordSets := dnsRecordsLoader.LastResult
-			for idx, domain := range dnsRecordsLoader.CentralDomainNames {
-				recordSet := recordSets[idx]
-				Expect(len(recordSet.ResourceRecords)).To(Equal(1))
-				record := recordSet.ResourceRecords[0]
-				Expect(*recordSet.Name).To(Equal(domain))
-				Expect(*record.Value).To(Equal(reencryptIngress.RouterCanonicalHostname)) // TODO use route specific ingress instead of comparing with reencryptIngress for all cases
-			}
+			testutil.AssertDNSMatchesRouter(dnsRecordsLoader.CentralDomainNames, recordSets, &reencryptIngress)
```
ebensh, line 227: :+1:

e2e/multicluster/multicluster_migration_test.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager"
+	fmImpl "github.com/stackrox/acs-fleet-manager/pkg/client/fleetmanager/impl"
+	corev1 "k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrlClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	dpCloudProvider = "standalone"
+	dpRegion        = "standalone"
```
ebensh, line 27: nit - a unique value from dpCloudProvider

ebensh (changes requested):
Several nits, nothing major - the flow of the e2e test makes sense and is easy enough to read with the comments you provided throughout

.gitignore
johannes94: Those files would be created and not cleaned up. So you can use them for debugging already, this is just the .gitignore to prevent those files from being commited and pushed to the repo.

e2e/dns/record_cleanup.go
johannes94: I'm not 100% sure about this. But I think FM would only start deleting the route records on `deleting` state.

We trigger a deletion and wait for deletion at the end of this test, which usually cleans up the records. 

This logic is intended as an additional measure on test failures. Because on failures we might be stuck somewhere in the usual reconciliation loop which can lead to a situation where the deletion lifecycle logic doesn't get executed. In a case like this we don't want to leak those resources.

e2e/e2e_suite_test.go
johannes94: Yes, I thought about splitting PRs as well along the way, decided to not do it since I only wanted very few changes. Then it grew over time to what you see now :) . I'll split it next time.

e2e/testutil/testutil.go
```diff
@@ -0,0 +1,63 @@
...
+
+// SkipIf skips a Gingko test container if condition is true
+func SkipIf(condition bool, message string) {
+	if condition {
+		Skip(message, 1)
+	}
+}
+
+// ObtainCentralRequest queries fleet-manager public API for the CentralRequest with id and stores in in the given pointer
+func ObtainCentralRequest(ctx context.Context, client *fleetmanager.Client, id string, request *public.CentralRequest) error {
```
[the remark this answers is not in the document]
johannes94: You're right. Its called that way because the private function I copied to `testuitl` was called like that before.

I changed it.

e2e/multicluster/multicluster_migration_test.go
johannes94: Do you mean dpCloudProvider should have a different value than dpRegion?

Unfortunately that's not so easy, because that's what we have for the cluster configurations all other tests do it like this as well.

ebensh (approved):
LGTM, thanks

PR #2149, opened 2025-01-17
ROX-27129: cleanup central CR reconciliation logic

ludydoo:
This PR changes some `CentralReconciler` method signatures to not accept a `v1alpha1.Central`, but only a `private.ManagedCentral`. This is in prevision of the complete removal of `v1alpha1.Central`


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -1296,8 +1303,11 @@ func (r *CentralReconciler) needsReconcile(changed bool, central *v1alpha1.Centr
 		return true
 	}
 
-	forceReconcile, ok := central.Labels["rhacs.redhat.com/force-reconcile"]
-	return ok && forceReconcile == "true"
+	if force, ok := remoteCentral.Spec.TenantResourcesValues["forceReconcile"].(bool); ok && force == true {
```
johannes94, line 1306:
```suggestion
	if force, ok := remoteCentral.Spec.TenantResourcesValues["forceReconcile"].(bool); ok && force {
```

johannes94 (approved):
Just a nitpick. Everything else looks good.

PR #2150, opened 2025-01-17
ROX-27129: remove v1alpha1.Central from the reconciler

ludydoo:
This PR removes the Central CR altogether from the `CentralReconciler`


fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -495,43 +360,11 @@ func (r *CentralReconciler) reconcileInstanceDeletion(ctx context.Context, remot
...
-
-	if central.Spec.Central == nil {
-		central.Spec.Central = &v1alpha1.CentralComponentSpec{}
-	}
-	if central.Spec.Central.DB == nil {
-		central.Spec.Central.DB = &v1alpha1.CentralDBSpec{}
-	}
-	central.Spec.Central.DB.IsEnabled = v1alpha1.CentralDBEnabledPtr(v1alpha1.CentralDBEnabledTrue)
-
+func (r *CentralReconciler) reconcileCentralDBConfig(ctx context.Context, remoteCentral *private.ManagedCentral) error {
```
johannes94, line 363: Not sure but I think this function is now not used anymore. And it's implementation also looks not useful (only return nil in all cases.

fleetshard/pkg/central/reconciler/argo_reconciler.go
```diff
@@ -33,16 +34,18 @@ type ArgoReconcilerOptions struct {
...
 	client ctrlClient.Client,
-	argoReconcilerOptions ArgoReconcilerOptions) *argoReconciler {
+	argoReconcilerOptions ArgoReconcilerOptions,
+	telemetry config.Telemetry,
+) *argoReconciler {
 	return &argoReconciler{
-		client:   client,
-		argoOpts: argoReconcilerOptions,
+		client:    client,
+		argoOpts:  argoReconcilerOptions,
```
johannes94, line 47: Consider moving Telemetry back to ArgoReconcilerOptionss. The idea was that every option that should influence / be passed through to the argo CD application, except per tenant values, should be part of that struct.   To prevent the constructor from ever growing number of function arguments.

This PR is a good example, by pulling telemetry outside the ArgoReconcilerOptions you had to modify the newArgoReconciler function, which in turn requires you to modify all usages of newArgoReconciler. Additionally this causes even more ambiguity for future changes, because for some options we use the ArgoReconcilerOptions for other we extend the constructor.

johannes94 (changes requested):

fleetshard/pkg/central/reconciler/argo_reconciler.go
ludydoo: Good point. This has revealed an existing probable bug, where we didn't set the `Telemetry` field in the `ArgoCDReconcilerOptions`: 

https://github.com/stackrox/acs-fleet-manager/pull/2150/files#diff-405fdfb027fcdcb359ab757ce7660f1511430e92eb03d82d1bdb05882d1d6a17R143

PR #2228, opened 2025-03-06
ROX-28194: Probe cleanup before execute

kovayur:
## Description
This PR makes changes to the probe service so that it can better handle unexpected errors. 

### Current problems
1. Cleanup is called after the instance creation. If the dataplane is unavailable, instance can be neither provisioned nor deleted, so even if the cleanup task fails, the probe will create a new one on the next iteration.
2. If multiple instances need to be deleted, the cleanup task waits until the first instance is deleted and then fails to execute the others by timeout. This leaves (N-1) instances pending, but they can at least be marked as deprovisioning so that fleetshard sync doesn't create resources for them on the cluster, when it's up again.
3. Cleanup task removes instances for all the regions. If a data plane cluster is unavailable in one region, it fails for both (because of #2) and keeps instances in the healthy region.
4. The prometheus rule is not filtered by the region, so if only one cluster is down, the alert will not be fired.

### What was done
1. Cleanup is now done on every probe execution
2. Common central functions are extracted to `central.Service`, which is abstracted from the FM API.
3. Instead of pre-configured "specs" (region, cloud provider), load them from FM at runtime.

## Test manual

Tested locally against integration environment


probe/pkg/probe/probe.go
```diff
@@ -119,43 +102,37 @@ func (p *ProbeImpl) cleanupFunc(ctx context.Context) error {
...
 		}
 	}
 
-	if success {
-		glog.Info("finished clean up attempt of probe resources")
+	if !centralsLeft {
+		glog.Infof("finished clean up attempt of probe resources. region=%s", p.spec.Region)
 		return nil
 	}
 	return errors.New("central clean up not successful")
```
ludydoo, line 118:
```suggestion
	if centralsLeft {
	  return errors.New("central clean up not successful")
	}
  glog.Infof("successfuly cleaned up attempt of probe resources. region=%s", p.spec.Region)
		return nil

```

probe/pkg/probe/probe.go
```diff
@@ -65,49 +51,46 @@ func (p *ProbeImpl) newCentralName() (string, error) {
...
+	glog.Infof("probe run has been started: fleetManagerEndpoint=%s, region=%s",
 		p.config.FleetManagerEndpoint,
-		spec.CloudProvider,
-		spec.Region,
+		p.spec.Region,
 	)
 	defer glog.Info("probe run has ended")
-	defer recordElapsedTime(time.Now(), spec.Region)
+	defer p.recordElapsedTime(time.Now())
+	if err := p.cleanup(ctx); err != nil {
```
ludydoo, line 61: Maybe add a comment to explain why cleanup is done before ?

probe/config/config.go
```diff
@@ -5,22 +5,18 @@ import (
...
 	AuthType                string        `env:"AUTH_TYPE" envDefault:"RHSSO"`
-	CentralSpecs            CentralSpecs  `env:"CENTRAL_SPECS" envDefault:"[{ \"cloudProvider\": \"standalone\", \"region\": \"standalone\" }]"`
 	FleetManagerEndpoint    string        `env:"FLEET_MANAGER_ENDPOINT" envDefault:"http://127.0.0.1:8000"`
 	MetricsAddress          string        `env:"METRICS_ADDRESS" envDefault:":7070"`
 	RHSSOClientID           string        `env:"RHSSO_SERVICE_ACCOUNT_CLIENT_ID"`
 	ProbeName               string        `env:"PROBE_NAME" envDefault:"${HOSTNAME}" envExpand:"true"`
-	ProbeCleanUpTimeout     time.Duration `env:"PROBE_CLEANUP_TIMEOUT" envDefault:"5m"`
 	ProbeHTTPRequestTimeout time.Duration `env:"PROBE_HTTP_REQUEST_TIMEOUT" envDefault:"5s"`
 	ProbePollPeriod         time.Duration `env:"PROBE_POLL_PERIOD" envDefault:"5s"`
 	ProbeRunTimeout         time.Duration `env:"PROBE_RUN_TIMEOUT" envDefault:"30m"`
```
ludydoo, line 22: Maybe increase this

probe/pkg/probe/probe.go
kovayur: Done.

PR #2270, opened 2025-03-27
ROX-27239: Remove auth provider check because of declarative config

kurlov:
## Description

After migration Central to declarative config there is no longer need to check auth provider manually from Fleetshard-sync service. Thus this reconciliation step can be removed


johannes94 (changes requested):
Looks mostly good.

But this line https://github.com/stackrox/acs-fleet-manager/pull/2270/files#diff-8b2c412ebc3fa158d46c2dfb0136cfef521abb4fb2dda0a77b403a32afd9c58eR1087

Would never return true with the current implementation. So reconciliation would effectively never be skipped.

fleetshard/pkg/central/reconciler/reconciler.go
```diff
@@ -351,24 +341,6 @@ func (r *CentralReconciler) restoreCentralSecrets(ctx context.Context, remoteCen
...
-		return true, nil
-	}
-
-	exists, err := r.verifyAuthProviderFunc(ctx, remoteCentral, r.client)
-	if err != nil {
-		return false, errors.Wrapf(err, "failed to verify that the default auth provider exists within "+
-			"Central %s/%s", remoteCentral.Metadata.Namespace, remoteCentral.Metadata.Name)
-	}
-	if exists {
-		r.hasAuthProvider = true
```
johannes94, line 366: This line needs to be added somewhere else, see comment above.

johannes94 (approved):

PR #2670, opened 2026-04-20
feat(fleetshard): add Pyroscope continuous profiling support

ebensh:
## Summary
- Add opt-in Pyroscope push-based continuous profiling to fleetshard-sync
- Configurable via `PYROSCOPE_ENABLED`, `PYROSCOPE_SERVER_ADDRESS`, `PYROSCOPE_AUTH_TOKEN`, and `PYROSCOPE_APPLICATION_NAME` env vars
- Profiles CPU, memory (alloc/inuse), goroutines, mutex, and block contention; tags all profiles with `cluster_id`
- Disabled by default; validation ensures server address is set when enabled

## Test plan
- [x] Config tests for enabled/disabled/missing-address scenarios pass
- [x] `go build ./fleetshard/...` compiles cleanly
- [ ] Deploy to dev cluster with Pyroscope instance, verify profiles appear in Pyroscope UI

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fleetshard/main.go
```diff
@@ -93,6 +94,39 @@ func main() {
 	pprofServer.Start()
 	defer pprofServer.Stop()
 
+	if config.Pyroscope.Enabled {
```
kovayur, line 97: Shall we disable/remove the standard pprof server?

kovayur:
It looks like changes to the Helm chart are missing.

kovayur:
Stale, closing

PR #2800, opened 2026-09-09
ROX-36893: allow unsetting billing parameters via admin API

johannes94:
Admin API requests:
- CentralBillingChangeRequest
- CentralSubscriptionChangeRequest

don't allow to set parameters to an empty string due to Gorms behavior of not considering empty values in update requests.

Because of that, changing a billing type from `marketplace-aws` to `standard` is inconsistent, since the associated CloudAccountID can't be set to empty string.

This PR changes both endpoints, so that setting to empty is enabled. The site effect is that on a requests all parameters must be in the body, or otherwise they'll be cleaned.

This is necessary to cleanup billing for a customer case.

Also changes the golang version used for pre-commit linting. Because local pre-commit version can use different default go version, which broke it for me.

## PR summary

[ROX-36893](https://redhat.atlassian.net/browse/ROX-36893) updates the `CentralBillingChangeRequest` and `CentralSubscriptionChangeRequest` admin API endpoints to allow billing parameters to be set to empty strings.

This addresses Gorm update behavior that ignores empty values, which currently prevents changing a billing type from `marketplace-aws` to `standard` while clearing the associated `CloudAccountID`.

With this change, all parameters must be included in requests; omitted parameters will be cleared.

---

*The following was generated by `@coderabbitai` and may be updated automatically.*

## Summary

- Updated [`ROX-36893`](https://issues.redhat.com/browse/ROX-36893) to persist empty billing values.
- Changed billing and subscription updates to use explicit field maps.
- Documented full-state replacement semantics in the API models and OpenAPI specifications.
- Pinned the `golangci-lint` pre-commit hook to Go `1.26.0`.

## Test manual

1. Send a billing change request with `billing_type` set to `standard` and an empty `cloud_account_id`.
2. Confirm that the billing type changes and `CloudAccountID` is cleared.
3. Omit a billing field from a request.
4. Confirm that the omitted field is cleared.
5. Repeat the checks for `CentralSubscriptionChangeRequest`.

---

[ROX-36893]: https://redhat.atlassian.net/browse/ROX-36893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kovayur (approved):
I'd love to see a few unit tests for this

johannes94:
@kovayur I changed the existing unit tests, since they were failing. I'm not happy with the pattern of testing the SQL that is send to the DB, but I worked with what is there and the test now ensure we send explicit updates to those fields.

